PR #58, opened 2017-08-02
Added batch for client internally

nazrhom:


src/sbvr-api/sbvr-utils.coffee
```diff
@@ -540,14 +541,34 @@ exports.runRule = do ->
...
-	class PinejsClient extends PinejsClientCore(_, Promise)
-		_request: ({ method, url, body, tx, req, custom }) ->
+	class PinejsClient extends PinejsClientCore(_, Promise, mimemessage)
+		compileBatch: (data) -> _.map(data, batchify)
+		_request: ({ method, url, body, multipart, tx, req, custom }) ->
+			if _.isArray(multipart?.data)
+				body = multipart.data
 			return runURI(method, url, body, tx, req, custom)
 
+batchify = (req) ->
```
Page-: Could you change `req` to `data` or `body` or similar, I expect `req` to refer specifically to en expressjs req object and based on `compileBatch: (data) -> _.map(data, batchify)` it's not

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -540,14 +541,34 @@ exports.runRule = do ->
 		.nodeify(callback)
 
 exports.PinejsClient =
-	class PinejsClient extends PinejsClientCore(_, Promise)
-		_request: ({ method, url, body, tx, req, custom }) ->
+	class PinejsClient extends PinejsClientCore(_, Promise, mimemessage)
+		compileBatch: (data) -> _.map(data, batchify)
+		_request: ({ method, url, body, multipart, tx, req, custom }) ->
+			if _.isArray(multipart?.data)
```
Page-: If `multipart.data` exists what can it be other than an array? And if it is something else should we return an error instead?

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -540,14 +541,34 @@ exports.runRule = do ->
 		.nodeify(callback)
 
 exports.PinejsClient =
-	class PinejsClient extends PinejsClientCore(_, Promise)
-		_request: ({ method, url, body, tx, req, custom }) ->
+	class PinejsClient extends PinejsClientCore(_, Promise, mimemessage)
```
Page-: Since we override `compileBatch` I don't think we actually need to pass in `mimemessage`?  Which would mean we can drop it as a dependency altogether

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -540,14 +541,34 @@ exports.runRule = do ->
...
+		compileBatch: (data) -> _.map(data, batchify)
```
Page-: I think it'd be worth adding a comment here that this is an optimisation to avoid the need to compile/parse the batch request (at least I assume that's why it's done?)

Page-: Doesn't this need to run the requests via pinejsClient.compile like in https://github.com/resin-io/pinejs-client-js/pull/34/files#diff-7211713139c36b3a80c048e901477db2R488 ?

Issue #60, opened 2017-08-10
Allow filtering of available roles during permission fetching


afitzek:
We want to be able to restrict credentials (like a JWT), to only a given subset of the roles the user actually has and therefore only to the permissions associated with that given role.

Flowdock: https://www.flowdock.com/app/rulemotion/r-security/threads/6aoRkUtQ2XpwuPkwBu9rvnNrXb_

PR #61, opened 2017-08-10
WIP: Allow filtering of roles during permission fetching

afitzek:
This adds an optional external whitelist of roles available to a user.

We want to be able to restrict credentials (like a JWT), to only a given subset of the roles the user actually has and therefore only to the permissions associated with that given role.

Flowdock: https://www.flowdock.com/app/rulemotion/r-security/threads/6aoRkUtQ2XpwuPkwBu9rvnNrXb_

Change-Type: minor
Connects-To: #60

**This is untested so far, I just tested it by changing the js directly in the api node_modules.**

### DIFF
--- a/src/sbvr-api/permissions.coffee
+++ b/src/sbvr-api/permissions.coffee
@@ -180,7 +180,10 @@ exports.setup = (app, sbvrUtils) ->
 			throw err
 		.nodeify(callback)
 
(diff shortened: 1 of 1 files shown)


Page- (changes requested):
Could you add a description of the use case to the PR description please

PR #82, opened 2017-09-30
Convert uri-parser to typescript

Page-:
Change-type: patch


tsconfig.json
```diff
@@ -9,7 +9,8 @@
 		"removeComments": true,
 		"rootDir": "src",
 		"sourceMap": true,
-		"strictNullChecks": true
+		"strictNullChecks": true,
+		"target": "es2015"
```
Page-: This was necessary to allow us to use `Map` in typescript

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+}
+
+export const parseOData = (b: UnparsedRequest) => {
+	return Promise.try<ODataRequest | ODataRequest[]>(() => {
+		if (b._isChangeSet && b.changeSet != null) {
+			const env = new Map<ODataRequest['id'], ODataRequest>()
+			// We sort the CS set once, we must assure that requests which reference
+			// other requests in the changeset are placed last. Once they are sorted
+			// Map will guarantee retrival of results in insertion order
+			const sortedCS = _.sortBy(b.changeSet, (el) => el.url[0] !== '/')
```
dfunckt: Parentheses are optional around the closure argument, i.e. it can be `el => el.url[0] !== '/'`. Same below.

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+			return {
+				method: b.method as SupportedMethod,
+				vocabulary: apiRoot,
+				resourceName: odata.tree.resource,
+				odataBinds: odata.binds,
+				odataQuery: odata.tree,
+				values: b.data,
+				custom: {},
+				_defer: false,
+			}
```
dfunckt: TIL it's this guy that provides the `request` argument in hooks :)

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+		throw new BadRequestError('Content-Id must be unique inside a changeset')
+	}
+
+	let defer: boolean
+	let odata
+	let apiRoot: string
+	let url
+
+	if (b.url[0] === '/') {
+		({ url, apiRoot } = splitApiRoot(b.url))
```
dfunckt: Why have parentheses around the whole statement?

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+		odata = odataParser.matchAll(url, 'Process')
+		defer = false
+	} else {
+		url = b.url
+		odata = odataParser.matchAll(url, 'Process')
+		const { bind } = odata.tree.resource
+		const [ , id ] = odata.binds[bind]
```
dfunckt: Destructuring is Good, but why not write `const id = odata.binds[bind][1]` instead?

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+	}
+	env.set(contentId, parseResult)
+	return env
+}
+
+const splitApiRoot = (url: string) => {
+	let urlParts = url.split('/')
+	const apiRoot = urlParts[1]
+	if (apiRoot == null || odata2AbstractSQL[apiRoot] == null) {
+		throw new ParsingError(`No such api root: ${apiRoot}`)
```
dfunckt: Not much point logging `No such api root: null` if `apiRoot` is null -- maybe break in two separate ifs that throw the same error with different messages?

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+	}
+	url = '/' + urlParts.slice(2).join('/')
+	return { url: url, apiRoot: apiRoot }
+}
+
+const mustExtractHeader = (body: { headers?: { [header: string]: string } }, header: string) => {
+	const h: any = _.get(body.headers, [header, 0])
+	if (_.isEmpty(h)) {
```
dfunckt: Isn't `if (!h)` pretty much the same?

src/sbvr-api/uri-parser.ts
Page-: I know, but I prefer to have those parens there, it makes it clearer to me that it's a function rather than a value

src/sbvr-api/uri-parser.ts
Page-: Because it's required for destructuring into existing vars in es6

src/sbvr-api/uri-parser.ts
Page-: `_.isEmpty(h)` and `!h` have different behaviours, and I don't know for sure if the `!h` behaviour is ok in this

dfunckt: Based on the type, `h` will be a `string` (or `undefined` if there's no `headers` in body). So `!h` will be true for all of `'' | null | undefined`. If that's not true, let's make sure it's reflected in the argument's type as well.

src/sbvr-api/uri-parser.ts
Page-: Because this matches how it was in the coffeescript version ( https://github.com/resin-io/pinejs/pull/82/files#diff-c52cf3760792616d7d710ad38b04787dL86 ) and in general I try to avoid changes beyond what is necessary for typescript to be happy (to minimise the chance of introducing an unintended issue)

src/sbvr-api/uri-parser.ts
CameronDiver: Agreed, seeing that error in logs is still going to leave questions open, might as well minimise that.

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
+import * as _AbstractSQLCompiler from '@resin/abstract-sql-compiler'
+
+import * as Promise from 'bluebird'
+import TypedError = require('typed-error')
+const { ODataParser } = require('@resin/odata-parser')
```
CameronDiver: I feel like a bit of whitespace would go a long way here.

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+
+const notBadRequestOrParsingError = (e: Error) => {
+	return !((e instanceof BadRequestError) || (e instanceof ParsingError))
+}
+
+export const parseOData = (b: UnparsedRequest) => {
+	return Promise.try<ODataRequest | ODataRequest[]>(() => {
+		if (b._isChangeSet && b.changeSet != null) {
+			const env = new Map<ODataRequest['id'], ODataRequest>()
+			// We sort the CS set once, we must assure that requests which reference
```
CameronDiver: NIT: changeset set -> changeset

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+				console.error('Failed to translate url: ', JSON.stringify(odataQuery, null, '\t'), method, e, e.stack)
+				throw new TranslationError('Failed to translate url')
+			}
+		},
+		{ normalizer: JSON.stringify }
+	)
+	return (vocabulary: string, odataQuery: ODataQuery, method: SupportedMethod, body: {}) => {
+		// Sort the body keys to improve cache hits
+		const { tree, extraBodyVars } = memoizedOdata2AbstractSQL(vocabulary, odataQuery, method, _.keys(body).sort())
+		_.assign(body, extraBodyVars)
```
LucianBuzzo: I know this is pre-existing, but does this function return a value *and* mutate the `body` parameter? Is there a cleaner way of doing this?

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+	return !((e instanceof BadRequestError) || (e instanceof ParsingError))
+}
+
+export const parseOData = (b: UnparsedRequest) => {
+	return Promise.try<ODataRequest | ODataRequest[]>(() => {
+		if (b._isChangeSet && b.changeSet != null) {
+			const env = new Map<ODataRequest['id'], ODataRequest>()
+			// We sort the CS set once, we must assure that requests which reference
+			// other requests in the changeset are placed last. Once they are sorted
+			// Map will guarantee retrival of results in insertion order
```
LucianBuzzo: NIT `retrieval`

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+			const sortedCS = _.sortBy(b.changeSet, (el) => el.url[0] !== '/')
+			return Promise.reduce(sortedCS, parseODataChangeset, env)
+			.then((env) => Array.from(env.values()) as ODataRequest[])
```
LucianBuzzo: `env` is a shadowed variable, can we change the name to `parsed` ?

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+		}
+	}).catch(SyntaxError, () => {
+		throw new BadRequestError(`Malformed url: '${b.url}'`)
+	}).catch(notBadRequestOrParsingError, (e) => {
+		console.error('Failed to parse url: ', b.method, b.url, e, e.stack)
+		throw new ParsingError(`Failed to parse url: '${b.url}'`)
+	})
+}
+
+const parseODataChangeset = (env: Map<ODataRequest['id'], ODataRequest>, b: UnparsedRequest) => {
```
LucianBuzzo: I know this is pre-existing but can we give `b` a more semantic name?

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+	const apiRoot = urlParts[1]
+	if (apiRoot == null || odata2AbstractSQL[apiRoot] == null) {
+		throw new ParsingError(`No such api root: ${apiRoot}`)
+	}
+	url = '/' + urlParts.slice(2).join('/')
+	return { url: url, apiRoot: apiRoot }
+}
+
+const mustExtractHeader = (body: { headers?: { [header: string]: string } }, header: string) => {
+	const h: any = _.get(body.headers, [header, 0])
```
LucianBuzzo: Can this just be `h: string`?

src/sbvr-api/uri-parser.ts
```diff
@@ -0,0 +1,225 @@
...
+export class BadRequestError extends TypedError {}
+
+type SupportedMethod = "GET" | "PUT" | "POST" | "PATCH" | "MERGE" | "DELETE" | "OPTIONS"
+type ODataQuery = any[]
+
+export interface UnparsedRequest {
+	method: string
+	url: string
+	data: any
+	headers?: { [header: string]: string }
```
nazrhom: I think this should be `{ [header: string]: string[] }` 
(Im not sure why header values are list of strings instead of simple strings, but this comes from dicer which is the multipart request parser)

src/sbvr-api/uri-parser.ts
nazrhom: Akis is correct on the type here. This currently works because you can index into strings directly to access the characters, but the argument is supposed to be ` { [header: string]: string[] }`

src/sbvr-api/uri-parser.ts
nazrhom: I think this should be `{ [header: string]: string[] }` 
(Im not sure why header values are list of strings instead of simple strings, but this comes from dicer which is the multipart request parser)

Page-:
@resin-ci retest

PR #83, opened 2017-09-30
Convert control-flow to typescript

Page-:


src/sbvr-api/control-flow.ts
```diff
@@ -0,0 +1,61 @@
...
+			return Promise.resolve(a).then(fn)
+		}
+	// TODO: We cast here to merge `Promise<U> | Promise<U[]>` into `Promise<U | U[]>,
+	// which is equivalent for our uses but fixes issues where typing was being rejected
+	}) as ((a: T | T[]) => Promise<U | U[]>)
+}
+
+interface MappingFunction {
+	<T, U>(a: T[], fn: (v: T) => U | U[] | Promise<U>): Promise<Array<U | U[] | Error>>
+}
```
nazrhom: I think you can drop the `U[]` in here. 
```
interface MappingFunction {
	<T, U>(a: T[], fn: (v: T) => U | Promise<U>): Promise<Array<U | Error>>
}
```
Works for me

src/sbvr-api/control-flow.ts
```diff
@@ -0,0 +1,61 @@
...
+
+// Maps fn over collection and returns an array of Promises. If any promise in the
+// collection is rejected it returns an array with the error, along with all the
+// promises that were fulfilled up to that point
+const mapTill: MappingFunction = <T, U>(a: T[], fn: (v: T) => U) => {
+	const runF = Promise.method(fn)
+	const results: Array<U | any> = []
+	return Promise.each(a, (p) => {
+		return runF(p)
+		.then((result) => {
```
nazrhom: This is actually giving me an implicit `any` type error. I think the problem is with the signature in `Promise.method`. I wrapped `Promise.method` in 
```
function toPromiseFn<T> (fn:(...args: any[]) => (T | Promise<T>)) {
	return Promise.method(fn) as (...args: any[]) => Promise<T>
}
```
This way, it will correctly infer the type of result in here

src/sbvr-api/control-flow.ts
```diff
@@ -0,0 +1,61 @@
...
+const settleMapSeries: MappingFunction = (a, fn) => {
+	const runF = Promise.method(fn)
+	return Promise.mapSeries(a, _.flow(runF, wrap))
+}
+
+// Wrap a promise with reflection. This promise will always succeed, either
+// with the value or the error of the promise it is wrapping
+const wrap = <T>(p: Promise<T>) => {
+	return p.then(_.identity as (value: T) => T, _.identity as (value: Error) => Error)
+}
```
nazrhom: Minor: I like this a bit better as 
```
const wrap = <T>(p: Promise<T>):(Promise<T|Error>) => {
	return p.then(_.identity, _.identity)
}
```

Page-: If I try that I get:
```
Type 'Bluebird<undefined>' is not assignable to type 'Bluebird<Error | T>'.
  Type 'undefined' is not assignable to type 'Error | T'.

File: src/sbvr-api/control-flow.ts (hunk 2 above)
Comment (Page-) [reply src/sbvr-api/control-flow.ts]: For me `result` is typed as `U`, can you check you've got all the types installed?

File: src/sbvr-api/control-flow.ts (hunk 1 above)
Comment (Page-) [reply src/sbvr-api/control-flow.ts]: I think I added it because of sbvr-utils usage needing it, I'll check that

Comment (Page-) [reply src/sbvr-api/control-flow.ts]: It isn't immediately obvious the the `U[]` is required so I've removed it and we can cross that bridge later (also I updated bluebird definitions and the `U[]` started causing issues.. so it definitely needs to be done differently if it does need to be done)

File: src/sbvr-api/control-flow.ts (hunk 2 above)
Comment (nazrhom) [reply src/sbvr-api/control-flow.ts]: looks like the problem is with the required bluebird version. If I update to 3.5.16 it infers like yours, but the problem is that I had 3.5.4 installed previously which satisfied the constraint in package.json but made it fail.

File: src/sbvr-api/control-flow.ts [hunk 4]
```diff
@@ -0,0 +1,61 @@
...
+		if (_.isArray(a)) {
+			// This must not be a settle as if any operation fails in a cs
+			// we want to discard the whole
+			return Promise.mapSeries(a, fn)
+		} else {
+			return Promise.resolve(a).then(fn)
+		}
+	// TODO: We cast here to merge `Promise<U> | Promise<U[]>` into `Promise<U | U[]>,
+	// which is equivalent for our uses but fixes issues where typing was being rejected
+	}) as ((a: T | T[]) => Promise<U | U[]>)
```
dfunckt: Instead of casting, can you try specifying the function's return type explicitly and see if that works?

Also, why is this not a proper function? Function variables suck because they're anonymous and therefore invisible in back traces.

src/sbvr-api/control-flow.ts
nazrhom: Hmmm this is also apparently due to a patch version mismatch. I had @types/lodash@4.14.64 and it works fine with that, updating to 4.14.77 gives me the same error you have. (Which might be a regression as the error doesn't really make much sense)

Page-: I've updated package.json to the latest versions to help a bit with with this

src/sbvr-api/control-flow.ts
Page-: I've updated package.json to the latest versions to help a bit with with this

src/sbvr-api/control-flow.ts
```diff
@@ -0,0 +1,60 @@
...
+			return Promise.resolve(a).then(fn)
+		}
+	}
+}
+
+interface MappingFunction {
+	<T, U>(a: T[], fn: (v: T) => U | Promise<U>): Promise<Array<U | Error>>
+}
+
+// The settle- versions of
```
dfunckt: This comment looks like it’s partial?

src/sbvr-api/control-flow.ts
```diff
@@ -0,0 +1,60 @@
+import * as _ from 'lodash'
+import * as Promise from 'bluebird'
+
+// TODO: We force the return type here to merge `Promise<U> | Promise<U[]>` into `Promise<U | U[]>,
+// which is equivalent for our uses but fixes issues where typing were being rejected
```
dfunckt: Another point for requiring explicit function return types? :p

src/sbvr-api/control-flow.ts
```diff
@@ -0,0 +1,60 @@
...
+export const liftP = <T, U>(fn: (v: T) => U | Promise<U>): (a: T | T[]) => Promise<U | U[]> => {
```
CameronDiver: I agree we should be marking this as a proper function, I know we try to minimise the difference between the CS and TS versions when converting, but let's not make things more complicated just to keep meeting that end.

src/sbvr-api/control-flow.ts
```diff
@@ -0,0 +1,60 @@
...
+			return Promise.mapSeries(a, fn)
+		} else {
+			return Promise.resolve(a).then(fn)
+		}
+	}
+}
+
+interface MappingFunction {
+	<T, U>(a: T[], fn: (v: T) => U | Promise<U>): Promise<Array<U | Error>>
+}
```
CameronDiver: You can just do 
```typescript
type MappingFunction = <T, U>(a: T[], fn: (v: T) => U | Promise<U>) => Promise<Array<U | Error>>

Comment (Page-) [reply src/sbvr-api/control-flow.ts]: You can, but I was under the impression that interfaces are preferable? @pimterry ?

File: src/sbvr-api/control-flow.ts (hunk 7 above)
Comment (Page-) [reply src/sbvr-api/control-flow.ts]: Why would it make it simpler to use `function...` instead? It would just mean that `this` could be anything, not that it matters either way when we don't use it

File: src/sbvr-api/control-flow.ts (hunk 5 above)
Comment (Page-) [reply src/sbvr-api/control-flow.ts]: Yeah, I just left it completely as-is from the coffeescript, @nazrhom what is the comment meant to be?

File: src/sbvr-api/control-flow.ts (hunk 6 above)
Comment (Page-) [reply src/sbvr-api/control-flow.ts]: Not really, `Promise<U> | Promise<U[]>` is the more accurate return type (it's obvious why when you think of an `Array<U> | Array<U[]>` case), it's just it was breaking types and in this particular case is safe to cast to the alternative form - I think I can switch to an even better typing though

File: src/sbvr-api/control-flow.ts (hunk 7 above)
Comment (dfunckt) [reply src/sbvr-api/control-flow.ts]: It's simpler because top level functions don't have "this" anyway, so it doesn't matter that they're functions, and you get better debug-ability for free due to the function name being visible in stack traces.

File: src/sbvr-api/control-flow.ts (hunk 8 above)
Comment (pimterry) [reply src/sbvr-api/control-flow.ts]: https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases suggests interfaces are mildly preferable, but personally I don't have any strong opinions on it for this kind of case. It does make a larger difference if you're defining an object type, since you can't extend type aliases, but for this I'm not sure it makes a lot of difference.

File: src/sbvr-api/control-flow.ts (hunk 6 above)
Comment (Page-) [reply src/sbvr-api/control-flow.ts]: Ok, so the better typing was
```typescript
export const liftP = <T, U>(fn: (v: T) => U | Promise<U>) => {
	return ((a: T | T[]) => {
		if (_.isArray(a)) {
			// This must not be a settle as if any operation fails in a cs
			// we want to discard the whole
			return Promise.mapSeries(a, fn)
		} else {
			return Promise.resolve(a).then(fn)
		}
	}) as {
		(a: T[]): Promise<U[]>
		(a: T): Promise<U>
	}
}
```
which is valid typing and can differentiate the response correctly, and is generally the more accurate typing, but unfortunately won't work for our needs as it still ends up as `Promise<U> | Promise<U[]>`

src/sbvr-api/control-flow.ts
Page-: I changed it to ``The settle version of `Promise.mapSeries` `` which seems sane

src/sbvr-api/control-flow.ts
Page-: > One difference is that interfaces create a new name that is used everywhere. Type aliases don’t create a new name — for instance, error messages won’t use the alias name.

That one seems like it could be pretty helpful at times (seeing massive long type defs like this has been a pita for me in the past, compared to if it used a name)

src/sbvr-api/control-flow.ts
Page-: I didn't see your comment yesterday (I blame github's failed updating!) but I agree this case should always have used the return value to cast the return type.  I expect I just wasn't aware you could do it that way when I first wrote it

src/sbvr-api/control-flow.ts
Page-: The function name will be visible with the arrow style as well.  And I would expect the top level function to have a `this` variable, depending on the context it's called in of course, eg
```
import * as controlFlow from 'control-flow'
controlFlow.liftP(...)
``
I would expect to have `this` be `controlFlow` for the `function` style (certainly if using an ES5 target, maybe true ES6 modules have a special case for this)

But my real problem is that I really don't get the argument that it's simpler, it's an alternative for this function for sure as it doesn't change anything, but it doesn't seem in any way simpler to use `function liftP (...) {` vs `liftP = (...) => {` - so I'm against changing it for the reasoning that it's "simpler" because I don't see the argument or benefit from doing so and I'd like to understand why you're saying it is

src/sbvr-api/control-flow.ts
Page-: cc @CameronDiver you may want to know this about interfaces vs type aliases if you're currently preferring type aliases

CameronDiver: Ah interesting. I would have considered them equivalent with type aliases being the succinct way of writing it. Thanks.

PR #89, opened 2017-10-09
Add a .travis.yml to test on travisci

Page-:
Change-type: patch


brownjohnf:
Why are you using travis instead of circle?

Page-:
Travis allows easily selecting the node version and running tests against multiple node versions, whilst it's possible to do on circle with a lot of heavy lifting it's really easy to do with travis (it's what the sdk/cli use, presumably for the same reason)

pimterry:
The SDK & CLI do both use it, but largely for historical reasons. Moving them to circle isn't very high on my todo list, and I haven't really played with it at all yet, but it is in the plan.

Issue #91, opened 2017-11-15
Create a canAccess custom endpoint


afitzek:
Implement an endpoint to augment user permission to support different actions on resources.

See: https://github.com/resin-io/resin-api/issues/640

dfunckt:
This is done.

PR #92, opened 2017-11-15
Add meta endpoint `$canAccess` for every api root, to allow checking …

afitzek:
…resource access for arbitrary actions

This adds an `$canAccess` endpoint to every api root. This endpoint accepts POST requests
with parameters `resource`, `action` and `id`.
- The `resource` parameter holds the resource name in question.
- The `action` parameter holds the action in question.
- The `id` parameter holds the resource id in question.

The endpoint responds with an empty response with http status code 200 if access is granted.
The endpoint responds with an http status code 401 if access is not granted.
The endpoint can also respond with http status code 400 or 500 in case of client or server errors.

Change-Type: minor
Connects-To: #91
---- Autogenerated Waffleboard Connection: Connects to #91

### DIFF
--- a/src/config-loader/config-loader.coffee
+++ b/src/config-loader/config-loader.coffee
@@ -2,7 +2,6 @@ _ = require 'lodash'
 Promise = require 'bluebird'
 sbvrUtils = require '../sbvr-api/sbvr-utils'
 permissions = require '../sbvr-api/permissions'
-resourceaccess = require '../sbvr-api/resource-access'
 fs = Promise.promisifyAll(require('fs'))
 
 # Setup function
@@ -94,7 +93,6 @@ exports.setup = (app) ->
 				Promise.map data.models, (model) ->
 					if model.modelText?
 						apiRoute = '/' + model.apiRoot + '/*'
-						app.post('/' + model.apiRoot + '/([\$])canAccess', sbvrUtils.authorizationMiddleware, resourceaccess.canAccessV1Request)
 						app.options(apiRoute, (req, res) -> res.sendStatus(200))
 						app.all(apiRoute, sbvrUtils.handleODataRequest)
 
--- a/src/config-loader/config-loader.coffee
+++ b/src/config-loader/config-loader.coffee
@@ -2,6 +2,7 @@ _ = require 'lodash'
 Promise = require 'bluebird'
 sbvrUtils = require '../sbvr-api/sbvr-utils'
 permissions = require '../sbvr-api/permissions'
+resourceaccess = require '../sbvr-api/resource-access'
 fs = Promise.promisifyAll(require('fs'))
 
 # Setup function
@@ -93,6 +94,7 @@ exports.setup = (app) ->
 				Promise.map data.models, (model) ->
 					if model.modelText?
 						apiRoute = '/' + model.apiRoot + '/*'
+						app.post('/' + model.apiRoot + '/([\$])canAccess', sbvrUtils.authorizationMiddleware, resourceaccess.canAccessV1Request)
 						app.options(apiRoute, (req, res) -> res.sendStatus(200))
 						app.all(apiRoute, sbvrUtils.handleODataRequest)
 
--- a/src/sbvr-api/resource-access.ts
+++ b/src/sbvr-api/resource-access.ts
@@ -29,7 +29,7 @@ interface PermissionHolder {
 
 function manipulatePermissions(originalPermissions: string[], resource: string, action: string): string[] {
 	const targetActionPermissions = `resin.${resource}.${action}`;
-	let permissionsInQuestion: string[] = [];
+	const permissionsInQuestion: string[] = [];
 	originalPermissions.forEach((permission: string) => {
 		if(_.startsWith(permission, targetActionPermissions)) {
 			const rewrittenPermission = `resin.${resource}.read` + permission.slice(targetActionPermissions.length);
--- a/src/sbvr-api/resource-access.ts
+++ b/src/sbvr-api/resource-access.ts
@@ -3,18 +3,10 @@ import * as _ from 'lodash'
 import * as Promise from 'bluebird'
 import TypedError = require('typed-error')
 import { AnyObject } from 'pinejs-client/core';
+const { BadRequestError } = require('./uri-parser');
 import PinejsClient = require('pinejs-client');
 
-class BadInputError extends TypedError {}
 class UnauthorizedError extends TypedError {}
-
-const sbvrUtils = require('./sbvr-utils');
-
-const getAPIClient: GetAPIClient = sbvrUtils.getAPIClient as GetAPIClient
-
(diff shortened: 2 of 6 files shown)


src/sbvr-api/resource-access.ts
```diff
@@ -0,0 +1,153 @@
+import * as express from 'express'
+import * as _ from 'lodash'
+import * as Promise from 'bluebird'
+import TypedError = require('typed-error')
+import { AnyObject } from 'pinejs-client/core';
+import PinejsClient = require('pinejs-client');
+
+class BadInputError extends TypedError {}
+class UnauthorizedError extends TypedError {}
+
```
nazrhom: Similar errors are generated and exported in sbvr-utils and uri-parser, maybe have a look to see if there is any overlap with those

src/sbvr-api/resource-access.ts
```diff
@@ -0,0 +1,153 @@
...
+interface PermissionHolder {
+	actor?: number
+	key?: string
+	permissions: string[]
+}
+
+function manipulatePermissions(originalPermissions: string[], resource: string, action: string): string[] {
+	const targetActionPermissions = `resin.${resource}.${action}`;
+	let permissionsInQuestion: string[] = [];
```
nazrhom: I think this can be a `const`, arrays can be constant even if you push to them

src/sbvr-api/resource-access.ts
afitzek: I would rather stay with these errors. uri-parser exports a `BadRequestError` but it is still in coffee script and I did not find a similar `UnauthorizedError`.

src/config-loader/config-loader.coffee
```diff
@@ -2,6 +2,7 @@ _ = require 'lodash'
 Promise = require 'bluebird'
 sbvrUtils = require '../sbvr-api/sbvr-utils'
 permissions = require '../sbvr-api/permissions'
+resourceaccess = require '../sbvr-api/resource-access'
```
flesler: I feel I'm commenting too many NITs but, can you upper case the `a` so it's proper camel case?

src/config-loader/config-loader.coffee
```diff
@@ -93,6 +94,7 @@ exports.setup = (app) ->
 				Promise.map data.models, (model) ->
 					if model.modelText?
 						apiRoute = '/' + model.apiRoot + '/*'
+						app.post('/' + model.apiRoot + '/([\$])canAccess', sbvrUtils.authorizationMiddleware, resourceaccess.canAccessV1Request)
```
flesler: This should match what? `/canAccess` and `/$canAccess`?

afitzek: This matches only `/$canAccess`.

flesler: Cool, why not then:

```
'/' + model.apiRoot + '/$canAccess'
```

afitzek: I had to look this up, but if you want to use `$` in a path in an express route you need to escape it like this. 

"If you need to use the dollar character ($) in a path string, enclose it escaped within ([ and ]). For example, the path string for requests at “/data/$book”, would be “/data/([\$])book”." (see: https://expressjs.com/en/guide/routing.html)

src/config-loader/config-loader.coffee
afitzek: Don't worry you are totally right, the upper case `A` makes it much easier to read. :)

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -559,6 +559,11 @@ exports.runRule = do ->
 					return result
 		.nodeify(callback)
 
+exports.getAPIClient = (vocab) ->
```
Page-: You can just use `sbvrUtils.api[vocab]` directly (it's what everything else does), there's no need for this function

src/sbvr-api/resource-access.ts
```diff
@@ -0,0 +1,153 @@
...
+
+		const id: number = parseInt(idString);
+
+		return checkAccessForResourceWithRequest(req, api, resource, action, id);
+	})
+	.then( () => {
+		res.sendStatus(200);
+	})
+	.catch( (err) => {
+		if (err instanceof UnauthorizedError) {
```
Page-: Can you catch this in the standard bluebird way please, ie
```typescript
.catch(UnauthorizedError, () => {
	res.sendStatus(401)
})
...
```

src/sbvr-api/resource-access.ts
```diff
@@ -0,0 +1,153 @@
...
+	Promise.try( () => {
+		const resource: string = _.get(req.body, 'resource');
+		const action: string = _.get(req.body, 'action');
+		const idString: string = _.get(req.body, 'id');
+
+		if (resource === undefined || action === undefined || idString === undefined) {
+			throw new BadInputError('Missing parameters');
+		}
+
+		const id: number = parseInt(idString);
```
Page-: There shouldn't be any need for the `: number` here?

src/sbvr-api/resource-access.ts
Page-: I'd strongly prefer to just use the same `BadRequestError` in both places, right now any code that might hit either of these has to separately handle both of these errors which will almost certainly lead to cases where only one is handled and the other slips through, despite being essentially the same thing

src/sbvr-api/resource-access.ts
```diff
@@ -0,0 +1,153 @@
...
+	const url = req.url.split('/')
+	const apiRoot = url[1]
+	const api = getAPIClient(apiRoot);
+	if (api == null)
+		return next('route')
+
+	Promise.try( () => {
+		const resource: string = _.get(req.body, 'resource');
+		const action: string = _.get(req.body, 'action');
+		const idString: string = _.get(req.body, 'id');
```
Page-: There's no guarantee that any of these are actually strings, do you rely on them being strings anywhere? Because if so then that could cause problems

src/config-loader/config-loader.coffee
Page-: I think this should actually be handled by the already added `*` path below, the same way other meta endpoints like `$metadata` are, and hook into the same `handleODataRequest` chain, that way internal api requests will be able to be made to this endpoint rather than only being able to make external http calls

afitzek: Sounds good. will use the same chain as it is used for `$metadata`.

afitzek:
@Page- I moved the `$canAccess` endpoint into the `handleODataRequest`. I could not implement it in the exact same way as `$metadata` because the ODataParser would not accept the custom endpoint `$canAccess`. It should be fine now also for internal api requests, since the `runURI` uses the `handleODataRequest`.

afitzek:
Closing this PR because, #98 superseded this PR.

PR #94, opened 2017-11-17
Guarantee that any error returned from the settle functions is an error object, this fixes later code that checks if an error was returned by using `_.isError`

Page-:
Change-type: patch


package.json
```diff
@@ -30,21 +30,21 @@
     "typed-error": "^2.0.0"
   },
   "devDependencies": {
-    "@types/bluebird": "^3.5.16",
-    "@types/body-parser": "^1.16.3",
+    "@types/bluebird": "3.5.16",
```
Page-: These are changes I had locally to ensure we used fixed types versions because they break the build in patch version all the time

PR #96, opened 2017-11-28
Set up circleci auto-publishing

Page-:
Change-type: patch


circle.yml
```diff
@@ -1,37 +1,89 @@
...
+          at: $CIRCLE_WORKING_DIRECTORY
+      - run:
+          name: Login
+          command: |
+            echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
+            npm whoami
+      - deploy:
+          name: Deploy
+          command: npm publish
+          # Output used for publish is from node 8 build
```
wrboyce: If publishing a node 8 build, should [`package-lock.json`](https://docs.npmjs.com/files/package-lock.json) be included in the repo?

Page-: package-lock.json is ignored when publishing, so I don't think it matters either way really

PR #98, opened 2017-12-08
Add generic `canAccess` action for every resource

afitzek:
Alternative to: https://github.com/resin-io/pinejs/pull/92

Change-Type: minor
Connects-To: #91
---- Autogenerated Waffleboard Connection: Connects to #91


src/sbvr-api/permissions.coffee
```diff
@@ -122,7 +122,24 @@ exports.setup = (app, sbvrUtils) ->
...
 			apiKeyMiddleware(req)
 		POSTPARSE: ({ req, request }) ->
+			if (request.method == 'POST' and request.odataQuery.property?.resource == 'canAccess')
+				request.odataQuery.options = { '$select': { 'properties': [ { 'name': 'id' } ] } }
+				{ action } = request.values
+				request.permissionType = action
+				request.resourceName = request.resourceName.slice(0, request.resourceName.length - '#canAccess'.length)
+				request.odataQuery.resource = request.resourceName
+				delete request.odataQuery.property
+				req.method = 'GET'
```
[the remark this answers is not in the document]
nazrhom: This is problematic when using batch requests. The reason is that the `req` object is shared for all the requests inside a batch, so everything set onto `req` directly should be independent of the `request` object. 
Can we set `isAction` directly on `request`?

src/sbvr-api/permissions.coffee
```diff
@@ -122,7 +123,28 @@ exports.setup = (app, sbvrUtils) ->
...
+				if method? and methodPermissions[method]?
+					request.permissionType = methodPermissions[method]
+				else if action?
+					request.permissionType = action
+				else
+					throw new MissingOrInvalidArgumentsError()
+				request.resourceName = request.resourceName.slice(0, request.resourceName.length - '#canAccess'.length)
+				request.odataQuery.resource = request.resourceName
+				delete request.odataQuery.property
+				request.method = 'GET'
```
nazrhom: I guess we are changing the method on the request for the subsequent addPermissions call. What about calling `_addPermission` directly with `methodPermissions.GET`? This way we could avoid mutation on `request`

afitzek: No we change the method to perform only a read only select, because an action is always called as `POST`. The permissions are collected by setting `request.permissionType`.

nazrhom: Ohhh, that makes sense. I guess mutating it makes sense then

src/sbvr-api/permissions.coffee
```diff
@@ -122,7 +123,28 @@ exports.setup = (app, sbvrUtils) ->
 		PREPARSE: ({ req }) ->
 			apiKeyMiddleware(req)
 		POSTPARSE: ({ req, request }) ->
+			if (request.method == 'POST' and request.odataQuery.property?.resource == 'canAccess')
+				request.odataQuery.options = { '$select': { 'properties': [ { 'name': 'id' } ] } }
+				{ action, method } = request.values
+				if method? and methodPermissions[method]?
```
Page-: I think it would make sense to return a `400` if both `action` and `method` are specified, rather than silently dropping the `action`, that way it avoids the possibility of confusion

src/sbvr-api/permissions.coffee
```diff
@@ -122,7 +123,28 @@ exports.setup = (app, sbvrUtils) ->
 		PREPARSE: ({ req }) ->
 			apiKeyMiddleware(req)
 		POSTPARSE: ({ req, request }) ->
+			if (request.method == 'POST' and request.odataQuery.property?.resource == 'canAccess')
+				request.odataQuery.options = { '$select': { 'properties': [ { 'name': 'id' } ] } }
```
Page-: You should add a `$top=1` as well, to avoid unnecessary work

src/sbvr-api/permissions.coffee
```diff
@@ -122,7 +123,28 @@ exports.setup = (app, sbvrUtils) ->
...
+				request.resourceName = request.resourceName.slice(0, request.resourceName.length - '#canAccess'.length)
+				request.odataQuery.resource = request.resourceName
+				delete request.odataQuery.property
+				request.method = 'GET'
+				request.custom.isAction = 'canAccess'
 			addPermissions(req, request)
+		PRERESPOND: ({ request, data }) ->
+			if (request.custom.isAction == 'canAccess')
+				if (_.isArray(data) and data.length == 1)
+					data[0] = _.pick(data[0], 'id', '__metadata')
```
Page-: Why do we return the `id` rather than just true/false?

src/sbvr-api/permissions.coffee
```diff
@@ -122,7 +123,28 @@ exports.setup = (app, sbvrUtils) ->
...
+					throw new MissingOrInvalidArgumentsError()
+				request.resourceName = request.resourceName.slice(0, request.resourceName.length - '#canAccess'.length)
+				request.odataQuery.resource = request.resourceName
+				delete request.odataQuery.property
+				request.method = 'GET'
+				request.custom.isAction = 'canAccess'
 			addPermissions(req, request)
+		PRERESPOND: ({ request, data }) ->
+			if (request.custom.isAction == 'canAccess')
+				if (_.isArray(data) and data.length == 1)
```
Page-: Why is it an error to be able to access more than one instance of a resource?

afitzek: I guess there should not be more then one instance, although it might be empty.

src/sbvr-api/permissions.coffee
afitzek: I thought this makes more sense, in case we ever want to extend this to get all devices one can access for a specific action, in that case all returned `ids` are allowed.

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -724,6 +724,8 @@ constructError = (e) ->
 		{ status: 500 }
 	.catch UnsupportedMethodError, (err) ->
 		{ status: 405 }
+	.catch permissions.MissingOrInvalidArgumentsError, (err) ->
```
Page-: This seems to have the same meaning as `uriParser.BadRequestError`, can you at least merge them to one catch block, or even better use a single error for both - in future it'd be nice for hooks to be able to throw a `uriParser.BadRequestError` in order to get a 400 response rather than the default ..404? error, and having multiple equivalent errors would be confusing imo

src/sbvr-api/permissions.coffee
```diff
@@ -124,7 +125,31 @@ exports.setup = (app, sbvrUtils) ->
...
+				request.resourceName = request.resourceName.slice(0, request.resourceName.length - '#canAccess'.length)
+				request.odataQuery.resource = request.resourceName
+				delete request.odataQuery.property
+				request.method = 'GET'
+				request.custom.isAction = 'canAccess'
 			addPermissions(req, request)
+		PRERESPOND: ({ request, data }) ->
+			if (request.custom.isAction == 'canAccess')
+				if (_.isArray(data))
+					for i in [0..data.length - 1]
```
Page-: This would be nicer as
```coffee
for v, i in data
	data[i] = _.pick(v, 'id', '__metadata')
```

src/sbvr-api/permissions.coffee
```diff
@@ -124,7 +125,31 @@ exports.setup = (app, sbvrUtils) ->
...
+						data[i] = _.pick(data[i], 'id', '__metadata')
```
Page-: What fields are you excluding here btw? As the `$select` you add should already cover this I thought?

src/sbvr-api/permissions.coffee
```diff
@@ -124,7 +125,31 @@ exports.setup = (app, sbvrUtils) ->
 		POSTPARSE: ({ req, request }) ->
 			# If the abstract sql query is already generated then adding permissions will do nothing
 			return if request.abstractSqlQuery?
+			if (request.method == 'POST' and request.odataQuery.property?.resource == 'canAccess')
+				request.odataQuery.options = { '$select': { 'properties': [ { 'name': 'id' } ] }, $top: 1 }
```
Page-: This should really use the true id field (which may not be called `id`) via
```coffee
abstractSqlModel = sbvrUtils.getAbstractSqlModel(request)
abstractSqlModel.tables[request.resourceName].idField
```
once https://github.com/resin-io/pinejs/pull/106/files is merged

src/sbvr-api/permissions.coffee
```diff
@@ -124,7 +125,31 @@ exports.setup = (app, sbvrUtils) ->
...
+				{ action, method } = request.values
+				if method? and action?
+					throw new MissingOrInvalidArgumentsError()
+				else if method? and methodPermissions[method]?
+					request.permissionType = methodPermissions[method]
+				else if action?
+					request.permissionType = action
+				else
+					throw new MissingOrInvalidArgumentsError()
+				request.resourceName = request.resourceName.slice(0, request.resourceName.length - '#canAccess'.length)
```
Page-: This would be better as
```coffee
request.resourceName.slice(0, -'#canAccess'.length)
```
to avoid the need for fetching `request.resourceName.length`

src/sbvr-api/permissions.coffee
afitzek: I'm concerned, that with an `expand` linked resources could be fetched. Although I was not able to build a query for this, so that it fetches sub resources. But since I'm not certain that's why I added this filter.

src/sbvr-api/permissions.coffee
afitzek: Yeah that is a good point. When will #106 be merged? Should I wait for it?

src/sbvr-api/permissions.coffee
```diff
@@ -124,7 +125,42 @@ exports.setup = (app, sbvrUtils) ->
...
+				else if action?
+					request.permissionType = action
+				else
+					throw new BadRequestError()
+				abstractSqlModel = sbvrUtils.getAbstractSqlModel(request)
+				request.resourceName = request.resourceName.slice(0, -'#canAccess'.length)
+				resourceName = sbvrUtils.resolveSynonym(request)
+				resourceTable = abstractSqlModel.tables[resourceName]
+				if !resourceTable?
+					console.error('>>>> Tables: ', abstractSqlModel.tables)
```
Page-: Leftover debug

package.json
```diff
@@ -92,4 +92,4 @@
     "node": ">=4.0.0",
     "npm": ">=3.0.0"
   }
-}
\ No newline at end of file
+}
```
Page-: @hedss what's the story on fixing this? I thought it had already been done ages back

PR #99, opened 2017-12-13
Fix batch validation

nazrhom:
~~Save request specific hooks on the request object so they don't get mixed up in batch requests.~~
Defer validation of requests inside a changeset until the changeset has commited. This is achieved by marking requests with a `_deferValidation` flag. If another request in the changeset refers to a previous one via its `Content-ID` and updates some fields, validation will only be run on the updated fields.

This means we can create resources which temporarily violate rules inside a changeset, as long as a subsequent request inside the changeset will refer to the violating resource via the `$` syntax and fix any eventual rule violations.

Edit: First part got split into independent PR (https://github.com/resin-io/pinejs/pull/101) as it is blocking for another [PR](https://github.com/resin-io/resin-api/pull/678)

Change-type: patch

### DIFF
--- a/src/sbvr-api/sbvr-utils.coffee
+++ b/src/sbvr-api/sbvr-utils.coffee
@@ -168,7 +168,8 @@ isRuleAffected = do ->
 		# If for some reason there are no referenced fields known for the rule then we just assume it may have been modified
 		if not rule.referencedFields?
 			return true
-		modifiedFields = AbstractSQLCompiler.getModifiedFields(request.abstractSqlQuery)
+		modifiedFields = request.modifiedFields
+
 		# If we can't get any modified fields we assume the rule may have been modified
 		if not modifiedFields?
 			console.warn("Could not determine the modified table/fields info for '#{request.method}' to #{request.vocabulary}", request.abstractSqlQuery)
@@ -179,6 +180,8 @@ isRuleAffected = do ->
 
 
 exports.validateModel = validateModel = (tx, modelName, request) ->
+	if request?._deferValidation
+		return Promise.resolve()
 	Promise.map sqlModels[modelName].rules, (rule) ->
 		if not isRuleAffected(rule, request)
 			# If none of the fields intersect we don't need to run the rule! :D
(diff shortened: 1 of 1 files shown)


src/sbvr-api/sbvr-utils.coffee
```diff
@@ -768,18 +785,42 @@ runChangeSet = (req, res, tx) ->
...
+					env.validation.set(id, validation)
+					uriParser.parseId(result.body.id)
 			else
 				[tag, id]
 		request.sqlQuery = memoizedCompileRule(request.abstractSqlQuery)
+		request.modifiedFields = modifiedFields
 	return request
 
+
+# Removes from oldFields every field appearinf in newFields
```
thgreasi: typo: `appearing`

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -768,18 +785,42 @@ runChangeSet = (req, res, tx) ->
...
+invalidate = (oldFields, newFields) ->
+	if _.isArray(oldFields)
+		return _.map(oldFields, _.partial(invalidate, _, newFields))
+	else if _.isArray(newFields)
+		return _.reduce(newFields, invalidate, oldFields)
+	else if _.isUndefined(newFields)
```
thgreasi: I guess that this could be the first thing to check.
Also I guess that all these could be `return ... if ...` statements, which would remove all the `else`s, but I'm not sure which we prefer.

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -179,6 +180,8 @@ isRuleAffected = do ->
 
 
 exports.validateModel = validateModel = (tx, modelName, request) ->
+	if request?._deferValidation
```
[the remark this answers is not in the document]
nazrhom: Yes, we use this function to validate models on startup.

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -659,15 +662,25 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
...
 					return request
 
 			.then (request) ->
 				# Run the request in its own transaction
 				runTransaction req, request, (tx) ->
 					if _.isArray request
-						env = new Map()
+						env =
+							results: new Map()
+							validations: new Map()}
```
dfunckt: Does this even compile?

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -659,15 +662,25 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
...
 					if _.isArray request
-						env = new Map()
+						env =
+							results: new Map()
+							validations: new Map()
 						Promise.reduce(request, runChangeSet(req, res, tx), env)
-						.then (env) -> Array.from(env.values())
+						.tap (env) ->
+							validations = Array.from(env.validations.values())
+							Promise.map validation, (requestToValidate) ->
```
Page-: I think this whole PR can be made simpler and more performant by just adding on to the modified fields, so you end up with all of the tables and fields that the changeset modfies and then run the validation once which will still run all potentially affected rules and also guarantee that any given rule will only be run once per changeset

PR #112, opened 2018-01-12
Delete hooks from the external req object after setting them on the internal request. 

nazrhom:
This prevents hooks by being collected twice in runHook


src/sbvr-api/sbvr-utils.coffee
```diff
@@ -670,7 +670,9 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
 		mapSeries body, (bodypart) ->
 			uriParser.parseOData(bodypart)
 			.then controlFlow.liftP (request) ->
-				# Get the full hooks list now that we can
+				# Get the full hooks list now that we can. We clear the hooks on
+				# the global req to avoid duplication
+				req.hooks = []
```
Page-: This should be
```coffee
req.hooks = {}

Issue #117, opened 2018-01-26
BadRequestErrors should propagate their message body


thgreasi:
When a BadRequestError is thrown from within a hook with a specific error message, the response that the client receives doesn't contain the message in its body.

See: https://github.com/resin-io/resin-api/pull/711#discussion_r164041377

thgreasi:
Resolved by #118

PR #118, opened 2018-01-26
Propagate the error message of BadRequestErrors

thgreasi:
Connects-To: #117
Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #117


flesler:
Can we expose the message of any other errors, or even all of them in this PR? I found this situation myself, I see no reason to propagate the error messages, you can always control what you do with it from the caller code

thgreasi:
@flesler I pushed a new commit that exposes the error messages of all 4xx errors.
Decided to not expose the 5xx errors, since you might not always able to control the error returned (eg when the error occurs on a pine resource request) and I thought that in that case you probably don't want to propagate messages that might expose implementation internals (eg `SqlCompilationError`).
Let's see whether @page agrees.

cc: @CameronDiver @afitzek

Issue #116, opened 2018-01-26
Pine should export its TypedErrors


thgreasi:
Currently if we want to throw a specific pine error inside a hook we need to do something like:
```
{ BadRequestError } = require('@resin/pinejs/out/sbvr-api/uri-parser')
```
It would be great if we could do something like:
```
{ BadRequestError } = require('@resin/pinejs')
```
This would allow throwing errors of specific type & code from withing hooks without having to worry whether the nested path (`@resin/pinejs/out/sbvr-api/uri-parser`) will continue to export the errors in the future.

See: https://github.com/resin-io/resin-api/pull/711#discussion_r163133579

thgreasi:
Resolved by #119

PR #119, opened 2018-01-26
Export sbvr-api errors

thgreasi:
Connects-To: #116
Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #116


thgreasi:
I'm not sure whether this is a good approach, but most of the resin modules I've used have a separate file to declare all errors. Used this approach here as well, hoping it will give better typings out of the box. Kept the old exports as well, in case any consume project is doing deep imports, but we might want to change that in the next major release.

src/sbvr-api/permissions.coffee
```diff
@@ -3,14 +3,16 @@ Promise = require 'bluebird'
...
 env = require '../config-loader/env'
 userModel = require './user.sbvr'
-{ metadataEndpoints, BadRequestError, resolveSynonym } = require './uri-parser'
+{ metadataEndpoints, resolveSynonym } = require './uri-parser'
+{ BadRequestError } = require './errors'
 { ODataParser } = require '@resin/odata-parser'
 memoize = require 'memoizee'
 TypedError = require 'typed-error'
 sbvrUtils = require './sbvr-utils'
+{ PermissionError, PermissionParsingError } = require './errors'
```
nazrhom: NIT: Can we join the two `error` requires to a single one here?

thgreasi: Oh my... I really don't see how I missed that... Changing now

src/server-glue/module.ts
```diff
@@ -31,6 +31,13 @@ const configLoader = require('../config-loader/config-loader')
 const migrator = require('../migrator/migrator')
 
 export const sbvrUtils = require('../sbvr-api/sbvr-utils')
+export const {
```
Page-: Can you export these from `sbvr-utils` instead please, since pinejs modules get called with `sbvrUtils` as one of their arguments and so could access the errors directly from there

thgreasi: @Page- just to confirm how you intend this to work, consumer projects would do 
```
{ sbvrUtils: { BadRequestError } } = require '@resin/pinejs'
```
?

Or would we prefer `require('@resin/pinejs').sbvrUtils.errors.BadRequestError`?

PR #123, opened 2018-02-22
Add an `onlyid` option to POST requests…

Page-:
...which when set will return only the id in the response and save fetching the entire result

Ideas on  a better name would be appreciated as well, as I'm not sold on `onlyid`

Change-type: minor


CameronDiver:
+1 on `returnResource: false`

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -898,10 +898,13 @@ respondPost = (req, res, request, result, tx) ->
 	id = result
 	location = odataResourceURI(vocab, request.resourceName, id)
 	api[vocab].logger.log('Insert ID: ', request.resourceName, id)
-	runURI('GET', location, null, tx, req)
-	.catch ->
+	Promise.try ->
+		onlyId = d: [{ id }]
+		if _.get(request, [ 'odataQuery', 'options', 'onlyid' ])
```
flesler: 2 things:
- Wouldn't it be better to camel case into `'onlyId'`?
- Using the array version rather than `'odataQuery.options.onlyid'` is for improved performance? Otherwise the string version seems "cleaner", easier to read

flesler:
@pimterry I like `fetchResource` better if we go that way. Seems accurate.

src/sbvr-api/sbvr-utils.coffee
Page-: For camelCase it means the request also needs to be made as camel case, which is a little awkward

As for the array version, it should be faster as the string version is converted to the array version internally anyway, but I also find it preferable as it also works much nicer with variable paths eg `[ foo, bar, 'baz' ]`.

PR #126, opened 2018-03-02
Memoise getting hooks for the request

Page-:
Change-type: patch


src/sbvr-api/sbvr-utils.coffee
```diff
@@ -985,6 +993,9 @@ exports.addHook = (method, apiRoot, resourceName, callbacks) ->
 		resourceHooks[callbackType] ?= []
 		resourceHooks[callbackType].push(callback)
 
+	getHooks.clear()
```
flesler: so if I get this right, this is just in case someone adds a hook after some requests went through the stack to clear the catch? So, in principle, shouldn't get to here again after the server initialization is complete right?

Page-: Yeah, as a convention hooks are only added during initial setup but it is possible to add them later so this `clear` just helps to guarantee correct behaviour in those cases

PR #127, opened 2018-03-02
Move search path setting to be run once in a pool hook on client creation, rather than for each transaction

Page-:
Change-type: patch


Page-:
It's much lower overhead than real queries, but it's easy to save it :)

PR #128, opened 2018-03-02
Memoise checking the permissions that apply to the current request

Page-:
Change-type: patch


src/sbvr-api/permissions.coffee
```diff
@@ -209,8 +209,9 @@ exports.setup = (app, sbvrUtils) ->
 			resource: 'permission'
 			passthrough: req: rootRead
 			options:
-				select: 'name'
-				filter: permsFilter
+				$select: 'name'
+				$filter: permsFilter
+				$orderby: name: 'asc'
```
nazrhom: Why are we ordering here?

Page-: Yeah it's to increase hit rate for memoisation whilst keeping the cost of ordering minimal - I'll add a comment

Page-: done

PR #130, opened 2018-03-02
Only create errors for nice stack traces in debug mode due to the overhead it causes

Page-:
Change-type: minor


src/database-layer/db.ts
```diff
@@ -127,14 +130,31 @@ if (process.env.TRANSACTION_TIMEOUT_MS) {
 } else {
 	timeoutMS = 10000
 }
+
+const rejectedExecuteSql = memoize(
+	(message: string): Tx['executeSql'] =>
+		(_sql, _bindings, callback) =>
+			Promise.reject(new Error(message)).nodeify(callback),
```
thgreasi: If I get it right, the principle is to only create `new Error()`s once per error message.
But, the memoized function returns high order function that will call `new Error()` in each call.
Should we use a closure for the rejected error?
Am I missing something?

src/database-layer/db.ts
```diff
@@ -196,14 +216,21 @@ export abstract class Tx {
...
+			if (DEBUG) {
+				// In debug mode we create the error here to give useful stack traces,
+				// but it adds significant overhead for a production environment
+				const rejectionValue = new Error(message)
+				// We return a new rejected promise on each call so that bluebird can handle
+				// logging errors if the rejection is not handled (but only if it is not handled)
+				this.executeSql = (_sql, _bindings, callback) => {
+					return Promise.reject(rejectionValue).nodeify(callback)
+				}
+				this.rollback = this.end = (callback) => {
```
thgreasi: Since the code looks identical, how about moving the `if (DEBUG)` inside `rejectedExecuteSql()` & `rejectedRollbackEnd()` and decide in there whether a memoized version should be used or not, based on the environment?

src/database-layer/db.ts
```diff
@@ -196,14 +216,21 @@ export abstract class Tx {
...
+					return Promise.reject(rejectionValue).nodeify(callback)
+				}
+			} else {
+				this.executeSql = rejectedExecuteSql(message)
+				this.rollback = this.end = rejectedRollbackEnd(message)
```
thgreasi: Would it make sense to combine the two methods into one (returning a composite object), that could possibly reuse the created error (similarly to what we do in `if (DEBUG)`)?

src/database-layer/db.ts
```diff
@@ -127,14 +130,31 @@ if (process.env.TRANSACTION_TIMEOUT_MS) {
...
 	timeoutMS = 10000
 }
+
+const rejectedExecuteSql = memoize(
+	(message: string): Tx['executeSql'] =>
+		(_sql, _bindings, callback) =>
+			Promise.reject(new Error(message)).nodeify(callback),
+	{ primitive: true }
+)
+const rejectedRollbackEnd = memoize(
```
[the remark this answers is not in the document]
Page-: There's two possible messages right now

src/database-layer/db.ts
Page-: Using a closure for it would make sense actually and I'll make that change - in reality though the difference should be minimal as it should be a very rare error case to actually call one of these functions after the transaction has been closed

Page-: Actually thinking about it a bit more I think it makes sense as-is, it would be better performance to closure-scope the error but only in error cases which shouldn't be happening in production mode anyway, and closure-scoping it would make the stack trace completely useless (whereas it would still be useful if it's created on-demand)

src/database-layer/db.ts
Page-: Done

src/database-layer/db.ts
thgreasi: So the whole point of memoize that reduces the overhead is to create those two functions less often? I thought that the `new Error()` was the costly operation.

src/database-layer/db.ts
Page-: Done

src/database-layer/db.ts
Page-: Yeah, `new Error()` is pretty costly (although `err.stack` is even more costly the first time it's called).  The memoisation was just to avoid constantly recreating lambdas but it really shouldn't matter in this case - I'll benchmark it tomorrow and probably end up removing it

Page-: Removing it doesn't seem to matter so I've done that

PR #135, opened 2018-03-03
Memoise basic odata queries.

Page-:
This lets us get the benefits in the simple/highly repeated cases where there's a high hit rate without paying the cost to `_.cloneDeep` for every query where there's low/no cache hits due to variable sections

Change-type: minor


src/sbvr-api/uri-parser.coffee
```diff
@@ -83,11 +100,11 @@ parseODataChangeset = (env, b) ->
 
 	if b.url[0] == '/'
 		{ url, apiRoot } = splitApiRoot(b.url)
-		odata = odataParser.matchAll(url, 'Process')
+		odata = memoizedParseOdata(url, 'Process')
```
afitzek: Do we need the `'Process'` argument here? Not sure where this ends up.

src/sbvr-api/uri-parser.coffee
```diff
@@ -83,11 +100,11 @@ parseODataChangeset = (env, b) ->
...
 		defer = false
 	else
 		url = b.url
-		odata = odataParser.matchAll(url, 'Process')
+		odata = memoizedParseOdata(url, 'Process')
```
nazrhom: Ditto

src/sbvr-api/uri-parser.coffee
Page-: Yep, that's a leftover from before - removed

PR #139, opened 2018-03-21
Unify information stored on the express request object with the infor…

nazrhom:
…mation we obtain parsing the request url/body.

Nothing changes when dealing with a single request, the objects are now
consistent when responding to a batch request.

Change-type: patch


src/sbvr-api/sbvr-utils.coffee
```diff
@@ -662,6 +661,13 @@ exports.getAbstractSqlModel = getAbstractSqlModel = (request) ->
 	request.abstractSqlModel ?= abstractSqlModels[request.vocabulary]
 	return request.abstractSqlModel
 
+cloneReq = (req, odata) ->
+	req.method = odata.method
```
Page-: This modifies the original `req` object which means the logging will show the request as being the first of the batch requests rather than as being a batch request

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -662,6 +661,13 @@ exports.getAbstractSqlModel = getAbstractSqlModel = (request) ->
...
+	req.url = odata.url
+	# set values back on odata body
+	req.body = odata.values
+	_.cloneDeep(req)
```
Page-: `cloneDeep` on the `req` object could potentially be pretty costly and a real performance hit

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -688,12 +695,11 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
 		mapSeries body, (bodypart) ->
 			uriParser.parseOData(bodypart)
 			.then controlFlow.liftP (request) ->
-				# Get the full hooks list now that we can. We clear the hooks on
-				# the global req to avoid duplication
-				req.hooks = {}
+				req = cloneReq(req, request)
```
Page-: Since we overwrite `req` each time it means that any modifications made to it by previous requests in the batch will be carried forward

nazrhom:
@Page- The reason is that I was thinking that the information we might want to pick from `req` is something that evolves, we might want to rely on other features later and would have to modify this code. On the other hand the information we want to set the other way is essentially limited to what we learn on the OData parse, so method, url and body (+ we could also add some extra logic for headers as pablo was interested in doing batch of requests with different api keys). + it makes the change less invasive as we don't have to modify all the hooks

petrosagg (commented):
No-op review to dismiss stale review request

PR #142, opened 2018-03-23
Add rollback function for hooks that perform side-effects.

nazrhom:


src/sbvr-api/sbvr-utils.coffee
```diff
@@ -353,17 +357,65 @@ getHooks = do ->
...
 	Promise.map hooks, (hook) ->
-		hook(args)
+		args.registerRollback = (args...) ->
+			for arg in args
+				hook.setArgument(arg)
+		hook.run(args)
+
+# The execution order of rollback actions is unspecified
+undoHooks = (request) ->
+	Promise.map _.flatMap(request.hooks, _.identity), (hook) ->
```
Page-: I think this should be `_.flatten(request.hooks)`

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -353,17 +357,65 @@ getHooks = do ->
...
+			@rollbackFn.apply(null, @rollbackArgs)
+
 runHooks = (hookName, args) ->
 	Object.defineProperty args, 'api',
 		get: _.once ->
 			return api[args.request.vocabulary].clone(passthrough: _.pick(args, 'req', 'tx'))
 	hooks = args.request?.hooks?[hookName] || []
 	Promise.map hooks, (hook) ->
-		hook(args)
+		args.registerRollback = (args...) ->
```
Page-: This seems like it would make more sense to just be:
```coffee
args.registerRollback = hook.registerRollback
```
and have the hook class handle it directly.  Alternatively it should be possible to have hooks be able to run `this.registerRollback()` and have them run in the context of the hook obj

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -353,17 +357,65 @@ getHooks = do ->
...
+		# The ROLLBACK function should take care of checking and performing the correct number of rollback actions.
+		.tapCatch =>
+			if not _.isEmpty(@rollbackArgs)
+				missingArgs = @rollbackFn.length - @rollbackArgs.length
+				for i in _.range(missingArgs)
+					@rollbackArgs.push(null)
+				@rollback()
+
+	rollback: ->
+		Promise.try =>
```
Page-: I think this can be:
```coffee
rollback: Promise.method ->
	@rollbackFn(@rollbackArgs...)

File: src/sbvr-api/sbvr-utils.coffee [hunk 4]
```diff
@@ -353,17 +357,65 @@ getHooks = do ->
...
+
+	setArgument: (arg) ->
+		if @rollbackArgs.length < @rollbackFn.length
+			@rollbackArgs.push(arg)
+		else
+			throw new Error('Too many arguments supplied to rollback function')
+
+	run: (args...) ->
+		Promise.try =>
+			@runFn.apply(null, args)
```
Page-: This can be
```coffee
@runFn(args...)

File: docs/Hooks.md [hunk 5]
```diff
@@ -1,16 +1,60 @@
...
+```
+
+The hook will now be passed the `registerRollback` function which can be used to store any information we need to perform the rollback (e.g. the ID of a resource we must delete). We can use registerRollback to supply these arguments, which will be later passed to the supplied `ROLLBACK` function if we need to undo the side-effects of the hook.
+`registerRollback` may be called once with all the arguments we intend to supply, or multiple times by passing each argument as soon as it becomes available. The arguments will be passed to the `ROLLBACK` function in the order they have been received.
+
+We can take advantage of this by writing the `ROLLBACK` function so that we can perform partial rollback of our side-effects, as such:
+```coffee
+PHASE:
+	RUN: ({ ..., registerRollback }) ->
+		createResource(1)
```
Page-: It might be worth noting that this is more a case of `createExternalResource` because if you use the hook's api/tx objects to do the creation then it will automatically be rolled back

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -721,6 +780,8 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
 					else
 						{ req, request } = requestEnv
 						runRequest(req, res, tx, request)
+				.tapCatch ->
```
Page-: This needs to be unindented a bunch, because right now it's won't roll back PREPARSE/POSTPARSE hooks if the failure happens during translation/parsing/etc

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -995,7 +1056,15 @@ exports.executeStandardModels = executeStandardModels = (tx, callback) ->
 		throw err
 	.nodeify(callback)
 
-exports.addHook = (method, apiRoot, resourceName, callbacks) ->
+exports.addHook = (method, apiRoot, resourceName, hooks) ->
+	pureHooks = _.mapValues hooks, (hook) ->
+		{
+			RUN: hook
```
Page-: It might be better to call this `HOOK` to avoid any possible confusion with `PRERUN`/`POSTRUN`

src/sbvr-api/sbvr-utils.coffee
nazrhom: Yeah, I did it like this as initially registerRollback would need a reference to the current `request`. I like `@registerRollback()` in the hooks more than setting it on the args tbh.

src/sbvr-api/sbvr-utils.coffee
nazrhom: `flatten` does not work here as `request.hooks` is an object like `{ PHASE: [hooks] }`. The current behaviour is  equivalent to `_.flatten(_.values(request.hooks))`

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -687,34 +765,38 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
...
-						.then (env) -> Array.from(env.values())
-					else
-						runRequest(req, res, tx, request)
+			.then (odata) ->
+				requestEnv = prepareEnv(req, odata)
+				Promise.resolve(requestEnv)
+				.then controlFlow.liftP (requestEnv) ->
+					{ req, request } = requestEnv
+					# Get the full hooks list now that we can, we overwrite the previous list to avoid duplication
+					request.hooks = getHooks(request)
```
nazrhom: There is an issue atm when refreshing the hook list, since we now rely on the executed flag to be set on the hook to rollback. 
This is not a problem as there is no PREPARSE hook that does any side effect action, nor the probably should ever be. Maybe we can prevent a PREPARSE side-effect hook from ever being created by checking in `addSideEffectHook` just to be sure.

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -687,34 +765,38 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
...
-				runTransaction req, (tx) ->
-					if _.isArray request
-						env = new Map()
-						Promise.reduce(request, runChangeSet(req, res, tx), env)
-						.then (env) -> Array.from(env.values())
-					else
-						runRequest(req, res, tx, request)
+			.then (odata) ->
+				requestEnv = prepareEnv(req, odata)
+				Promise.resolve(requestEnv)
```
Page-: It would be much nicer to do this as:
```
.then (odata) ->
	prepareEnv(req, odata)
.then controlFlow.liftP (requestEnv) ->

File: src/sbvr-api/sbvr-utils.coffee [hunk 10]
```diff
@@ -353,18 +356,65 @@ getHooks = do ->
...
+		# The ROLLBACK function should take care of checking and performing the correct number of rollback actions.
+		.tapCatch =>
+			if not _.isEmpty(@rollbackArgs)
+				missingArgs = @rollbackFn.length - @rollbackArgs.length
+				for i in _.range(missingArgs)
+					@rollbackArgs.push(null)
+				@rollback()
+
+	registerRollback: (args...) ->
+		for arg in args
```
Page-: I think this would be better as:
```coffee
registerRollback: (@rollbackArgs...) ->
	if @rollbackArgs.length > @rollbackFn.length
		throw new Error('Too many arguments supplied to rollback function')
```
and remove `setArgument` altogether.  Although that said the length check is a bit awkward altogether, eg default values don't count towards the function length but are still valid to supply:
```js
foo = (a = 'b') => {}
foo.length // 0
```

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -353,18 +356,65 @@ getHooks = do ->
...
+			if @rollbackArgs.length != @rollbackFn.length
+				throw new Error('registerRollback not supplied enough arguments')
+			@executed = true
+		# If we detect that some rollbacks are registered but the rollback function is not saturated, we add the missing number of arguments as null.
+		# The ROLLBACK function should take care of checking and performing the correct number of rollback actions.
+		.tapCatch =>
+			if not _.isEmpty(@rollbackArgs)
+				missingArgs = @rollbackFn.length - @rollbackArgs.length
+				for i in _.range(missingArgs)
+					@rollbackArgs.push(null)
```
Page-: It would be better not to fill in the extra args as null, because as null it will still throw errors if you try doing something like `null.fn()` or `null.prop`, but it will also bypass es6 default arguments, eg check out:
```js
foo = (a = { b: 'c' }) =>
	console.log(a)
foo(null) // null
foo() // {b: "c"}
foo(undefined) // {b: "c"}
```
where supplying `null` instead of the expected `undefined` will cause different behaviour (coffeescript 1 defaults apply for both null/undefined, coffeescript 2 follows the es6 defaults)

src/sbvr-api/sbvr-utils.coffee
Page-: I think `_.flatten(_.values(request.hooks))` makes sense then, since it makes the intent clearer

src/sbvr-api/sbvr-utils.coffee
nazrhom: The idea of the `setArgument` is to make hooks with multiple side-effects less awkward. So we can re-use the supplied rollback action as partial cancellation of side-effects if something goes wrong in the hook.
In general rollback functions should not define default arguments, but check and only perform the cancellation actions for the supplied arguments.

src/sbvr-api/sbvr-utils.coffee
nazrhom: Maybe this is not that bad after all, default arguments break the `rollback` function already as they mess with the arity count. This would help in preventing default arguments from being used there as they will essentially never work.

src/sbvr-api/sbvr-utils.coffee
nazrhom: I agree, however I need to declare it to make sure that `requestEnv` is in scope for the final `tapCatch` block (which also forces me to indent the next two `then` blocks). 
I don't like this that much (I will at least add a comment there, as the meaning is not clear and it might be changed by mistake in the future) however I would like to change it if you have any idea on how to make it better

Page-: What about
```coffee
.then (odata) ->
	prepareEnv(req, odata)
.tap controlFlow.liftP (requestEnv) ->
```
then?

src/sbvr-api/sbvr-utils.coffee
Page-: Oh, that is definitely not how I was envisioning rollback, how do you handle the following case
```coffee
HOOK: ({ ..., registerRollback }) ->
	Promise.all([
		createExternalResource(1)
		.then (id1) ->
			registerRollback(id1)

		createExternalResource(2)
		.then (id2) ->
			registerRollback(id2)
	])
```
where the rollbacks can be registered in different orders, or
```coffee
HOOK: ({ ..., registerRollback }) ->
	Promise.try ->
		if test()
			createExternalResource(1)
			.then (id1) ->
				registerRollback(id1)
	.then ->
		# Additional logic
		createExternalResource(2)
		.then (id2) ->
			registerRollback(id2)
		.catch (err) ->
			# No need to delete resource 1 here
```
where one of the early creations might be skipped?
I was envisioning something much more like
```coffee
HOOK: ({ ..., registerRollback }) ->
	Promise.all([
		createExternalResource(1)
		.then (id1) ->
			registerRollback ->
				deleteExternalResource(id1)

		createExternalResource(2)
		.then (id2) ->
			registerRollback ->
				deleteExternalResource(id2)
	])
```
so that we just past a selection of functions to be run on rollback rather than relying on a fragile order of adding arguments

src/sbvr-api/sbvr-utils.coffee
nazrhom: This still doesn't work as expected. If there is an error in the `tap` block, then `requestEnv` will be undefined in the in `tapCatch` that comes after. We need to declare it as a variable that exists outside the promise chain that does the compilation and execution in order to be able to refer to it in the `tapCatch` block.

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -353,18 +358,59 @@ getHooks = do ->
...
+		if @rolledBack
+			runFn()
+		else
+			@rollbackFns.push(runFn)
+
+	rollback: ->
+		# set rolledBack to true straight away, so that if any rollback action
+		# is registered after the rollback call, we will immediately execute it
+		@rolledBack = true
+		Promise.each(@rollbackFns, (fn) -> fn())
```
Page-: `(fn) -> fn()` can be replaced by `_.attempt`

Page-: I think this should be a `Promise.map`, otherwise if one of the rollbacks fails then none of the ones after it will be run

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -353,18 +358,59 @@ getHooks = do ->
...
+		.finally =>
+			@executed = true
+
+class SideEffectHook extends Hook
+	constructor: (hookFn) ->
+		@rollbackFns = []
+		@rolledBack = false
+		super(hookFn)
+
+	registerRollback: (fn) ->
```
Page-: You should add a `return` at the end of this function, right now it can return either a promise or a number and it's possible that will cause the promise to be waited upon by the parent in some cases, eg
```coffee
Promise.try =>
	@registerRollback(fn)
```
will sometimes wait on the fn and sometimes not

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -353,18 +358,59 @@ getHooks = do ->
...
+		runFn = Promise.method(fn)
```
Page-: There's no need to wrap the `fn` in a `Promise.method` for `@rollbackFns` because the `Promise.each` in `rollback` will automatically handle the wrapping.  As for the direct call version it probably makes more sense to use `Promise.try(fn)` - can you check if `fn` will be run synchronously in those cases though, as we should make sure it's done async to have the same behaviour whether we've already rolled back or not

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -353,18 +358,59 @@ getHooks = do ->
...
-	hooks = hooks.concat(args.request?.hooks?[hookName] || [])
+	hooks = args.request?.hooks?[hookName] || []
 	Promise.map hooks, (hook) ->
-		hook(args)
+		hook.run(args)
+
+# The execution order of rollback actions is unspecified
+undoHooks = (request) ->
+	Promise.map _.flatten(_.values(request.hooks)), (hook) ->
+		if hook instanceof SideEffectHook and hook.executed
```
Page-: I don't think there's any need for this `executed` flag - even if the hook hasn't finished we still want to run any rollback hooks it's hit so far, and also enable the instant triggering of any it adds later

src/sbvr-api/sbvr-utils.coffee
Page-: What about
```coffee
.then (odata) ->
	requestEnv = prepareEnv(req, odata)
	controlFlow.liftP (requestEnv) ->
```
then? It'll still avoid the `Promise.resolve` at least

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -662,6 +708,33 @@ exports.getAbstractSqlModel = getAbstractSqlModel = (request) ->
 	request.abstractSqlModel ?= abstractSqlModels[request.vocabulary]
 	return request.abstractSqlModel
 
+prepareEnv = (httpReq, odata) ->
+	if _.isArray(odata)
+		_.map(odata, _.partial(cloneReq, httpReq))
+	else cloneReq(httpReq, odata)
+
+cloneReq = (httpReq, odata) ->
+	req =
```
Page-: This is a `major` change btw, so we'll want to end up with this being merged along with the 7.x PR - so it might be worth targeting that with this PR or making the `req` changes be separate (I'm not certain if they're a hard dependency of the side effect hooks).

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -341,6 +340,12 @@ getHooks = do ->
 			getResourceHooks(methodHooks[vocabulary], resourceName)
 			getResourceHooks(methodHooks['all'], resourceName)
 		)
+	instantiateHooks = (hooks) ->
+		_.mapValues hooks, (typeHooks) ->
+			_.map typeHooks, (hook) ->
+				if hook.effects
+					new SideEffectHook(hook.HOOK)
+				else new Hook(hook)
```
Page-: Could you put the `new Hook` on a new line please

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -353,18 +358,59 @@ getHooks = do ->
...
+				request.method
+				request.vocabulary
+				resourceName
+			)
+		)
 	_getHooks.clear = -> getMethodHooks.clear()
 	return _getHooks
 
-runHook = (hookName, args) ->
+class Hook
```
Page-: This is a 100% optional btw, but it would be nice to separate these classes out to a `hooks.ts` or similar - if not then it'll happen at a later date along with the rest of the ts conversion

Page-:
This was merged in 7.0.0

PR #147, opened 2018-03-29
Fix rethrowing a nicer db error.

Page-:
Change-type: patch


dfunckt:
(Removed the ready-to-merge label, VB didn’t pick it up anyway)

Page-:
This is a fix for https://github.com/resin-io/pinejs/pull/144 @dfunckt so it will go back to what it was before the most recent patch change - so it's an unbreaking change rather than breaking

PR #149, opened 2018-03-30
7.x major version

Page-:
All dependencies are merged, so just needs this now

* Change `db.transaction` to take a fn that after completing will automatically close/rollback the transaction, handling cancellation nicely
* Cancel any ongoing work when the request/response is closed.
* Update to pinejs-client-core 5.x
* Fix prettifying db errors, making the thrown constraint errors a bit nicer again.
* Return a 401 when a POST fails to create any resources.
* Fix prettifying db errors, making the thrown constraint errors a bit nicer again.
* Generate es2015 output, dropping support for node 4.
* Change the `rows` property of a db result from a websql-like format to a standard array.
* Remove the ability to use `undefined` to generate a default value, as undefined can't exist in tthe JSON body anyway.
* Remove special handling for arrays in the pg layer, node-pg support array binds now.
* Update to abstract-sql-compiler which now generates native postgres bindings.
* Use `Object.freeze` to prevent mutating the cached abstract sql which allows us to skip a costly Object.freeze

Specifically the following 3 combined allow us to completely remove the need for sql-binds.ts completely which removes a lot of overhead for postgres db calls:
* Remove the ability to use `undefined` to generate a default value, as undefined can't exist in tthe JSON body anyway.
* Remove special handling for arrays in the pg layer, node-pg support array binds now.
* Update to abstract-sql-compiler which now generates native postgres bindings.


dfunckt:
> Cancel any ongoing work when the request/response is closed

What does this mean for any work that runs out of band, like geocoding? (Though I think that specific case doesn’t make any Pine requests afterwards, but we may have other functionality that does, maybe billing?)

Page-:
Work that runs out of band won't be cancelled by default, because they aren't part of the same promise chain that is cancelled, however they could make use of the hooks rollback mechanism https://github.com/resin-io/pinejs/pull/142 to either cancel or undo any change they make if the main request gets cancelled (not included in this PR yet - but it's in part the description that this won't be ready to merge until that's incorporated)

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -685,11 +685,15 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
 	mapSeries = controlFlow.getMappingFn(req.headers)
 	# Get the hooks for the current method/vocabulary as we know it,
 	# in order to run PREPARSE hooks, before parsing gets us more info
-	req.hooks = getHooks(
+	req.hooks = reqHooks = getHooks(
```
nazrhom: NIT: Maybe we can call these PREPARSE hooks directly.

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -685,11 +685,15 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
...
 		method: req.method
 		vocabulary: apiRoot
 	)
-	runHooks('PREPARSE', { req, tx: req.tx })
+
+	req.tx?.on 'rollback', ->
```
nazrhom: This is only needed if the `tx` is created in a hook right? If we are in an internal request shouldn't the rollback handler be already setup from the runTransaction callback (lines 729-731)?

Page-: The rollback handler for the top level request would be set up there, but internal requests have their own sets of hooks that need to be rolled back and this attaches their rolling back to happen as part of the parent's tx rolling back

src/sbvr-api/permissions.coffee
```diff
@@ -167,6 +170,99 @@ _checkPermissions = (permissions, actorID, actionList, resourceName, vocabulary)
...
+		if selectField.length is 2 and _.isArray(selectField[0])
+			return selectField[0]
+		return selectField
+
+	return { extraBinds: odataBinds, abstractSqlQuery }
+
+# Call the function once and either return the same result or throw the same error on subsequent calls
+onceGetter = (obj, propName, fn) ->
+	thrownErr = undefined
+	Object.defineProperty obj, 'definition',
```
afitzek: Should this read `Object.defineProperty obj, propName,`?

PR #151, opened 2018-04-05
Switch to using the array form for _.get, improving performance in a hot function.

Page-:
This is mainly a large benefit because this is a hot function that's called on every single db query (of which there are a fair few for every api query), in general the difference probably wouldn't matter but in this case it drops from 146ms to 9.2ms


src/database-layer/db.ts
```diff
@@ -293,7 +293,7 @@ if (maybePg != null) {
 					map: (iterator, thisArg) => rows.map(iterator, thisArg),
 				},
 				rowsAffected: rowCount,
-				insertId: _.get(rows, '0.id'),
+				insertId: _.get(rows, [ 0, 'id' ]),
```
CameronDiver: Would it not be even quicker to not rely on a function call at all?

Page-: Possibly but it'd be a lot messier and only save a couple of % at most I expect, as opposed to the ~94% improvement of this change

PR #154, opened 2018-04-11
WIP: translations

nazrhom:
This PR is a WIP to move translations into pine


src/translations/index.coffee
```diff
@@ -0,0 +1,30 @@
+_ = require 'lodash'
```
Page-: New files should be in typescript btw

PR #155, opened 2018-04-12
Rollback hooks 7.x

nazrhom:


src/sbvr-api/sbvr-utils.coffee
```diff
@@ -703,7 +717,7 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
 				req.hooks = {}
```
nazrhom: We can avoid the `Promise.resolve()` in the lines above by duplicating the `tapCatch`

src/sbvr-api/hooks.ts
```diff
@@ -0,0 +1,69 @@
...
+			this.rollbackFns.push(fn)
+		}
+		return
+	}
+
+	rollback() {
+		// set rolledBack to true straight away, so that if any rollback action
+		// is registered after the rollback call, we will immediately execute it
+		this.rolledBack = true
+		return controlFlow.settleMapSeries(this.rollbackFns, _.attempt)
```
nazrhom: I think we should use `settleMapSeries` as it makes sense to run them sequentially, this should prevent the problem with `each`

src/sbvr-api/hooks.ts
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+
+// The execution order of rollback actions is unspecified
+const undoHooks = function (request: any) {
+	return controlFlow.settleMapSeries(_.flatten(_.values(request.hooks)), function(hook) {
```
Page-: I think it's nicer to use:
```typescript
(hook) => {
```

src/sbvr-api/hooks.ts
Page-: Can you add a `.return()`, it'll make the return type `Promise<void>`

src/sbvr-api/hooks.ts
```diff
@@ -0,0 +1,69 @@
...
+		if (hook instanceof SideEffectHook) {
+			return hook.rollback()
+		} else {
+			return Promise.resolve() as Promise<any>
```
Page-: With the change to the `rollback` function you'll be able to get rid of this cast

src/sbvr-api/hooks.ts
```diff
@@ -0,0 +1,69 @@
...
+
+exports.rollbackRequestHooks = function (request: any) {
+	if (_.isArray(request)) {
+		return controlFlow.settleMapSeries(request, undoHooks)
+	} else {
+		return undoHooks(request)
+	}
+}
+
+exports.Hook = Hook
```
Page-: You should use `export class ...` above instead for these

src/sbvr-api/hooks.ts
```diff
@@ -0,0 +1,61 @@
+import * as _ from 'lodash'
+import * as Promise from 'bluebird'
+import * as controlFlow from './control-flow'
+
+type RollbackAction = () => any
+
+export class Hook {
+	private hookFn: Function
+
+	constructor(hookFn: Function) {
```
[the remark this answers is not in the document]
nazrhom: This did not work for me, it complained about the misssing `super` call in a class extending a superclass

nazrhom:
@Page- I moved `instantiateHooks` to `hooks.ts` as I agree with separating the logic, however I think it makes sense to get them already instantiated after a `getHooks`. We really never want to do anything with non-instantiated hooks and having two different kinds of hooks popping up in the request is slightly more confusing imo.

docs/Hooks.md
```diff
@@ -1,16 +1,35 @@
 # Hooks
-Hooks are functions that you can implement in order to execute custom code when API calls are requested. The methods that are supported are `GET`, `POST`, `PUT`, `PATCH`. The sbvrUtils module of Pine.js is the mechanism that supports hooks addition. In particular, using `sbvrUtils.addHook`.
+Hooks are functions that you can implement in order to execute custom code when API calls are requested. The methods that are supported are `GET`, `POST`, `PUT`, `PATCH`. The sbvrUtils module of Pine.js is the mechanism that supports hooks addition. There are two kind of hooks that can be defined: side-effect hooks and pure hooks. These are respectively defined using `sbvrUtils.addSideEffectHook` and `sbvrUtils.addHook`.
```
Page-: You should update this to reference `addPureHook`

src/sbvr-api/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+type RollbackAction = () => any
+type HookFn = () => any
+type HookBluePrint = {
+	HOOK: HookFn
+	effects: boolean
+}
+export class Hook {
+	private hookFn: Function
+
+	constructor(hookFn: HookFn) {
```
Page-: Continuing from https://github.com/resin-io/pinejs/pull/155#discussion_r181603674 I think the issue is you tried to make the change to `SideEffectHook`, it should be fine to add to just `Hook` (and is the only place it's needed anyway as it replaces the `this.x = ...`)

src/sbvr-api/hooks.ts
```diff
@@ -0,0 +1,77 @@
...
+import { settleMapSeries } from './control-flow'
+
+type RollbackAction = () => any
+type HookFn = () => any
+type HookBluePrint = {
+	HOOK: HookFn
+	effects: boolean
+}
+export class Hook {
+	private hookFn: Function
```
Page-: This should be `hookFn: HookFn`

src/sbvr-api/hooks.ts
```diff
@@ -0,0 +1,77 @@
+import * as _ from 'lodash'
+import * as Promise from 'bluebird'
+import { settleMapSeries } from './control-flow'
+
+type RollbackAction = () => any
```
Page-: This should be `() => void | Promise<void>` since we don't use the return value (outside of waiting on a promise), and as such typing it as void will help avoid issues where people return other things by accident or because they expect them to do something

PR #158, opened 2018-04-23
Store api-key values in hashed format

afitzek:
This stores the api key values as sha256 hashes. The database values are prefixed with 'SHA256:HEX:' to indicate, which hash algorithm and which encoding are used.

Depends-On: https://github.com/resin-io-modules/sbvr-types/pull/14
Change-Type: major

### DIFF
--- a/src/sbvr-api/sbvr-utils.coffee
+++ b/src/sbvr-api/sbvr-utils.coffee
@@ -42,6 +42,8 @@ memoizedCompileRule = memoize(
 
 db = null
 
+exports.hashApiKey = permissions.hashApiKey
+
 exports.sbvrTypes = sbvrTypes
 
 fetchProcessing = _.mapValues sbvrTypes, ({ fetchProcessing }) ->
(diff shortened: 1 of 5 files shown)


flesler (approved):
The change looks OK but I really wonder if there are other people where this will impact, I expect some to be.

Page- (changes requested):
Can you add a template migration that can be used for the major upgrade

dfunckt:
@afitzek please let’s discuss what we’re going to do with the delta server before merging this, this is a serious blocker for the new deltas.

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -42,6 +42,8 @@ memoizedCompileRule = memoize(
 
 db = null
 
+exports.hashApiKey = permissions.hashApiKey
```
Page-: Everything from `permissions` is already exported below via an `_.extend(exports, permissions)`

petrosagg (commented):
No-op review to dismiss stale review request

PR #163, opened 2018-04-26
Add node 10 to the test matrix.

Page-:
This one looks like it'll have to wait a while for circleci to add node 10 images

Change-type: patch


circle.yml
```diff
@@ -72,6 +78,12 @@ workflows:
             tags:
               only: /.*/
       - node-8:
+          # Run for all tags (required to allow the deploy to trigger on version tags)
+          filters:
+            tags:
+              only: /.*/
+              only: /.*/
```
nazrhom: Do we need to repeat this twice?

Page-: We don't, I must have grabbed an extra line when copy/pasting

PR #168, opened 2018-05-03
Use payload property to return json structures for errors

afitzek:
Change-Type: minor

### DIFF
--- a/src/sbvr-api/errors.ts
+++ b/src/sbvr-api/errors.ts
@@ -1,5 +1,14 @@
 import TypedError = require('typed-error');
 
+export class PayloadError extends TypedError {
+	constructor(
+		name: string,
+		public payload: { [k: string]: any } | undefined
+	) {
+		super(name)
+	}
+}
+
 export class PermissionError extends TypedError {}
 export class PermissionParsingError extends TypedError {}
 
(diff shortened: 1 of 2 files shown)


src/sbvr-api/errors.ts
```diff
@@ -1,5 +1,14 @@
 import TypedError = require('typed-error');
 
+export class PayloadError extends TypedError {
+	constructor(
+		name: string,
+		public payload: { [k: string]: any } | undefined
```
thgreasi: NIT: `{ [k: string]: any }` is commonly referred as `AnyObject`, it might be useful make it a separate type.

src/sbvr-api/errors.ts
```diff
@@ -1,5 +1,14 @@
 import TypedError = require('typed-error');
 
+export class PayloadError extends TypedError {
+	constructor(
+		name: string,
```
thgreasi: This will end up in the `message` property right?
Is there a specific that this is the first arguments and not optional?

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -823,19 +824,22 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
 		console.error('An error occured while constructing the response', e, e.stack)
 		res.sendStatus(500)
 
+getErrorResponseBody = (e) ->
+	return e.payload or e.message
```
thgreasi: If the primary use is to be used as the response body, would it make sense to name it as `body`?

thgreasi:
@afitzek let's discuss this on Monday

afitzek:
Cool. This will be implemented in #172.

PR #172, opened 2018-05-04
Add typed errors for the primary http response statuses

thgreasi:
Resolves: #171
Change-Type: minor


thgreasi:
@flesler @pimterry  that was the way that the same thing is done in [resin-errors](https://github.com/resin-io-modules/resin-errors/blob/55a2f73eb1c59a0ff2786454882d961947f4b8b8/lib/errors.ts#L28) but I will give it a try.

src/sbvr-api/errors.ts
```diff
@@ -10,5 +10,237 @@ export class InternalRequestError extends TypedError {}
...
+	}
+}
+
+export class ExpectationFailedError extends HttpError {
+	constructor(error: string | Error = '') {
+		super(417, error);
+	}
+}
+
+export class IMATeapotError extends HttpError {
```
Page-: You skipped `402 Payment Required` I think you can skip this one as well safely :P

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -826,16 +826,18 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
...
 	.catch PermissionError, (err) ->
 		{ status: 401, body: err.message }
 	.catch SqlCompilationError, TranslationError, ParsingError, PermissionParsingError, InternalRequestError, (err) ->
 		{ status: 500 }
 	.catch UnsupportedMethodError, (err) ->
 		{ status: 405, body: err.message }
-	.catch ForbiddenError, (err) ->
-		{ status: 403, body: err.message }
+	.catch HttpError, (err) ->
+		if err.status >= 500
```
Page-: Is there any reason to drop the message for 500+? I feel like if a message has been attached then it's intentional? That said it's probably a sign that we should have a way to differentiate between internal error messages and ones we want to expose externally - and would also solve https://github.com/resin-io/pinejs/pull/168

thgreasi: Not strong opinon on this, I just saw that we already are not attaching the message for 500 and thought that it was on purpose (I guessed that it was to avoid exposing internal details). Will change this.

src/sbvr-api/errors.ts
thgreasi: Got the errors from https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
and missed 402 by accident (the html was different so I failed to extract that part).
Will add 402 and remove the teapot :+1:

thgreasi: Oh 402 is future reserved and not meant to be used atm!
Anyway just give me a note and can remove the 402

src/sbvr-api/errors.ts
```diff
@@ -10,5 +10,237 @@ export class InternalRequestError extends TypedError {}
...
+	}
+}
+
+export class UnavailableForLegalReasonsError extends HttpError {
+	constructor(error: string | Error = '') {
+		super(451, error);
+	}
+}
+
+export class InternalServerError extends HttpError {
```
Page-: There's an `InternalRequestError` which is the current equivalent, so it would be good to reuse it

src/sbvr-api/sbvr-utils.coffee
Page-: Specifically for the 500s where we ignore it's because they're unexpected and so they'll have an error message that could be dangerous to expose.  For explicitly thrown 500s then I think it's fine as they're expected and should only contain error messages that we're ok with showing to the user

src/sbvr-api/errors.ts
thgreasi: I was actually hesitant to reuse it, so that we don't have a breaking change where InternalRequestError suddenly starts passing it's message to the response.
After a quick search it seems that:
* In pine itself its only used once and w/o a message argument but it's exported.
* The API isn't using it at all (and I'm not aware of any other consumer).

Changing InternalRequestError to be an HttpError and dropping InternalServerError. Give me a note in case I misinterpreted the requested change here.

PS: TBH I liked the naming of InternalServerError more than InternalRequestError, it's more consistent with the http errors naming & other server libraries.

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -35,7 +34,8 @@ _.assign(exports, errors)
 	SqlCompilationError
 	TranslationError
 	UnsupportedMethodError
-	ForbiddenError
+	BadRequestError
+	HttpError
```
nazrhom: Should we export all the HTTPErrors now?

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -826,16 +826,16 @@ exports.handleODataRequest = handleODataRequest = (req, res, next) ->
 # Reject the error to use the nice catch syntax
 constructError = (e) ->
```
nazrhom: I wonder if we should change this so that we map internal errors to HTTPErrors and always respond with one of those

src/sbvr-api/sbvr-utils.coffee
thgreasi: We already do, there is a `_.assign(exports, errors)` right above :)

nazrhom: Ahhh, sorry missed that :D thought this was the list of the ones we would export

afitzek: I think they are all exported by `_.assign(exports, errors)`. This deconstructs the `errors` exports into the individual used once here.

src/sbvr-api/errors.ts
```diff
@@ -6,9 +6,257 @@ export class PermissionParsingError extends TypedError {}
...
+}
+
+export class HttpError extends TypedError {
+	body?: string | AnyObject
+
+	constructor(public status: number, error: string | Error = '') {
+		super(error)
+	}
+
+	setResponseBody(body: string | AnyObject) {
```
Page-: Any reason for having a separate setter rather than including it in the constructor? How are you envisioning this being used?

thgreasi: Like `throw new BadRequestError(err).setResponseBody({})` but can definitely also add that as a third (optional) parameter in the constructor (but then we will probably also have to update the constructors of of all `TypedError`s).

Page-: It would only need to update the constructors of the `HttpError`s here, and in most cases it would be a lot more fluid imo to just have the extra constructor param when calling

thgreasi: ok I'm on it

src/sbvr-api/errors.ts
```diff
@@ -6,9 +6,368 @@ export class PermissionParsingError extends TypedError {}
...
+	getResponseBody() {
+		return this.body !== undefined ?
+			this.body :
+			this.message
+	}
+}
+
+export class BadRequestError extends HttpError {
+	constructor(
+		error: string | Error = '',
```
Page-: You could change all of the subclasses to use
```typescript
error?: string | Error
```
as `HttpError` will handle the defaulting

thgreasi: Nice :+1:
Testing, rebasing, squashing & merging.

PR #174, opened 2018-05-09
Memoize the replacement of `$ACTOR.ID` as part of getting the permissions lookup object.

Page-:
Change-type: patch


Page-:
It can actually reduce memory usage, because before the cache key included every permission with the $ACTOR.ID replaced and in cases where there are multiple actor ids (both api key and jwt provided) it would duplicate the entire list of permissions for each actor whereas now it just stores the standard list + the actor id/ids as the key.  That said for the most part it's probably going to be pretty neutral on memory but avoids a lot of replaces which actually end up being surprisingly costly

PR #178, opened 2018-06-08
Add support for string bind keys

Page-:
Related-to: resin-io-modules/odata-parser#22
Related-to: resin-io-modules/abstract-sql-compiler#30
Change-type: minor


src/sbvr-api/sbvr-utils.coffee
```diff
@@ -183,6 +183,12 @@ getAndCheckBindValues = (vocab, odataBinds, bindings, values) ->
 					throw new Error('Invalid binding')
 				[dataType, value] = odataBinds[binding[1]]
 				field = { dataType }
+			else if _.isString(binding[1])
+				if !odataBinds.hasOwnProperty(binding[1])
+					console.error("Invalid binding '#{binding[1]}' for binds: ", odataBinds)
```
dfunckt: Might make sense to be consistent with the integer error above and the message be “Invalid binding string...”.

PR #179, opened 2018-06-14
Pinejs 8.x

Page-:
This includes the following 3 breaking changes

* Make db changes to support later updating pg to 7.x in a minor version
  The actual update still has issues related to using clients after they've been closed, I think due to the pooling system but I haven't been able to find why yet
* Move the api key cache settings to be standardised
* Move permissions to the sql layer rather than odata layer.
  This was originally part of #149 however due to performance issues had to be delayed, the latest 2 commits fix those performance issues by switching from the `$ACTOR.ID` hack to a standard odata parameter alias of `@__ACTOR_ID`.  This means the permissions we parse/use no longer include the actor id and so we can generate a single constrained model for a given set of permissions, and just provide a different bind variable (fixing the performance issues we had before).

You can check out this PR in use in https://github.com/resin-io/resin-api/pull/1095/files


Page-:
It wouldn't have been feasible to add the new things in typescript, it evolved from existing code and is heavily interlinked with the coffee-script so writing it in a new file would have been significantly harder.  If it was purely new isolated code I would definitely have done it in typescript though

src/sbvr-api/permissions.coffee
```diff
@@ -404,7 +505,27 @@ exports.setup = (app, sbvrUtils) ->
 	# A default api key middleware for convenience
 	exports.apiKeyMiddleware = apiKeyMiddleware = customApiKeyMiddleware()
 
-	exports.checkPermissions = checkPermissions = do ->
+	exports.checkPermissions = (req, actionList, resourceName, vocabulary) ->
+		getReqPermissions(req)
+		.then (permissionsLookup) ->
+			checkPermissions(permissionsLookup, actionList, vocabulary, resourceName)
```
dfunckt: Isn't this now `undefined`?

src/sbvr-api/permissions.coffee
```diff
@@ -425,160 +546,145 @@ exports.setup = (app, sbvrUtils) ->
...
-					'model'
-				else if methodPermissions[method]?
-					methodPermissions[method]
+	parseRewrittenPermissions = (abstractSqlModel, vocabulary, resourceName, filter, odataBinds) ->
+		{ tree, extraBinds } = _parsePermissions(filter)
+		tree = memoizedRewriteODataOptions(abstractSqlModel, vocabulary, resourceName, filter, tree)
+		return rewriteBinds({ tree, extraBinds }, odataBinds)
+
+	addODataPermissions = (permissionsLookup, permissionType, vocabulary, resourceName, odataQuery, odataBinds, abstractSqlModel) ->
+		conditionalPerms = checkPermissions(permissionsLookup, permissionType, vocabulary, resourceName)
```
dfunckt: This one as well?

src/sbvr-api/permissions.coffee
Page-: No, it's defined above where `_checkPermissions` was renamed to `checkPermissions`

PR #181, opened 2018-07-06
Add `getAffectedIds` which can be used to get a list of ids a request will affect.

Page-:
This takes over from a custom method of getting affected ids in our api and should have better performance along with more accurate results when used with non-GET requests

Change-type: minor


src/sbvr-api/sbvr-utils.coffee
```diff
@@ -723,6 +734,39 @@ exports.getAbstractSqlModel = getAbstractSqlModel = (request) ->
...
+		method: request.method
+		url: "/#{request.vocabulary}#{request.url}"
+	})
+	.then (request) ->
+		abstractSqlModel = getAbstractSqlModel(request)
+		resourceName = resolveSynonym(request)
+		resourceTable = abstractSqlModel.tables[resourceName]
+		if !resourceTable?
+			throw new Error('Unknown resource: ' + request.resourceName)
+		idField = resourceTable.idField
```
thgreasi: NIT: it seems that this could be destructured but I also like it as is :)

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -723,6 +734,39 @@ exports.getAbstractSqlModel = getAbstractSqlModel = (request) ->
...
+
+		_.set(request.odataQuery, [ 'options', '$select' ], { 'properties': [ { 'name': idField } ] })
```
thgreasi: How about dropping the single quotes from `'properties'` & `'name'` ?
Multiline formatting could also help IMO

src/sbvr-api/sbvr-utils.coffee
```diff
@@ -723,6 +734,39 @@ exports.getAbstractSqlModel = getAbstractSqlModel = (request) ->
...
+
+		permissions.addPermissions(req, request)
+
+		request.method = 'GET'
```
thgreasi: Would it make sense to also clear the `body` and the `$expand` & `$orderBy` options of the request?

Page-: The `$orderby` needs to be kept for cases like `$orderby=name&$top=10` where it affects which ids are affected.  `$expand` can be dropped but I think it's invalid on anything other than a `GET` so shouldn't be able to be here - I need to double check that though and also add a custom case for passing a `GET` request to here (which should never happen imo, but if it does it should probably either throw an error or return `[]` as a GET doesn't affect things)

thgreasi: Just-an-idea: How about dropping the `$orderby` when neither `$top` nor `$skip` is used in the current request.
I might be missing an edge case and this might be premature optimization though.

src/sbvr-api/uri-parser.coffee
```diff
@@ -91,13 +97,9 @@ exports.parseOData = (b) ->
...
 			odata = memoizedParseOdata(url)
 
-			# if we parse a canAccess action rewrite the resource to ensure we
-			# do not run the resource hooks
-			if odata.tree.property?.resource == 'canAccess'
-				odata.tree.resource = odata.tree.resource + '#' + odata.tree.property.resource
-
 			return {
 				method: b.method
+				url
```
nazrhom: NIT: would rather have `url: url` for consistency

src/sbvr-api/sbvr-utils.coffee
Page-: Added the check for passing in a GET request and also removed `$expand`s.  I'd say we should just leave $orderby in as the performance difference should be neglible

src/sbvr-api/uri-parser.coffee
Page-: It would be more consistent with the other entries in this specific object because they all use a different key, but it's consistent with what we usually do in other places with omitting the `: x` if we have `x: x` (especially in typescript where we always have the `{}`)

PR #189, opened 2018-08-07
Fix missing `promise: true` for memoisation

Page-:
Change-type: patch


Page-:
@resin-ci retest

PR #190, opened 2018-08-09
Update to typescript 3

Page-:
Change-type: patch


thgreasi:
The version bump of pinejs-client-core might have deserved an entry in the changelog :)
Isn't this allowing `/$count` in `$expand`s right away or does it need any extra work in pine itself?

thgreasi:
Oh I just realized that the sdk is still on pinejs-client v4...
Will add that to my todo list

PR #192, opened 2018-08-15
Add a type definition for `getAffectedIds`

Page-:
Change-type: patch


src/sbvr-api/sbvr-utils.d.ts
```diff
@@ -23,23 +23,26 @@ interface User {
 	permissions?: string[];
 }
 
+type HookReq = {
```
CameronDiver: Why are these types and not interfaces?

Page-: Basically I copied how `HookArgs` is specified

CameronDiver: You added those too though? :P

PR #194, opened 2018-08-18
Convert the migrator to typescript

Page-:
Change-type: minor


src/migrator/migrator.ts
```diff
@@ -0,0 +1,136 @@
...
+		resource: 'migration',
+		id: modelName,
+		passthrough: {
+			tx,
+			req: sbvrUtils.root,
+		},
+		body: {
+			model_name: modelName,
+			executed_migrations: executedMigrations,
+		},
```
dfunckt: I think you can pass `options: { returnResult: false }` here.

src/migrator/migrator.ts
```diff
@@ -0,0 +1,136 @@
...
+	.value()
+
+export const executeMigrations = (tx: Tx, migrations: Array<MigrationTuple> = []): Promise<string[]> =>
+	Promise.mapSeries(migrations, executeMigration.bind(null, tx))
+	.catch((err) => {
+		sbvrUtils.api.migrations.logger.error('Error while executing migrations, rolled back')
+		throw new MigrationError(err)
+	}).return(_.map(migrations, ([ migrationKey ]) => migrationKey)) // return migration keys
+
+export const executeMigration = (tx: Tx, [ key, migration ]: MigrationTuple) => {
```
dfunckt: Can you please specify the type of the return value explicitly?

It also seems this function returns a promise some times, but can also potentially throw synchronously (which `executeMigrations` above will catch, but given the method is also exported it makes sense to always return a promise, or not export it at all).

PR #196, opened 2018-08-23
Improve/fix db-layer types

Page-:
Change-type: patch


src/database-layer/db.ts
```diff
@@ -61,14 +61,20 @@ const alwaysExport = {
 	UniqueConstraintError,
 	ForeignKeyConstraintError,
 }
+
+interface TransactionFn {
+	<T>(fn: (tx: Tx) => Promise<T> | T): Promise<T>
+	(): Promise<Tx>
```
dfunckt: What sort of magic is this -- is this function overloading, selecting either of the two depending on whether an argument is provided? TS is crazy.

PR #199, opened 2018-08-29
More type improvements

Page-:
Change-type: patch


src/sbvr-api/sbvr-utils.d.ts
```diff
@@ -30,17 +30,27 @@ export type AnyObject = {
...
 }
 
-export interface User {
+interface Actor {
+	permissions?: string[];
+}
+
+export interface User extends Actor {
 	id: number;
 	permissions?: string[];
```
thgreasi: I guess that this can now be removed, since it is inherited from Actor now.

src/sbvr-api/sbvr-utils.d.ts
```diff
@@ -30,17 +30,27 @@ export type AnyObject = {
...
 }
 
+export interface ApiKey extends Actor {
+	key: string;
+	permissions?: string[];
```
thgreasi: Ditto, do we need to repeat this in here as well?

src/sbvr-api/sbvr-utils.d.ts
Page-: Yeah it can be, I meant to but I forgot - removed it now

Page-:
@resin-ci retest

PR #207, opened 2018-09-18
Add `initSqlPath`/`initSql` which are run only on first model execution

Page-:
Depends on #206 

Change-type: minor

### DIFF
--- a/src/migrator/migrator.ts
+++ b/src/migrator/migrator.ts
@@ -5,13 +5,11 @@ const modelText: string = require('./migrations.sbvr')
 import { Tx } from '../database-layer/db'
 import * as sbvrUtils from '../sbvr-api/sbvr-utils'
 import { migrator as migratorEnv } from '../config-loader/env'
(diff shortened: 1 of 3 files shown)


src/migrator/migrator.ts
```diff
@@ -21,6 +22,25 @@ export type Migration = string | MigrationFn
...
+	}
+
+	const modelName = model.apiRoot
+
+	return Promise.using(lockMigrations(tx, modelName), () =>
+		checkModelAlreadyExists(tx, modelName)
+		.then((exists) => {
+			if (!exists) {
+				sbvrUtils.api.migrations.logger.info('First time executing, running init script')
+				return tx.executeSql(initScript).return()
```
dfunckt: `initScript` makes you expect you can specify the path to a shell script -- it would probably be better to name the attribute `initSql` instead.

Page-: done

Page-:
@resin-ci retest

PR #208, opened 2018-09-25
Convert config-loader to typescript

Page-:
Change-type: patch


src/config-loader/config-loader.ts
```diff
@@ -0,0 +1,293 @@
...
+import * as fs from 'fs'
+// We force the type to match the specific overloaded version we want, as promisify can only choose one :(
+const readFileAsync = Promise.promisify(fs.readFile) as any as (filename: string, encoding: string) => Promise<string>
+const readdirAsync = Promise.promisify(fs.readdir)
+
+import * as sbvrUtils from '../sbvr-api/sbvr-utils'
+const permissions = require('../sbvr-api/permissions')
+
+export interface SetupFunction {
+	(app: _express.Application, sbvrUtils: any, db: any, done: (err: any) => void): Promise<void> | void
```
dfunckt: Shouldn't `done` be optional? Isn't this actually better as two overloaded functions -- one that returns void when given callback, another that doesn't accept a callback and returns promise

src/config-loader/config-loader.ts
```diff
@@ -0,0 +1,293 @@
...
+	permissions: string[]
+}
+export interface Config {
+	models: Model[]
+	users?: User[]
+}
+
+// Setup function
+export const setup = (app: _express.Application) => {
+	const loadConfig = (data: Config): Promise<void> =>
```
dfunckt: This must be the ugliest function I've ever seen, there's no chance one can follow control flow without getting a headache. It would be great to split it up in several smaller ones, though it's fine if you don't want to do this in this PR.

Page-: Yeah I'd rather keep it to primarily what is necessary for the typescript conversion and avoid unnecessary changes

PR #216, opened 2018-10-09
9.x

Page-:
This contains a couple the following breaking changes:
* Update abstract-sql-compiler to 6.x and sbvr-types to 2.x
  - This causes breakings changes to the reexported sbvrTypes property, but updates to the latest bcrypt which is required for nodejs 10
* Reject internal requests with errors instead of numbers/strings
  - This causes breaking changes if you're matching directly against a string/number in a rejected internal request, but helps standardise errors a lot and helps with propogating error codes


Page-:
The links will be switched before I merge, but having them as is means I can test the whole chain in the API using github deps

PR #223, opened 2018-11-05
Fix webpack builds

Page-:
Change-type: patch


Gruntfile.coffee
```diff
@@ -8,14 +9,15 @@ serverConfigs =
...
-				unused: false # We need this off for OMeta
-			mangle: false
+		new UglifyJsPlugin(
+			uglifyOptions:
+				output:
+					beautify: true
+					ascii_only: true
+				compress:
+					sequences: false
+					unused: false # We need this off for OMeta
```
zvin: Do we still need this?

Page-: Yes, we still have ometa -> js compiled code that would be affected

PR #226, opened 2018-11-15
Optimise `checkForExpansion`

Page-:
Change-type: patch


src/sbvr-api/sbvr-utils.coffee
```diff
@@ -474,11 +474,15 @@ exports.getID = (vocab, request) ->
...
-	try
-		field = JSON.parse(instance[fieldName])
-	catch
-		# If we can't JSON.parse the field then we use it directly.
-		field = instance[fieldName]
+	field = instance[fieldName]
+	if !field?
+		return
+
+	if _.isString(field)
```
Page-: Doing the `JSON.parse` unconditionally was very expensive as it would throw errors (which include stack traces) for every expandable field that we haven't expanded (multiplied by every row).  So we restrict to only trying to parse strings as they're the only thing that can be a JSON string

PR #227, opened 2018-11-16
Optimise `memoizedGetConstrainedModel`

Page-:
Change-type: patch

### DIFF
--- a/src/sbvr-api/permissions.coffee
+++ b/src/sbvr-api/permissions.coffee
@@ -246,26 +246,30 @@ deepFreezeExceptDefinition = (obj) ->
 			deepFreezeExceptDefinition(obj)
 	return
 
-memoizedGetConstrainedModel = memoizeWeak(
-	(abstractSqlModel, permissionsLookup, vocabulary) ->
-		abstractSqlModel = _.cloneDeep(abstractSqlModel)
-		addRelationshipBypasses(abstractSqlModel.relationships)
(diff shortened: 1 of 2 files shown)


src/sbvr-api/permissions.coffee
```diff
@@ -246,26 +246,30 @@ deepFreezeExceptDefinition = (obj) ->
...
-		_.each abstractSqlModel.tables, (table) ->
-			abstractSqlModel.tables["#{table.resourceName}$bypass"] = _.clone(table)
-			onceGetter table, 'definition', ->
-				# For $filter on eg a DELETE you need read permissions on the sub-resources,
-				# you only need delete permissions on the resource being deleted
-				generateConstrainedAbstractSql(permissionsLookup, methodPermissions.GET, vocabulary, sqlNameToODataName(table.name))
-		deepFreezeExceptDefinition(abstractSqlModel)
-		return abstractSqlModel
-	normalizer: (abstractSqlModel, args) ->
-		return JSON.stringify(args)
```
Page-: This normalizer currently takes 1.18% of the total api test time (including idle time, so its real effect is larger), by switching to a double weak memoizer it means we can pretty much completely remove the cost and it also means that we'll be able to free memory when `permissionLookup` objects are evicted from the `getPermissionsLookup` cache - as well as using significantly less memory for the cache key.

The reason we can do this is because we memoize `getPermissionsLookup` so we know we know the object ref will be consistent and can be relied upon

PR #233, opened 2018-11-30
exposed DB metrics via an event emitter on pinejs module (for API use) 

dt-rush:


dt-rush:
@Page- was rebased on master

Page-:
There's no reason to create a new PR btw, you can just push over the existing branch and the existing PR will automatically update - it makes it easier to track review passes and comments if it all happens in one PR

src/database-layer/db.ts
```diff
@@ -184,8 +187,16 @@ export abstract class Tx {
 	public executeSql(sql: Sql, bindings: Bindings = [], ...args: any[]): Promise<Result> {
 		this.incrementPending()
 
+		let t0 = Date.now();
 		return this._executeSql(sql, bindings, ...args)
-			.finally(() => this.decrementPending())
+			.finally(() => {
+				this.decrementPending()
+				const queryTime = (Date.now() - t0) / 1000;
```
Page-: I think it makes more sense to return the result in milliseconds and avoid this `/ 1000` which will trigger a switch from integer to floating point mode

src/database-layer/db.ts
```diff
@@ -184,8 +187,16 @@ export abstract class Tx {
...
+				metricsEmitter.emit('db_query_time', { 
+					queryTime: queryTime,
+					queryType: sql.substr(0, sql.indexOf(' '))
```
Page-: It's possible to run multiple statements in one `executeSql` call by separating the different queries with `;`, it's also possible that the first line is a comment like `-- Some comment`, and also it could be that the user just has leading space, eg ` SELECT`, so this query type detection is going to be shaky due to that.  I'm not sure if there's a better way to do it here but it definitely comes with a bunch of caveats as it stands.

- this doesn't block the PR, it would just be nice to improve on if possible

dt-rush: it's a pretty rough approximation, agreed. A poor approximation though it is, I've left it and added two comments tagged metrics-TODO: which point to two avenues of future improvement I've identified.

src/database-layer/db.ts
dt-rush: As in the API, agree fully and removed the / 1000.

src/database-layer/db.ts
```diff
@@ -2,12 +2,14 @@
...
 import * as _pg from 'pg'
 import * as _pgConnectionString from 'pg-connection-string'
-
+import * as EventEmitter from 'eventemitter3';
 import * as _ from 'lodash'
 import * as Promise from 'bluebird'
 import TypedError = require('typed-error')
 import * as env from '../config-loader/env'
 
+export const metricsEmitter = new EventEmitter();
```
Page-: I think renaming this to just `metrics` make things flow a little nicer, the `Emitter part feels redundant. eg
```ts
db.metrics.on('db_error', ...)
db.metrics.on('db_query_time', ...)
metrics.emit(...)

dt-rush: Nice!

PR #241, opened 2019-01-16
Convert permissions to typescript

Page-:
depends on #240 

Change-type: minor


src/sbvr-api/permissions.ts
```diff
@@ -0,0 +1,1058 @@
...
+	const permsFilter = {
+		$or: {
+			is_of__user: {
+				$any: {
+					$alias: 'uhp',
+					$expr: {
+						uhp: { user: userId },
+						$or: [{
+							uhp: { expiry_date: null }
+						},	{
```
dfunckt: A tab slipped in there.

src/sbvr-api/permissions.ts
```diff
@@ -0,0 +1,1058 @@
...
+											is_of__user: {
+												$any: {
+													$alias: 'uhr',
+													$expr: {
+														uhr: { user: userId },
+														$or: [{
+															uhr: { expiry_date: null }
+														}, {
+															uhr: {
+																expiry_date: { $gt: { $now: null } }
```
dfunckt: We really need a helper for lambdas -- it was passable before with CS dropping much of the decoration, but in TS they're very cumbersome to write.

src/sbvr-api/permissions.ts
```diff
@@ -0,0 +1,1058 @@
...
+	.then((permissions) => {
+		req.apiKey = {
+			key: apiKey,
+			permissions: permissions
+		}
+	})
+})
+
+export const customAuthorizationMiddleware = (expectedScheme = 'Bearer') => {
+	if (expectedScheme == null) {
```
dfunckt: You have the argument default above -- do you still need to check for null?

src/sbvr-api/permissions.ts
```diff
@@ -0,0 +1,1058 @@
...
+		}).then(() => {
+			next && next()
+		})
+}
+
+// A default bearer middleware for convenience
+export const authorizationMiddleware = customAuthorizationMiddleware()
+
+export const customApiKeyMiddleware = (paramName = 'apikey') => {
+	if (paramName == null) {
```
dfunckt: Same here.

src/sbvr-api/permissions.ts
```diff
@@ -0,0 +1,1058 @@
...
+		paramName = 'apikey';
+	  }
+	return (req: HookReq | express.Request, _res?: express.Response, next?: express.NextFunction): Promise<void> => {
+		const apiKey = req.params[paramName] != null
```
dfunckt: Why not just:

```js
const apiKey = req.params[paramName] || req.body[paramName] || req.query[paramName];
```

src/sbvr-api/permissions.ts
```diff
@@ -0,0 +1,1058 @@
...
+			options: {
+				$select: 'id',
+				$filter: {
+					username: 'guest'
+				}
+			}
+		})
+		.then((result: AnyObject[]) => {
+			if (result.length === 0) {
+				throw new Error('No guest permissions')
```
dfunckt: I know that would be semantically different, but I think it would be better to just return an empty array.

src/sbvr-api/permissions.ts
```diff
@@ -0,0 +1,1058 @@
...
+			} catch (e) {
+				console.warn('Failed to parse conditional permissions: ', permissionCheck)
+				throw new PermissionParsingError(e)
+			}
+		})
+
+		// if (permissionFilters === false) {
+		// 	throw new PermissionError()
+		// }
+		// if (permissionFilters !== true) {
```
dfunckt: Why did you comment this code?

src/sbvr-api/permissions.ts
```diff
@@ -0,0 +1,1058 @@
...
+	Object.defineProperty(obj, propName, {
+		enumerable: true,
+		configurable: true,
+		get() {
+			if (thrownErr != null) {
+				throw thrownErr
+			}
+			try {
+				const result = nullableFn!()
+				delete this[propName]
```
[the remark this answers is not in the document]
Page-: Basically the property is configurable but not writable, so `this[propName] = result` will do nothing. 
 However `delete this[propName]` is different and will remove it and all the properties associated with it which then means that `this[propName] = result` will work as expected

src/sbvr-api/permissions.ts
Page-: Yeah.. I thought prettier was set up on this repo but apparently not.. I'll add it in another PR and then update this one

src/sbvr-api/permissions.ts
Page-: Yeah, the argument default in ts/js/cs2 does
```ts
expectedScheme = expectedScheme === undefined ? 'Bearer' : expectedScheme
```
but this catches `null` as well to match what defaults in coffeescript 1 do.  I plan to remove this in future but it would be a potentially breaking change and need a major version

src/sbvr-api/permissions.ts
Page-: Because it's not actually necessary - `permissionFilters` can never be true/false here as you either need to pass in true/false or have the callback to `nestedCheck` return true/false in order for `nestedCheck` to return true/false.  And we know that the `conditionalPerms` we pass in cannot be true/false because we've already checked for it.

src/sbvr-api/permissions.ts
Page-: The current form matches the previous behaviour which is what I was trying to stick to for the most part - ie keep everything as-is but with added types in order to minimise the chance of introducing bugs with such a large diff

src/sbvr-api/permissions.ts
```diff
@@ -0,0 +1,1058 @@
...
+// object: Must have only one key of either `AND` or `OR`, with an array value that will be treated according to the key.
+const isAnd = <T>(x: any): x is NestedCheckAnd<T> => _.isObject(x) && 'and' in x
+const isOr = <T>(x: any): x is NestedCheckOr<T> => _.isObject(x) && 'or' in x
+export function nestedCheck<I, O>(check: string, stringCallback: (s: string) => O): O
+export function nestedCheck<I, O>(check: boolean, stringCallback: (s: string) => O): boolean
+export function nestedCheck<I, O>(check: NestedCheck<I>, stringCallback: (s: string) => O): Exclude<I, string> | O | MappedNestedCheck<typeof check, I, O>
+export function nestedCheck<I, O>(check: NestedCheck<I>, stringCallback: (s: string) => O): boolean | Exclude<I, string> | O | MappedNestedCheck<typeof check, I, O> {
+	if (_.isString(check)) {
+		return stringCallback(check)
+	} else if (_.isBoolean(check)) {
```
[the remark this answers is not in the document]
Page-: done as a separate commit

src/sbvr-api/permissions.ts
dfunckt: It is just surprising behaviour IMO, adding an extra special case that needs to be handled when it could be very well represented by the return value.

The function is called `getPermissions` IIRC so I'd expect it to return a set of permissions, including returning an empty set which unequivocally means the same thing as "no permissions". It should be up to the caller to decide whether that signifies an error condition.

Plus, I don't see why I, the caller, should be prepared to handle the "no permissions" case differently. I would only think to potentially catch errors like "the given user is not valid", or something similar.

Page-: @dfunckt that makes sense and actually this case is the one that there is no guest user and the error message is wrong/confusing.  I've updated the error message to say user rather than permissions which should also make it clearer when anyone hits the error

dfunckt: Haha, I completely missed that :) I sort of read `perms.length == 0` all along!

PR #242, opened 2019-01-16
Convert pinejs-session-store to typescript

Page-:
Change-type: minor


src/pinejs-session-store/pinejs-session-store.ts
```diff
@@ -0,0 +1,133 @@
...
+	Fact type:  session has session id
+		Necessity: Each session has exactly 1 session id
+		Necessity: Each session id is of exactly 1 session
+	Fact type:  session has expiry time
+		Necessity: Each session has at most 1 expiry time
+`
+
+export = class PinejsSessionStore extends Store {
+	get = ((sid, callback) => {
+		api.session.get({
```
dfunckt: Don’t you need to return the promise as well?

Page-: Nope, express-session only supports a callback interface (the typings would complain if I did return the promise)

Page-:
It is not

PR #246, opened 2019-01-22
Convert odata-metadata-generator to typescript

Page-:
Change-type: patch


typings/sbvr-types.d.ts
```diff
@@ -0,0 +1,18 @@
+declare module '@resin/sbvr-types' {
+	import * as Promise from 'bluebird';
```
zvin: NIT: call this Bluebird

src/sbvr-compiler/odata-metadata-generator.ts
```diff
@@ -0,0 +1,173 @@
...
+import * as sbvrTypes from '@resin/sbvr-types';
+import { SqlModel, AbstractSqlTable } from '@resin/abstract-sql-compiler';
+
+// tslint:disable-next-line:no-var-requires
+const { version }: { version: string } = require('../../package.json');
+
+const getResourceName = (resourceName: string): string => {
+	const idParts = resourceName.split('-');
+
+	return idParts.map(idPart => idPart.split(/[ -]/).join('_')).join('__');
```
zvin: Can `idPart` still contain a '-' after the first split above?

Page-: It can't, I've updated the code

typings/sbvr-types.d.ts
Page-: We use `Promise` everywhere else so I'd rather keep the consistency, we don't use native promises because we still support node versions where native promises are slower

zvin: I wasn't suggesting to switch to native Promises, just calling these things Bluebirds.
Let's keep it as `Promise` if it's like that everywhere.

PR #250, opened 2019-01-24
Fix rolling back req hooks

Page-:
Change-type: patch

### DIFF
--- a/src/sbvr-api/hooks.ts
+++ b/src/sbvr-api/hooks.ts
@@ -49,35 +49,23 @@ export class SideEffectHook extends Hook {
 }
 
 // The execution order of rollback actions is unspecified
-const undoHooks = Promise.method(
-	(request: { hooks?: InstantiatedHooks<object> }) => {
-		if (request.hooks == null) {
-			return [];
-		}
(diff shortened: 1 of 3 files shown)


src/sbvr-api/hooks.ts
```diff
@@ -49,35 +49,23 @@ export class SideEffectHook extends Hook {
...
-			},
-		);
-	},
-);
-
 export const rollbackRequestHooks = (
-	request:
-		| { hooks?: InstantiatedHooks<object> }
-		| Array<{ hooks?: InstantiatedHooks<object> }>,
+	hooks: InstantiatedHooks<object> | undefined,
```
CameronDiver:
```suggestion
	hooks?: InstantiatedHooks<object>,
```

Page-: I've done it this way because the parameter is not optional, it should always be provided or there's no point in calling the function, it's just it means that calling this as `rollbackRequestHooks(request.hooks)` becomes easier because `request.hooks` can be undefined

CameronDiver: Makes sense.

PR #256, opened 2019-01-28
Extract odata response

Page-:
Depends on #255

### DIFF
--- a/src/sbvr-api/odata-response.ts
+++ b/src/sbvr-api/odata-response.ts
@@ -1,0 +1,207 @@
+declare module '@resin/abstract-sql-compiler' {
+	interface AbstractSqlTable {
+		fetchProcessingFields?: {
+			[field: string]: (field: any) => Promise<any>;
+		};
+		localFields?: {
+			[odataName: string]: true;
+		};
+	}
+}
+
+import * as _ from 'lodash';
+import * as Promise from 'bluebird';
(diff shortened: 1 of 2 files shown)


src/sbvr-api/odata-response.ts
```diff
@@ -0,0 +1,207 @@
...
+			}
+		}
+	}
+	return table.localFields;
+};
+const getFetchProcessingFields = (table: AbstractSqlTable) => {
+	if (table.fetchProcessingFields == null) {
+		table.fetchProcessingFields = _(table.fields)
+			.filter(
+				({ dataType }) =>
```
dfunckt: Prettier becomes weirder and weirder each release -- what sort of formatting is that :/

src/sbvr-api/odata-response.ts
```diff
@@ -0,0 +1,207 @@
...
+	const processedFields = _.filter(
+		_.keys(instances[0]),
+		fieldName =>
+			!_.startsWith(fieldName, '__') &&
+			fetchProcessingFields.hasOwnProperty(fieldName),
+	);
+	if (processedFields.length > 0) {
+		instancesPromise = instancesPromise
+			.then(() => {
+				return Promise.map(instances, instance => {
```
dfunckt: You can remove all these `return`s if you like.

Issue #263, opened 2019-02-28
Add automatic modifiedAt field to all resources


petrosagg:

Page-:
https://github.com/balena-io/lf-to-abstract-sql/pull/24

Page-:
https://github.com/balena-io-modules/abstract-sql-compiler/pull/53

PR #264, opened 2019-03-04
11.x

Page-:
* Drop dual promise/callback based interfaces in favour of promise based
* Stop applying default for `null` in `runURI` and `customAuthorizationMiddleware`
    This matches standard javascript behavior for defaults
* Add an automatic "migrated at" field, with a required migration

Implements #263

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "grunt-ts": "^6.0.0-beta.22",
     "grunt-webpack": "^3.1.3",
     "husky": "^4.0.10",
-    "lint-staged": "^9.5.0",
+    "lint-staged": "^10.0.0",
     "load-grunt-tasks": "^5.1.0",
     "prettier": "^1.19.1",
     "raw-loader": "^4.0.0",
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@types/method-override": "0.0.31",
     "@types/multer": "^1.3.10",
     "@types/mysql": "0.0.34",
-    "@types/node": "^6.14.9",
+    "@types/node": "^10.17.13",
     "@types/passport": "^0.4.7",
     "@types/passport-local": "1.0.33",
     "@types/passport-strategy": "^0.2.35",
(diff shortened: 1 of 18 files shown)


gelbal:
Hey @Page-, what do you think about exposing this new field on `external` schema as well?

Not that I know of a use case right now. But it might be useful eventually. And I prefer to keep the external schema on par with the main schema.

Page-:
@gelbal it can be exposed there, but that's a separate concern since those are just views - adding them here just means they'll be available to include in those views if we want

gelbal:
@Page- fair enough. I'll raise a new issue / PR when we find a use case for this column.

package.json
```diff
@@ -36,7 +36,7 @@
     "@types/method-override": "0.0.31",
     "@types/multer": "1.3.7",
     "@types/mysql": "0.0.34",
-    "@types/node": "^6.14.4",
+    "@types/node": "^10.14.4",
```
thgreasi: Shouldn't this be `^8.x.x`, so that we don't get typings for functions that will eventually fail on the node 8 test worker?

Page-:
Reminder: Update package.json engines field before merging

package.json
Page-: The idea is to target 10+ as it's what we use everywhere and is the active LTS and stops us from being limited to a major bump to make use of node 10 specific features if they become useful whilst at the same time keeping the test worker around so as not to accidentally break node 8 support without a good reason, eg you can break node 6 support by adding a trailing `,` in `function (a, b, c,) {}` and I'd like to avoid breaking for something like that

That said the "test" worker mainly just ensures the codebase can be built so it's of limited use for enforcing if something works or not and instead pinejs usually gets heavily tested via our api tests

roman-mazur:
@Page- do we also track `createdById` and `modifiedById`?

Page-:
no @roman-mazur, if we did it would have to be pointed at the actor id instead (because it's not just "users" that can do modifications) and then there'd have to be added the concept of a "root"/"system" actor since internal requests don't necessarily happen from the context of a specific actor atm

roman-mazur:
Yep, adding something like "admin" or "system" is what I've done before.

So let's imagine we decide to add this type of functionality too. In such case, we would need to perform another round of manual migrations. Just a motivation to change how migrations are done (maybe in 11.1) :)

Page-:
@roman-mazur I've changed migrations in https://github.com/balena-io/pinejs/pull/332 so we can self-migrate

PR #265, opened 2019-03-06
Update pg to ^7.8.1

Page-:
Change-type: minor

### DIFF
--- a/src/config-loader/env.ts
+++ b/src/config-loader/env.ts
@@ -23,6 +23,7 @@ export const cache = {
 export const db = {
 	poolSize: 50,
 	idleTimeoutMillis: 30000,
+	connectionTimeoutMillis: 30000,
 };
 
 export const migrator = {
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@types/passport": "^0.3.5",
     "@types/passport-local": "1.0.33",
     "@types/passport-strategy": "^0.2.35",
-    "@types/pg": "^6.1.47",
+    "@types/pg": "^7.4.13",
     "@types/websql": "0.0.27",
     "bluebird": "^3.5.3",
     "coffeescript": "^2.3.2",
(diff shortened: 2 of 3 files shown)


src/config-loader/env.ts
```diff
@@ -23,6 +23,7 @@ export const cache = {
 export const db = {
 	poolSize: 50,
 	idleTimeoutMillis: 30000,
+	connectionTimeoutMillis: 30000,
```
dfunckt: What is this timeout for exactly? Is there any way for it to unexpectedly close connections, eg. in the middle of a transaction?

Page-:
@resin-ci retest

src/config-loader/env.ts
Page-: No, this is for initially creating a connection, eg `pool.connect()` - right now it will wait forever so if we use up all of the pool then things will start to just appear to hang, this will just cause connection attempts that have been waiting for too long to fail (which they almost certainly would anyway due to request timeouts/etc)

Page-:
@resin-ci retest

src/config-loader/env.ts
dfunckt: I see, that makes sense, thanks. Is a much shorter timeout, maybe around 5 seconds, reasonable? I mean if it takes 5 seconds to get a free connection from the pool you're already in a pretty bad state, having the timeout at 30 feels like things could really get clogged up.

dfunckt: And, incidentally, 30 seconds is already much more than most API clients have for their total request timeout (I mean devices and backend components).

Page-:
@resin-ci retest

nazrhom:
@resin-ci retest

src/config-loader/env.ts
Page-: So specifically it's to stop the api itself getting into a broken state due to all the waiting connection attempts when it's presumably already heavily overloaded.  There are other mechanisms around cancelling timed out requests and similar that handle stopping in-progress work when a request ends but this is a special case that we have less control over.

All that aside it's set up so that the values can be adjusted by any downstream users to fit their needs (which we already do for adjusting the pool size)

dfunckt: Agreed, thanks for the explanation!

Page-: I also want to start off conservative because right now we have no connection timeout at all, so I'd rather not set it too low by default and then find out the hard way

PR #266, opened 2019-03-07
Fix for latest bluebird typings

Page-:
Change-type: patch


src/sbvr-api/permissions.ts
```diff
@@ -634,7 +634,10 @@ const getPermissions = (
 				},
 			},
 		})
-		.map(permission => permission.name)
+		.then(
+			(permissions: AnyObject[]): string[] =>
```
Page-: Specifically because `permissions` here is typed as potentially a number or object it meant that `Promise.map` failed due to expecting an iterable however we know it can only be an array based on the call so this casts it and we do a synchronous map instead - which should also be better performance.

PR #267, opened 2019-03-07
Use `?` to check for complex urls as it's more accurate than `$`

Page-:
Change-type: patch

### DIFF
--- a/src/sbvr-api/uri-parser.ts
+++ b/src/sbvr-api/uri-parser.ts
@@ -89,7 +89,7 @@ export const memoizedParseOdata = (() => {
 		max: env.cache.parseOData.max,
 	});
 	return (url: string) => {
-		if (_.includes(url, '$')) {
+		if (_.includes(url, '?')) {
 			// If we're doing a complex url then skip caching due to # of permutations
 			return parseOdata(url);
 		} else {


src/sbvr-api/uri-parser.ts
```diff
@@ -89,7 +89,7 @@ export const memoizedParseOdata = (() => {
 		max: env.cache.parseOData.max,
 	});
 	return (url: string) => {
-		if (_.includes(url, '$')) {
+		if (_.includes(url, '?')) {
```
dfunckt: Don't we also want to skip caching plain requests such as `/v5/device(12345)`?

Page-: Potentially but I think this covers the most important cases (based on checking cache entries in the api).  I also wonder whether with the much faster odata-parser if it's worth having this cache at all but that's a separate discussion

Page-:
@resin-ci retest

Page-:
@resin-ci retest

Page-:
@resin-ci retest

Page-:
@resin-ci retest

nazrhom:
@resin-ci retest

Page-:
@resin-ci retest

PR #268, opened 2019-03-08
Definition based DML permissions

Page-:


package.json
```diff
@@ -31,7 +31,7 @@
     "@types/deep-freeze": "^0.1.1",
     "@types/express": "4.16.0",
     "@types/express-session": "^1.15.12",
-    "@types/lodash": "^4.14.122",
+    "@types/lodash": "4.14.122",
```
dfunckt: Is that deliberate? Why did you need to pin the version?

src/sbvr-api/permissions.ts
```diff
@@ -505,53 +509,132 @@ const createBypassDefinition = (definition: Definition) =>
 		}
 	});
 
+const getAlias = (name: string) => {
+	// TODO-MAJOR: Change $bypass to $permissionbypass or similar
+	if (name.endsWith('$bypass')) {
+		return 'bypass';
+	} else {
```
dfunckt: You don't need the `else` block.

src/sbvr-api/permissions.ts
```diff
@@ -1092,107 +1175,13 @@ const rewriteODataOptions = (
...
-	}
-};
-
 export const addPermissions = Promise.method(
 	(
 		req: PermissionReq,
 		request: ODataRequest & { permissionType?: PermissionCheck },
 	): Promise<void> => {
 		const { vocabulary, resourceName, odataQuery, odataBinds } = request;
+		// const { vocabulary, resourceName, odataBinds } = request;
```
dfunckt: Stray comment.

package.json
Page-: It's necessary because it changed `isObject` to be a true type guard so that
```ts
export const resolveOdataBind = (odataBinds: ODataBinds, value: any) => {
	if (_.isObject(value) && value.bind != null) {
		[, value] = odataBinds[value.bind];
	}
	return value;
};
```
now fails because `Property 'bind' does not exist on type 'object'.`

PR #270, opened 2019-03-25
Add support for read-only transactions for postgres/mysql backends

Page-:
Change-type: minor

### DIFF
--- a/src/database-layer/db.ts
+++ b/src/database-layer/db.ts
@@ -94,6 +94,7 @@ export type Database = {
 		bindings?: Bindings,
 	) => Promise<Result>;
 	transaction: TransactionFn;
+	readTransaction?: TransactionFn;
 };
 
 export const engines: {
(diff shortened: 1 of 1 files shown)


src/database-layer/db.ts
```diff
@@ -555,21 +562,28 @@ if (maybeMysql != null) {
...
+					tx.executeSql('START TRANSACTION;');
+					return tx;
+				}),
+			),
+			readTransaction: createTransaction(stackTraceErr =>
+				connect().then(client => {
+					const close = () => client.release();
+					const tx = new MySqlTx(client, close, stackTraceErr);
+					tx.executeSql('SET TRANSACTION READ ONLY;');
+					tx.executeSql('START TRANSACTION;');
```
dfunckt: These statements are in the opposite order of that in Postgres, is this deliberate?

Page-: Yeah, in postgres the change applies to the current transaction (and so is essentially a no-op outside of a transaction), in mysql it applies to the next transaction and will throw an error if run inside of one - just fun little differences :P

dfunckt: Cool, that's what I thought as well but didn't bother to lookup docs and wanted to confirm this isn't by accident 👍

PR #277, opened 2019-04-22
Only validate the model if the query affected at least 1 row

Page-:
Change-type: patch

### DIFF
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -337,7 +337,7 @@ export const validateModel = (
 	tx: _db.Tx,
 	modelName: string,
 	request?: uriParser.ODataRequest,
-) => {
+): Promise<void> => {
 	return Promise.map(models[modelName].sql.rules, rule => {
 		if (!isRuleAffected(rule, request)) {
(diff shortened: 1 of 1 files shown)


src/sbvr-api/sbvr-utils.ts
```diff
@@ -1569,7 +1570,7 @@ const runPut = (
 	_res: _express.Response,
 	request: uriParser.ODataRequest,
 	tx: _db.Tx,
-) => {
+): Promise<undefined> => {
```
dfunckt: Why not `void`?

Page-: Because the callers expect `undefined | number | db.Result` - this is just to explicitly type the result with the value it's currently implicitly typed with

dfunckt: Cool, I thought it would be due to what the callers expect, but I expected the result was communicated simply via the promise being resolved or rejected, in which case `void` was more appropriate.

Page-: Yeah, in this case success/failure is what really matters but there's other `runX` methods that return real values so the `undefined` is to distinguish that

PR #279, opened 2019-05-01
✨: Add an `ERROR` handler in resource hooks

afitzek:
Change-type: minor

### DIFF
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -144,6 +144,9 @@ export interface Hooks {
 			data?: any;
 		},
 	) => HookResponse;
+	'POSTRUN-ERROR'?: (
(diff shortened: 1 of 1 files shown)


src/sbvr-api/sbvr-utils.ts
```diff
@@ -1302,6 +1305,9 @@ const runRequest = (
 			}
 			throw err;
 		})
+		.tapCatch(err => {
+			return runHooks('ERROR', request.hooks, { req, request, tx, error: err });
```
Page-: Can you use `POSTRUN-ERROR` instead in case we need error handlers at more places in future

Page-: I think it could be worth putting this hook point after the `POSTRUN` hooks in case they throw an error it will allow handling, although there are potential arguments for both

PR #281, opened 2019-05-01
Expose sbvr-compiler as a bin script

Page-:
This means that it will now match the module dependencies of the installed pinejs

Change-type: minor


Page-:
This is ported from https://github.com/balena-io-modules/sbvr-compiler/ so that dependency management becomes a lot more straightforward/invisible

PR #282, opened 2019-05-03
Expose odata-compiler as a bin script

Page-:
This means that it will now match the module dependencies of the installed pinejs

Change-type: minor


src/bin/odata-compiler.ts
```diff
@@ -0,0 +1,125 @@
...
+import * as _sbvrUtils from '../sbvr-api/sbvr-utils';
+import * as _uriParser from '../sbvr-api/uri-parser';
+
+import * as program from 'commander';
+import * as fs from 'fs';
+import '../server-glue/sbvr-loader';
+
+// tslint:disable:no-var-requires
+const { version } = JSON.parse(
+	fs.readFileSync(require.resolve('../../package.json'), 'utf8'),
```
thgreasi: It would be nice if we also bumped the `engines` field in package.json to `"node": ">=8.0.0",`  (even thought that's not the only node 8  fn that we are using).

src/bin/odata-compiler.ts
```diff
@@ -0,0 +1,125 @@
...
+);
+
+const getSE = (inputFile: string) => fs.readFileSync(inputFile, 'utf8');
+
+const generateAbstractSqlModel = (sbvrFile: string) => {
+	const {
+		generateLfModel,
+		generateAbstractSqlModel,
+	} = require('../sbvr-api/sbvr-utils') as typeof _sbvrUtils;
```
thgreasi: i guess that we use `require()` and "type cast" so that we have sync lazy loading?
(Correct me if I'm wrong but I couldn't guess any other reason to not use `import from` / `await import()`).

src/bin/odata-compiler.ts
```diff
@@ -0,0 +1,125 @@
...
+	.description('parse the input OData URL into OData AST')
+	.action(parseOData);
+
+program
+	.command('translate <sbvr-file> <input-url> [output-file]')
+	.description('translate the input OData URL into abstract SQL')
+	.action(translateOData);
+
+program
+	.command('compile <sbvr-file> <input-url> [output-file]')
```
thgreasi: When I do
```
odata-compiler compile .resin.sbvr /device?$orderby=device_name%20asc&$top=1&$select=id
```
I also get
```
Command '=id' not found, did you mean:

  command 'gid' from deb id-utils
  command 'kid' from deb python-kid
  command 'fid' from deb id-utils
  command 'lid' from deb id-utils
  command 'jid' from deb jid
  command 'aid' from deb id-utils
  command 'id' from deb coreutils
  command 'eid' from deb id-utils
```
but when I add an extra select prop, that doesn't appear.
I also have to do an extra Ctrl+C to go back to my shell.
Am I doing something wrong?

src/bin/odata-compiler.ts
```diff
@@ -0,0 +1,125 @@
...
+	const translatedRequest = generateAbstractSqlQuery(sbvrFile, odata);
+	const {
+		compileRequest,
+	} = require('../sbvr-api/abstract-sql') as typeof _abstractSql;
+	const compiledRequest = compileRequest(translatedRequest);
+	const json = JSON.stringify(compiledRequest.sqlQuery, null, 2);
+	if (outputFile) {
+		fs.writeFileSync(outputFile, json);
+	} else {
+		console.log(json);
```
thgreasi: It would be great if we could also get the `query` prop nicely formatted.
Maybe we could have a specific pretty print option/command for this that prints each prop of the json separately in this case with better formatting.

src/bin/odata-compiler.ts
```diff
@@ -0,0 +1,125 @@
...
+		'postgres',
+	);
+
+program
+	.command('parse <input-url> [output-file]')
+	.description('parse the input OData URL into OData AST')
+	.action(parseOData);
+
+program
+	.command('translate <sbvr-file> <input-url> [output-file]')
```
thgreasi: An example `<input-url>` would be great :pray:

src/bin/odata-compiler.ts
dfunckt: @thgreasi you need to wrap the URL in quotes, like so: `'/device?$orderby=device_name%20asc&$top=1&$select=id'`

Page-:
I'd expect it to work with `/device?$filter=name eq 'somename'` already, but automatically escaping it isn't particularly easy due to some bits having to stay unescaped - I'll have a look why it's not working without escaping as it might be an easy fix

src/bin/odata-compiler.ts
Page-: Yeah, it's to keep things sync so errors are reported nicely, commander doesn't support promises and I don't feel a need to learn/rewrite in another command line parser (this is ported in/converted to typescript from https://github.com/balena-io-modules/odata-compiler just to make it easier to use with the right deps within a pinejs project)

src/bin/odata-compiler.ts
Page-: I don't think anything on this line requires nodejs 8?  That said the package.json engines is really really out of date so I've updated it to the current requirements (and added a note to do a matching change in https://github.com/balena-io/pinejs/pull/264)

src/bin/odata-compiler.ts
Page-: What sort of example are you thinking?  Because it depends pretty much entirely on the sbvr file that you use

Also I just realised that we should accept config and abstract sql model files as input as well in order to match what pinejs.setup would accept a bit more closely and would be required for the output of this to match our api runtime output a bit better (it won't take into account hooks but I don't think that's feasible, certainly not without a hefty time investment)

src/bin/odata-compiler.ts
Page-: Done, with `--json` to force json output

PR #285, opened 2019-05-13
Update pinejs-client-core to 5.5.0 and make use of prepared queries

Page-:
Update pinejs-client-core from 5.3.6 to 5.5.0

Change-type: patch


Page-:
Depends on https://github.com/balena-io/pinejs-client-js/pull/72

Page-:
Nah I haven't measured it, it'll probably be negligible at this point as the step that we skip is all of 1ms at most and whilst it gets called regularly it's not *that* regularly.  Where this will shine is when we override the `prepare` method for the internal client to also pre-prepare later steps in the pipeline as well (eg the odata parsing)

PR #288, opened 2019-05-20
Remove unused functions

Page-:
Change-type: patch


Page-:
I expect because `rewriteODataOptions` is recursive so it is technically used, just only by itself, so it avoids a simple "are there any references" check - I expect at some point in the future it would be caught automatically

PR #294, opened 2019-05-31
Update dependencies

Page-:
Update abstract-sql-compiler from 6.4.2 to 6.5.0
Update odata-parser from 1.0.3 to 1.0.4

Change-type: patch


src/sbvr-api/sbvr-utils.ts
```diff
@@ -969,60 +969,62 @@ export const runURI = (
...
-				this.statusCode = statusCode;
-				return this;
-			},
-			sendStatus: (statusCode: number) => {
-				if (statusCode >= 400) {
-					const ErrorClass =
-						statusCodeToError[statusCode as keyof typeof statusCodeToError];
-					if (ErrorClass != null) {
-						reject(new ErrorClass());
+	return new Promise<PinejsClientCoreFactory.PromiseResultTypes>(
```
Page-: The only actual change here is specifying the promise's generic type (adding the `<PinejsClientCoreFactory.PromiseResultTypes>`), the rest of the changes are due to the line going past a certain length so prettier reformatted the whole thing

PR #295, opened 2019-06-03
Update lodash types

Page-:
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "@types/deep-freeze": "^0.1.1",
     "@types/express": "^4.16.1",
     "@types/express-session": "^1.15.12",
-    "@types/lodash": "4.14.123",
+    "@types/lodash": "^4.14.133",
     "@types/memoizee": "0.4.2",
     "@types/method-override": "0.0.31",
     "@types/multer": "1.3.7",
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -255,9 +255,13 @@ export function nestedCheck<I, O>(
 	throw new Error('Cannot parse required checks: ' + check);
 }
 
+interface CollapsedFilter<T> extends Array<string | T | CollapsedFilter<T>> {
+	0: string;
+}
+
 const collapsePermissionFilters = <T>(
 	v: NestedCheck<{ filter: T }>,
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -35,7 +35,7 @@
     "@types/deep-freeze": "^0.1.1",
     "@types/express": "^4.16.1",
     "@types/express-session": "^1.15.12",
-    "@types/lodash": "4.14.123",
+    "@types/lodash": "^4.14.133",
```
thgreasi: Are you so confident about their stability and decided to use `^`?
I'm fine with it, but I'm still hesitant...

src/sbvr-api/permissions.ts
```diff
@@ -255,21 +255,29 @@ export function nestedCheck<I, O>(
...
-			.map(collapsePermissionFilters)
-			.value() as [string, T];
+		if ('and' in v) {
+			return ['and', ...v.and.map(collapsePermissionFilters)];
+		}
+		if ('or' in v) {
+			return ['or', ...v.or.map(collapsePermissionFilters)];
+		}
+		throw new Error(
+			'Permission filter objects must have `filter` or `and` or `or` keys',
```
pdcastro: Does this change still count as `Change-type: patch`? (Not saying it doesn't, just a thought...)

Page-: It should do, it should be impossible to ever hit this branch because it's based on the output of `nestedCheck` so right now this is unreachable code, but it's here for typescript to verify the types properly and also to make sure that other changes within pinejs don't break things in non-obvious ways (ie they'll hit this error instead of weirdness around `undefined` where it's unexpected)

package.json
Page-: Using a fixed version can cause problems downstream where you have different versions of the same types so having a floating version makes this easier for downstream, and since we use a declaration file we should be safe from a floating version on publish and it'll only bite us on rerunning publish with newer types.  Overall bluebird and lodash types are by far the most problematic but it's a really awkward trade off with floating versions and downstream - I wrote https://github.com/balena-io/pinejs/pull/296 primarily to make dealing with differing bluebird types much less of an issue for downstream.

Anyway, I'm not confident about either bluebird or lodash types, but the floating version has some benefits for downstream life

PR #298, opened 2019-06-07
Update odata-parser to make use of its types

Page-:
Update odata-parser from 1.0.4 to 1.2.0
Update odata-to-abstract-sql from 3.2.0 to 3.3.0
Update sbvr-parser from 0.2.2 to 0.2.3

Change-type: patch


typings/sbvr-parser.d.ts
```diff
@@ -1,10 +1,4 @@
 declare module '@resin/sbvr-parser' {
-	import {
```
Page-: This was never needed and I presume was from copy/pasting from another .d.ts file

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1067,9 +1067,12 @@ export const getAffectedIds = Promise.method(
 				}
 				const { idField } = resourceTable;
 
-				_.set(request.odataQuery, ['options', '$select'], {
+				if (request.odataQuery.options == null) {
```
Page-: This change should be marginally better performance but primarily allows us to make use of the typings to validate the
```ts
request.odataQuery.options.$select = ...
```
assignment and also to avoid the need for a `!` in
```ts
delete request.odataQuery.options.$expand;
````

File: package.json [hunk 3]
```diff
@@ -74,19 +74,19 @@
     "grunt-text-replace": "^0.4.0",
     "grunt-ts": "^6.0.0-beta.22",
     "grunt-webpack": "^3.1.3",
-    "husky": "^2.3.0",
-    "lint-staged": "^8.1.7",
+    "husky": "^2.4.0",
+    "lint-staged": "^8.2.0",
     "load-grunt-tasks": "^4.0.0",
-    "prettier": "^1.17.1",
+    "prettier": "^1.18.1",
```
Page-: This is the cause of the whitespace changes btw

PR #307, opened 2019-07-05
Schedule clearing compile cache on any use

Page-:
This means the cache will also be freed in the cache hit case

Change-type: patch

### DIFF
--- a/src/sbvr-api/cached-compile.ts
+++ b/src/sbvr-api/cached-compile.ts
@@ -15,6 +15,8 @@ try {
 	fs = require('fs');
 } catch (e) {}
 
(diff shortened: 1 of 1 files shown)


src/sbvr-api/cached-compile.ts
```diff
@@ -53,5 +54,10 @@ export const cachedCompile = <T>(
 		_.set(cache, key, result);
 		saveCache();
 	}
+	// Schedule clearing the cache once we have made use of it since it usually means we're
+	// done with it and it allows us to free up the memory - if it does end  up being
```
zvin: extra space between "end" and "up"

PR #312, opened 2019-08-23
Accept `instanceof Error` as valid errors

Page-:
This fixes issues where `err.message` is not a string

Change-type: patch


Page-:
@dfunckt no it doesn't: https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L11591

Page-:
@balena-ci retest

PR #314, opened 2019-09-11
Update dependencies

Page-:
Update abstract-sql-compiler from 6.8.0 to 6.8.1 (https://github.com/balena-io-modules/abstract-sql-compiler/pull/64)

Change-type: patch


Page-:
Locking types was a reaction to in the past when updating types pretty much always broke things but thankfully that hasn't really been the case in a long time - we can lock them again if it does prove to be a problem though

PR #316, opened 2019-09-17
permissions: Set actor on api key

roman-mazur:
Invocation of actor ID lookup by api key is moved upstream,
so it can be stored on the apiKey object and reused.

This way, we can use actor ID in both session tokek and api key
authentication cases without extra DB query.

Change-type: minor
Signed-off-by: Roman Mazur <roman@balena.io>

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -956,18 +956,20 @@ const checkApiKey = Promise.method((req: PermissionReq, apiKey: string) => {
 	if (apiKey == null || req.apiKey != null) {
 		return;
 	}
-	return getApiKeyPermissions(apiKey)
-		.catch(err => {
-			console.warn('Error with API key:', err);
-			// Ignore errors getting the api key and just use an empty permissions object
-			return [];
-		})
-		.then(permissions => {
+	return Promise.join(
+		getApiKeyPermissions(apiKey),
+		getApiKeyActorId(apiKey),
+		(permissions, actor) => {
 			req.apiKey = {
 				key: apiKey,
 				permissions: permissions,
+				actor,
 			};
-		});
+		},
+	).catch(err => {
+		console.warn('Error with API key:', err);
+		// Ignore errors getting the api key.
+	});
 });
 
 export const customAuthorizationMiddleware = (expectedScheme = 'Bearer') => {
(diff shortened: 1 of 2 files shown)


roman-mazur:
Not tested yet - trying in the context of access logs enhancements.

src/sbvr-api/permissions.ts
```diff
@@ -1086,53 +1088,41 @@ const getGuestPermissions = memoize(
 );
 
 const getReqPermissions = (req: PermissionReq, odataBinds: ODataBinds = []) =>
-	Promise.join(
```
Page-: This is a breaking change because it's possible to provide a `req.apiKey` to internal requests and previously it was not required to have an actor set but now it is

src/sbvr-api/permissions.ts
```diff
@@ -950,18 +950,20 @@ const checkApiKey = Promise.method((req: PermissionReq, apiKey: string) => {
...
-		.catch(err => {
-			console.warn('Error with API key:', err);
-			// Ignore errors getting the api key and just use an empty permissions object
-			return [];
-		})
-		.then(permissions => {
+	return Promise.join(
+		getApiKeyPermissions(apiKey),
+		getApiKeyActorId(apiKey),
+		(permissions: any, actor: number) => {
```
Page-: If you remove the explicit typings then the types should be automatically inferred correctly (and will be more precise than `any`)

src/sbvr-api/permissions.ts
```diff
@@ -950,18 +950,20 @@ const checkApiKey = Promise.method((req: PermissionReq, apiKey: string) => {
...
 			req.apiKey = {
 				key: apiKey,
 				permissions: permissions,
+				actor,
 			};
-		});
+		},
+	).catch(err => {
+		console.warn('Error with API key:', err);
+		// Ignore errors getting the api key.
```
Page-: This now ignores errors in a different way, such that `req.apiKey` will no longer be set in the case no permissions could be fetched, I wouldn't expect it to break anything but it does have that potential

src/sbvr-api/permissions.ts
roman-mazur: You mean it's possible pass a request object that has `apiKey` but didn't go through `checkApiKey`?

roman-mazur: I found a case in https://github.com/balena-io/open-balena-api/blob/v0.31.0/src/platform/jwt.ts#L171-L174
Do you know about other examples?

Do you think doing an optional actor ID lookup if it's not set on the request object is better or just upgrading open-balena-api?

src/sbvr-api/permissions.ts
roman-mazur: Not sure I follow. `apiKey` will not be set either if there is an error with `getApiKeyPermissions` (as in the previous version) or with `getApiKeyActorId` (the new behavior). But every api key should be associated with an actor, to my understading.

src/sbvr-api/permissions.ts
roman-mazur: Done

src/sbvr-api/permissions.ts
Page-: I think doing the optional lookup and then removing that in 11.x would be better

src/sbvr-api/permissions.ts
roman-mazur: ok... let me guys know what I should do here :)

src/sbvr-api/permissions.ts
roman-mazur: Done in 089170b47d766a49048bd36bc69b7f12a3a5470d PTAL

src/sbvr-api/permissions.ts
```diff
@@ -1089,15 +1091,17 @@ const getReqPermissions = (req: PermissionReq, odataBinds: ODataBinds = []) =>
...
-				req.apiKey != null &&
-				req.apiKey.permissions != null &&
-				req.apiKey.permissions.length > 0
-			) {
-				return getApiKeyActorId(req.apiKey.key);
+			// TODO: Remove this extra actor ID lookup making actor non-optional and updating open-balena-api.
+			if (req.apiKey == null) {
+				return null;
+			}
+			if (req.apiKey.actor == null) {
```
Page-: I'd join this with the other check as
```suggestion
			if (req.apiKey != null && req.apiKey.actor == null) {
```

src/sbvr-api/permissions.ts
Page-: I've just realised that it checks `req.apiKey != null` above and does an early return, so with this change it would mean we check the api key multiple times since we no longer set that property, so from a performance perspective it's probably worth keeping at least something we can check against

roman-mazur: It should be kept unchanged now (moved catch to get permissions call).

src/sbvr-api/permissions.ts
roman-mazur: Addressed.

roman-mazur:
@balena-ci retest, please

@Page- all points seem to be addressed now

src/sbvr-api/permissions.ts
```diff
@@ -950,18 +950,21 @@ const checkApiKey = Promise.method((req: PermissionReq, apiKey: string) => {
...
+	return Promise.join(
+		getApiKeyPermissions(apiKey).catch(err => {
 			console.warn('Error with API key:', err);
-			// Ignore errors getting the api key and just use an empty permissions object
+			// Ignore errors getting the api key and just use an empty permissions object.
 			return [];
-		})
-		.then(permissions => {
+		}),
+		getApiKeyActorId(apiKey),
```
Page-: This will now throw instead @roman-mazur which is also a breaking change. I think to be backwards compatible it needs to be:
```ts
		getApiKeyPermissions(apiKey)
		.catch(err => {
			console.warn('Error with API key:', err);
			// Ignore errors getting the api key and just use an empty permissions object.
			return [];
		})
		.then(permissions => {
			Promise.try(() => {
				if (permissions.length > 0) {
					getApiKeyActorId(apiKey)
				}
			}).then(actor => {
				req.apiKey = {
					key: apiKey,
					permissions,
					actor,
				};
			})
		})

File: src/sbvr-api/permissions.ts [hunk 6]
```diff
@@ -1089,15 +1092,14 @@ const getReqPermissions = (req: PermissionReq, odataBinds: ODataBinds = []) =>
...
 		getGuestPermissions(),
 		Promise.try(() => {
-			if (
-				req.apiKey != null &&
-				req.apiKey.permissions != null &&
-				req.apiKey.permissions.length > 0
-			) {
-				return getApiKeyActorId(req.apiKey.key);
+			// TODO: Remove this extra actor ID lookup making actor non-optional and updating open-balena-api.
+			if (req.apiKey != null && req.apiKey.actor == null) {
```
Page-: And on that logic this also needs to be
```suggestion
			if (req.apiKey != null && req.apiKey.actor == null && req.apiKey.permissions != null && req.apiKey.permissions.length > 0) {
```

src/sbvr-api/permissions.ts
roman-mazur: @Page- sorry, but I'm lost
Where does it throw?
`getApiKeyActorId` is a function that returns a promise.
And why do we need a check for `permissions.length > 0`? Is it for optimization? But how can we make actor non-optional later in such case?

roman-mazur: Actor should be non-optional in practice, shouldn't it? Otherwise, the is no point in having apiKey set on a request at all.

src/sbvr-api/permissions.ts
roman-mazur: Asked above: I don't understand why we need this permissions length check.

src/sbvr-api/permissions.ts
roman-mazur: Please take another look.

roman-mazur:
@balena-ci retest please

PR #321, opened 2019-10-14
Reuse a single permission error to reduce cached size

Page-:
Change-type: patch

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -387,6 +387,7 @@ const $checkPermissions = (
 	});
 };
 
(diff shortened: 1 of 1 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -925,13 +928,16 @@ const getApiKeyActorIdQuery = _.once(() =>
 		},
 	}),
 );
+const apiActorPermissionError = new PermissionError();
```
Page-: We use two separate errors to get unique stack traces/names so it's possible to see from which of the two places it came from when debugging

PR #323, opened 2019-11-04
Update dependencies to enable enum definitions to check constraint

Page-:
Update abstract-sql-compiler from 6.9.2 to 6.10.1
Update lf-to-abstract-sql from 2.1.0 to 2.2.0

Change-type: minor


Page-:
@balena-ci retest

PR #329, opened 2019-12-02
Explain Pine's SBVR in docs/SBVR.md

Ereski:
- Explain SBVR as understood by Pine in more programmer-friendly
  terms
- Map out how Pine interprets SBVR's structured english


docs/SBVR.md
```diff
@@ -0,0 +1,355 @@
...
+
+SBVR files are composed of statements, one per line, in the form `<header>: <body>`. Indentation is not significant, and `--` introduces comments that go until the end of the line. SBVR files accepted by Pine.js are case-insensitive and usually have the `.sbvr` extension. `Vocabulary`, `Term`, `Fact Type`, and `Rule` are toplevel declarations that can be followed by _attributes_. See the **Attributes** subsection below.
+
+A vocabulary can be started with:
+
+```
+Vocabulary: <vocabulary-name>
+```
+
+Every new definition is part of the last declared vocabulary. If no vocabulary has been explicitly declared, a `Default` vocabulary is implicitly assumed. Within vocabularies, term declarations introduces _types_:
```
nazrhom: This could be expanded a bit here: in what sense is a term a type?

docs/SBVR.md
```diff
@@ -0,0 +1,355 @@
...
+
+```
+Term: <term-name>
+```
+
+where `<term-name>` is the name of the term. Because SBVR focuses on structuring natural languages, terms can have spaces in their names. By default, every instance of a term is identified by an `id` and persisted into a table.
```
nazrhom: Maybe we can phrase this differently, as Concept-types (introduced later) make this technically not true (and its not just an implementation detail, most of the terms you define are in fact columns). The intuition to convey is that terms are the building blocks of our objects, and terms themselves can be formed by composing other terms. In the context of a relational DB it is easy to see how this maps to tables and columns.

docs/SBVR.md
```diff
@@ -0,0 +1,355 @@
...
+Term: bar
+    Concept Type: Text (Type)
+
+Fact Type: foo has bar
+    Necessity: each foo has exactly one bar
+```
+
+### Database Table Name
+
+The `Database Table Name` defines the name of the database table.
```
nazrhom: What implication does this have? Can this attribute only be applied to `Terms`?

docs/SBVR.md
```diff
@@ -0,0 +1,355 @@
...
+| Variation | Equivalent to |
+| --- | --- |
+| `It is impossible that` | It is necessary that not |
+| `It is not possible that` | It is necessary that not |
+| `It is possible that` | It is not necessary that not |
+| `It is prohibited that` | It is obligatory that not |
+| `It is forbidden that` | It is obligatory that not |
+| `It is permitted that` | It is not obligatory that not |
+
+Currently Pine parses modalities but assumes they are all equivalent to necessities. It also assumes that rules that can be represented directly in SQL (schema, UNIQUE/NOT NULL/etc) are alethic, and everything else deontic.
```
nazrhom: This is interesting, never fully understood how we differentiate between the two. Can we add an example of a deontic rule here? Also, we should explicitly mention the runtime cost that is associated with the two: as the first comes for "free" since they are enforced by the DB, but we have to explicitly check for other rules after an insert/update. What happens in practice if a deontic rule is violated?

Ereski: Within SBVR, the difference is about what to expect: violations of alethic rules are not expected at all, while violations of deontic rules are not desired but possible. As far as the database is concerned, the difference is how you get an error: either the DB refuses to work (alethic, can't do anything about that unless you change the schema behind Pine's back) or a a validation check fails (deontic, user-driver check that can be ignored). I haven't gone down the code path dealing with validation checks though.

It's hard to talk about runtime costs and the alethic/deontic split in practice because it depends on the DB, the Pine version, and how you phrase your constraint. The basic gist seems to be that if Pine sees an opportunity to encode a constraint into the schema, it takes it. Everything else is put into validation checks.

But regardless, SBVR modalities are completely irrelevant in practice. You could scrap them without losing functionality (even `impossible` is treated as `necessary`), so I don't think it makes much sense to delve into that.

docs/SBVR.md
Ereski: Here's the explanation I put:

Terms are _types_: they represent a collection of instances that obey an uniform "interface". And that is defined through fact types and rules.

PR #332, opened 2020-01-09
Switch to SQL migrations to enable migrating internal models

Page-:
This is used in https://github.com/balena-io/pinejs/pull/264 to allow automatic migration to 11.x

Change-type: minor

### DIFF
--- a/src/migrator/migrator.ts
+++ b/src/migrator/migrator.ts
@@ -6,8 +6,9 @@ const modelText: string = require('./migrations.sbvr');
 import { Tx } from '../database-layer/db';
 import * as sbvrUtils from '../sbvr-api/sbvr-utils';
 import { migrator as migratorEnv } from '../config-loader/env';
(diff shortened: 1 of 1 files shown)


src/migrator/migrator.ts
```diff
@@ -21,6 +22,24 @@ export type Migration = string | MigrationFn;
 
 export class MigrationError extends TypedError {}
 
+// Tagged template to convert binds from `?` format to the necessary output format,
+// eg `$1`/`$2`/etc for postgres
+const binds = (strings: TemplateStringsArray, ...binds: number[]) =>
```
dfunckt: I don't think this works as intended or I miss something: given `S`, this will always return `S` as is for all strings I see in this PR.

dfunckt: Cool, yeah, I had to look up about tagged templates and what I read didn't add up but I wasn't sure whether it was me not misunderstanding.

PR #333, opened 2020-01-10
Remove unnecessary Promise.method

Page-:


src/sbvr-api/hooks.ts
```diff
@@ -50,24 +50,24 @@ export class SideEffectHook extends Hook {
...
-				}
-			},
-		);
-	},
-);
+export const rollbackRequestHooks = <T extends InstantiatedHooks<any>>(
+	hooks: T | undefined,
+): void => {
+	if (hooks == null) {
+		return;
```
pdcastro: In another PR, you had explained that hook functions don't need to return promises... I guess it applies here as well, because `Promise.method` was removed.

Page-: This is actually a different situation, but here we are always returning void and never anything to actually wait on (and we never tried to wait on it either) so the `Promise.method` was unnecessary/confusing and would also have been a slight performance hit, probably neglible but still better to avoid

PR #334, opened 2020-01-14
Improve tx event handlers

Page-:
* Run tx event handlers asynchronously
* Clear queued tx event handlers once they've been scheduled to run
* Immediately run tx event handlers if the tx event has already triggered

### DIFF
--- a/src/database-layer/db.ts
+++ b/src/database-layer/db.ts
@@ -106,8 +106,8 @@ const atomicExecuteSql: Database['executeSql'] = function(sql, bindings) {
 	return this.transaction(tx => tx.executeSql(sql, bindings));
 };
 
-const tryFn = (fn: () => any) => {
-	Promise.try(fn);
+const asyncTryFn = (fn: () => any) => {
+	Promise.resolve().then(fn);
 };
 
 let timeoutMS: number;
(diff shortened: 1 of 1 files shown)


src/database-layer/db.ts
```diff
@@ -106,8 +106,8 @@ const atomicExecuteSql: Database['executeSql'] = function(sql, bindings) {
 	return this.transaction(tx => tx.executeSql(sql, bindings));
 };
 
-const tryFn = (fn: () => any) => {
-	Promise.try(fn);
+const asyncTryFn = (fn: () => any) => {
+	Promise.resolve().then(fn);
```
roman-mazur: What is the actual difference with the previous variant?

src/database-layer/db.ts
```diff
@@ -237,7 +248,9 @@ export abstract class Tx {
 	}
 	public rollback(): Promise<void> {
 		const promise = this._rollback().finally(() => {
-			this.listeners.rollback.forEach(tryFn);
+			this.listeners.rollback.forEach(asyncTryFn);
```
roman-mazur: If I read this correctly, errors produced by the rollback or end listeners will not be handled, will they (because here we create new promises that nobody subscribes to)? Is it by design?

src/database-layer/db.ts
Page-: This will schedule `fn` to be called asynchronously rather than synchronously

src/database-layer/db.ts
Page-: Yes it's by design, we want to fire/forget the errors and it's on the subscriber to handle any errors they have

wrboyce:
LGTM

PR #335, opened 2020-01-16
✨: Add support for `canAccess()` function in permission.

afitzek:
This will allow to write simpler permissions, which point
to access depending on other resources access.

Change-type: minor

Depends-On: https://github.com/balena-io-modules/odata-parser/pull/45
Depends-On: https://github.com/balena-io-modules/odata-to-abstract-sql/pull/48

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,8 @@
   "dependencies": {
     "@resin/abstract-sql-compiler": "^6.13.1",
     "@resin/lf-to-abstract-sql": "^3.1.1",
-    "@resin/odata-parser": "^1.3.1",
-    "@resin/odata-to-abstract-sql": "^4.3.0",
+    "@resin/odata-parser": "^1.4.0",
+    "@resin/odata-to-abstract-sql": "^4.4.0",
     "@resin/sbvr-parser": "^0.2.7",
     "@resin/sbvr-types": "^2.0.4",
     "@types/bluebird": "^3.5.29",
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -1,4 +1,12 @@
 import {
+	Definition,
+	OData2AbstractSQL,
+	odataNameToSqlName,
+	ResourceFunction,
+	sqlNameToODataName,
+} from '@resin/odata-to-abstract-sql';
+
+import {
 	AbstractSqlModel,
 	AbstractSqlType,
 	AliasNode,
@@ -7,15 +15,13 @@ import {
 } from '@resin/abstract-sql-compiler';
 import * as ODataParser from '@resin/odata-parser';
 import { ODataBinds, ODataQuery, SupportedMethod } from '@resin/odata-parser';
-import {
-	Definition,
-	odataNameToSqlName,
-	sqlNameToODataName,
-} from '@resin/odata-to-abstract-sql';
+
 import * as Bluebird from 'bluebird';
+
 import * as express from 'express';
 import * as _ from 'lodash';
 import * as memoize from 'memoizee';
+import * as randomstring from 'randomstring';
 import * as env from '../config-loader/env';
 import * as sbvrUtils from '../sbvr-api/sbvr-utils';
 import {
@@ -24,7 +30,6 @@ import {
 	PermissionParsingError,
 } from './errors';
 import { AnyObject, ApiKey, HookReq, User } from './sbvr-utils';
-import * as uriParser from './uri-parser';
 import {
 	memoizedParseOdata,
 	metadataEndpoints,
@@ -387,12 +392,60 @@ const $checkPermissions = (
 	});
 };
 
-const constrainedPermissionError = new PermissionError();
-const generateConstrainedAbstractSql = (
+const convertToLambda = (filter: AnyObject, identifier: string) => {
+	// We need to inject all occurences of properties for the new lambda function
+	// for example if there is an `or` operator, where different properties of
+	// the target resource are used, we need to change all occurences of these.
+	// We do this in a recursive way to also cover all nested cases like or [ and, and ]
+	const replaceObject = (object: AnyObject) => {
+		if (typeof object === 'string') {
+			return;
+		}
+		if (Array.isArray(object)) {
+			object.forEach(element => {
+				replaceObject(element);
+			});
+		}
+
+		if (object.hasOwnProperty('name')) {
+			object.property = _.clone(object);
+			object.name = identifier;
+			delete object.lambda;
+		}
+	};
+
+	replaceObject(filter);
+};
+
+const rewriteSubPermissionBindings = (filter: AnyObject, counter: number) => {
+	const rewrite = (object: AnyObject) => {
+		if (object == null) {
+			return;
+		}
+
+		if (typeof object.bind === 'number') {
+			object.bind = counter + object.bind;
+		}
+
+		if (Array.isArray(object) || _.isObject(object)) {
+			_.forEach(object, v => {
+				rewrite(v);
+			});
+		}
+	};
+
+	rewrite(filter);
+};
+
+const buildODataPermission = (
 	permissionsLookup: PermissionLookup,
 	actionList: PermissionCheck,
 	vocabulary: string,
 	resourceName: string,
+	odata: {
+		tree: ODataParser.ODataQuery;
+		binds: ODataParser.ODataBinds;
+	},
 ) => {
 	const conditionalPerms = $checkPermissions(
 		permissionsLookup,
@@ -409,8 +462,6 @@ const generateConstrainedAbstractSql = (
 		// If we have full access then no need to provide a constrained definition
 		return false;
 	}
-
-	const odata = memoizedParseOdata(`/${resourceName}`);
 
 	const permissionFilters = nestedCheck(conditionalPerms, permissionCheck => {
 		try {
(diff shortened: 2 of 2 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -24,7 +27,8 @@ import {
 	PermissionParsingError,
 } from './errors';
 import { AnyObject, ApiKey, HookReq, User } from './sbvr-utils';
-import * as uriParser from './uri-parser';
+//import * as uriParser from './uri-parser';
```
Page-: debug

src/sbvr-api/permissions.ts
```diff
@@ -430,16 +472,109 @@ const generateConstrainedAbstractSql = (
 	const collapsedPermissionFilters = collapsePermissionFilters(
 		permissionFilters,
 	);
+
+	return collapsedPermissionFilters;
+};
+
+const getResourceName = (resourceName: string) =>
```
Page-: There's a `sqlNameToODataName` function for this already

src/sbvr-api/permissions.ts
```diff
@@ -430,16 +472,109 @@ const generateConstrainedAbstractSql = (
...
+			);
+			convertToLamda(collapsedParentPermissionFilters, lId);
+			_.set(property, 'lambda', {
+				method: 'any',
+				identifier: lId,
+				expression: collapsedParentPermissionFilters,
+			});
+
+			compiler.extraBindVars.push(...parentOdata.binds);
+			//odata.binds = _.concat(odata.binds, parentOdata.binds)
```
Page-: debug?

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +391,52 @@ const $checkPermissions = (
 	});
 };
 
-const constrainedPermissionError = new PermissionError();
-const generateConstrainedAbstractSql = (
+const convertToLamda = (filter: AnyObject, identifier: string) => {
+	const replaceOject = (object: AnyObject) => {
+		if (_.isString(object)) {
```
Page-: I think
```suggestion
		if (typeof object === 'string') {
```
is preferable nowadays (typescript validates it from memory)

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +391,52 @@ const $checkPermissions = (
...
+			return;
+		}
+		if (_.isArray(object)) {
+			object.forEach(element => {
+				replaceOject(element);
+			});
+		}
+
+		if (_.has(object, 'name')) {
```
Page-: Are all these `_.has` scattered around actually necessary? It feels like unnecessary lodash usage (lodash is great, it's just it's easy to use even when it's not really necessary)

src/sbvr-api/uri-parser.ts
```diff
@@ -144,7 +144,7 @@ export const memoizedParseOdata = (() => {
 
 const memoizedGetOData2AbstractSQL = memoizeWeak(
 	(abstractSqlModel: _AbstractSQLCompiler.AbstractSqlModel) => {
-		return new OData2AbstractSQL(abstractSqlModel);
+		return new OData2AbstractSQL(abstractSqlModel, {});
```
Page-: You can revert this change - the default is already `{}`

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +391,56 @@ const $checkPermissions = (
...
+			delete object.lambda;
+		}
+	};
+
+	replaceObject(filter);
+};
+
+const rewriteSubPermissionBindings = (filter: AnyObject, counter: number) => {
+	const rewrite = (object: AnyObject) => {
+		if (_.isNil(object)) {
```
Page-:
```suggestion
		if (object == null) {
```

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +391,56 @@ const $checkPermissions = (
...
+			return;
+		}
+
+		if (_.isNumber(object.bind)) {
```
Page-:
```suggestion
		if (typeof object.bind === 'number') {
```

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +391,56 @@ const $checkPermissions = (
...
+			object.bind = counter + object.bind;
+		}
+
+		if (_.isArray(object) || _.isObject(object)) {
```
Page-:
```suggestion
		if (Array.isArray(object) || _.isObject(object)) {
```

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +391,56 @@ const $checkPermissions = (
...
 };
 
-const constrainedPermissionError = new PermissionError();
-const generateConstrainedAbstractSql = (
+const convertToLamda = (filter: AnyObject, identifier: string) => {
+	const replaceObject = (object: AnyObject) => {
+		if (typeof object === 'string') {
+			return;
+		}
+		if (_.isArray(object)) {
```
Page-:
```suggestion
		if (Array.isArray(object)) {
```

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +391,56 @@ const $checkPermissions = (
 	});
 };
 
-const constrainedPermissionError = new PermissionError();
-const generateConstrainedAbstractSql = (
+const convertToLamda = (filter: AnyObject, identifier: string) => {
+	const replaceObject = (object: AnyObject) => {
```
Page-: What is this function for?

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +391,56 @@ const $checkPermissions = (
...
+			object.property = _.cloneDeep(object);
+			object.name = identifier;
+			delete object.lambda;
+		}
+	};
+
+	replaceObject(filter);
+};
+
+const rewriteSubPermissionBindings = (filter: AnyObject, counter: number) => {
```
Page-: Isn't this equivalent to `rewriteBinds` in odata-to-abstract-sql?

src/sbvr-api/permissions.ts
```diff
@@ -430,17 +476,116 @@ const generateConstrainedAbstractSql = (
...
+			const targetResource = compiler.NavigateResources(
+				compiler.defaultResource,
+				property.name,
+			);
+
+			const targetResourceName = sqlNameToODataName(
+				targetResource.resource.name,
+			);
+
+			if (_.findIndex(canAccessTrace, v => targetResourceName === v) >= 0) {
```
Page-: What is this for? It seems like it should be a `.includes` to me

src/sbvr-api/permissions.ts
```diff
@@ -430,17 +476,116 @@ const generateConstrainedAbstractSql = (
...
+				parentOdata,
+			);
+
+			if (collapsedParentPermissionFilters === false) {
+				// We reuse a constant permission error here as it will be cached, and
+				// using a single error instance can drastically reduce the memory used
+				throw constrainedPermissionError;
+			}
+
+			const lId = `${lambdaAlias}+${inc}`;
```
Page-: Is `lId` short for `lambdaId`?

src/sbvr-api/permissions.ts
```diff
@@ -430,17 +476,116 @@ const generateConstrainedAbstractSql = (
...
+			inc = inc + 1;
+			rewriteSubPermissionBindings(
+				collapsedParentPermissionFilters,
+				compiler.bindVarsLength + compiler.extraBindVars.length,
+			);
+			convertToLamda(collapsedParentPermissionFilters, lId);
+			_.set(property, 'lambda', {
```
Page-: This can be simplified to
```suggestion
			property.lambda = {
```
right?

src/sbvr-api/permissions.ts
afitzek: I'll add a comment to it as well.
Basically we need to inject all occurences of properties for the new lambda function
for example if there is an `or` operator, where different properties of
the target resource are used, we need to change all occurences of these.
We do this in a recursive way to also cover all nested cases like `or [ and, and ]`

src/sbvr-api/permissions.ts
afitzek: I think the `rewriteBinds` in `odata-to-abstract-sql` operates on the abstract sql representation, where this operates on the parsed odata representation. So this happens on the odata level before transforming it into abstract sql.

src/sbvr-api/permissions.ts
```diff
@@ -430,17 +481,117 @@ const generateConstrainedAbstractSql = (
...
+			method: 'any',
+			identifier: lambdaId,
+			expression: collapsedParentPermissionFilters,
+		};
+
+		this.extraBindVars.push(...parentOdata.binds);
+
+		canAccessTrace.push(targetResourceName);
+		const value = this.Property(property);
+		_.remove(canAccessTrace, v => targetResourceName === v);
```
Page-: Shouldn't this be just a `canAccessTrace.pop()` as it's a stack?

src/sbvr-api/permissions.ts
```diff
@@ -430,17 +481,117 @@ const generateConstrainedAbstractSql = (
...
+
+		const targetResource = this.NavigateResources(
+			this.defaultResource,
+			property.name,
+		);
+
+		const targetResourceName = sqlNameToODataName(targetResource.resource.name);
+
+		if (_.includes(canAccessTrace, targetResourceName)) {
+			// return a boolen subquery
```
Page-: Why does it need to be a subquery, does returning `['Boolean', true/false]` not work?

src/sbvr-api/permissions.ts
Page-: Ah, so this handles adding the lambda alias where previously it didn't exist, that makes sense

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +392,60 @@ const $checkPermissions = (
...
+			return;
+		}
+		if (Array.isArray(object)) {
+			object.forEach(element => {
+				replaceObject(element);
+			});
+		}
+
+		if (object.hasOwnProperty('name')) {
+			object.property = _.cloneDeep(object);
```
Page-: Does this need to be a `cloneDeep`? I think a simple `clone` would work just as well (and be more efficient)

package.json
```diff
@@ -24,8 +24,8 @@
   "dependencies": {
     "@resin/abstract-sql-compiler": "^6.13.1",
     "@resin/lf-to-abstract-sql": "^3.1.1",
-    "@resin/odata-parser": "^1.3.1",
-    "@resin/odata-to-abstract-sql": "^4.3.0",
+    "@resin/odata-parser": "1.4.0",
+    "@resin/odata-to-abstract-sql": "4.4.0-allow-resource-function-registration-058b3b3ec8bf9dd4da68005ec4a4e5caae77f800",
```
Page-: I think this can be updated now

src/sbvr-api/permissions.ts
afitzek: No it does not, I have a boolean entry first, but if I remember correctly, it translates a boolean entry into `1` or `0`. If this is used in an `and` statement it becomes in SQL:

```
... AND 1 ...
```
which is not valid SQL.

src/sbvr-api/permissions.ts
afitzek: You are probably right. I'll give it a try.

src/sbvr-api/permissions.ts
Page-: Ah, well `['Equals', ['Boolean', true], ['Boolean', true]]` should work and still be somewhat preferable

src/sbvr-api/permissions.ts
```diff
@@ -430,17 +481,117 @@ const generateConstrainedAbstractSql = (
...
+		'GET',
+		[],
+		odata.binds.length,
+	);
+
+	let abstractSqlQuery = tree;
+	odata.binds.push(...extraBindVars);
+	const odataBinds = odata.binds;
+
+	abstractSqlQuery = _.clone(abstractSqlQuery!);
```
Page-:
```suggestionS
	odata.binds.push(...extraBindVars);
	const odataBinds = odata.binds;

	const abstractSqlQuery = _.clone(tree!);
```

src/sbvr-api/permissions.ts
```diff
@@ -430,17 +481,117 @@ const generateConstrainedAbstractSql = (
...
+			identifier: lambdaId,
+			expression: collapsedParentPermissionFilters,
+		};
+
+		this.extraBindVars.push(...parentOdata.binds);
+
+		canAccessTrace.push(targetResourceName);
+		const value = this.Property(property);
+		canAccessTrace.pop();
+		return value;
```
Page-: Using a try/finally would make this safer from edge cases/future changes:
```suggestion
		canAccessTrace.push(targetResourceName);
		try {
			return this.Property(property);
		} finally {
			canAccessTrace.pop();
		}
```

src/sbvr-api/permissions.ts
Page-: Bump

afitzek: Ah missed that.

src/sbvr-api/permissions.ts
```diff
@@ -430,17 +481,112 @@ const generateConstrainedAbstractSql = (
...
+	const lambdaAlias = randomstring.generate(20);
+	let inc = 0;
+
+	// We need to trace the processed resources, to be able to break
+	// permissions circles.
+	const canAccessTrace: string[] = [resourceName];
+
+	const canAccessFunction: ResourceFunction = function(property: AnyObject) {
+		// remove method property so that we won't loop back here again at this point
+		delete property.method;
```
Page-: If we know we can only be here for a method then we can probably get better typings for `property`

src/sbvr-api/permissions.ts
```diff
@@ -430,17 +481,112 @@ const generateConstrainedAbstractSql = (
...
+		rewriteSubPermissionBindings(
+			collapsedParentPermissionFilters,
+			this.bindVarsLength + this.extraBindVars.length,
+		);
+		convertToLambda(collapsedParentPermissionFilters, lambdaId);
+		property.lambda = {
+			method: 'any',
+			identifier: lambdaId,
+			expression: collapsedParentPermissionFilters,
+		};
```
Page-: Ideally this would be handled by the better typings for property but if not is it possible to add an `as LambdaProperty` or similar type (assuming there's one available? if not then it'll have to stay as-is - I know the odata-parser types are still fairly loose)

src/sbvr-api/permissions.ts
```diff
@@ -430,17 +481,112 @@ const generateConstrainedAbstractSql = (
...
+
+		const targetResource = this.NavigateResources(
+			this.defaultResource,
+			property.name,
+		);
+
+		const targetResourceName = sqlNameToODataName(targetResource.resource.name);
+
+		if (canAccessTrace.includes(targetResourceName)) {
+			// return a boolen subquery
```
Page-:
```suggestion
			// return a boolean subquery that is always false
```
(or `true` and update the returned query if that was the real intention)

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +392,60 @@ const $checkPermissions = (
...
+		if (typeof object === 'string') {
+			return;
+		}
+		if (Array.isArray(object)) {
+			object.forEach(element => {
+				replaceObject(element);
+			});
+		}
+
+		if (object.hasOwnProperty('name')) {
```
Page-: It doesn't make sense to check `object.hasOwnProperty('name')` if `object` is actually an array
```suggestion
		if (Array.isArray(object)) {
			object.forEach(element => {
				replaceObject(element);
			});
		} else if (object.hasOwnProperty('name')) {
```

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +392,60 @@ const $checkPermissions = (
...
 };
 
-const constrainedPermissionError = new PermissionError();
-const generateConstrainedAbstractSql = (
+const convertToLambda = (filter: AnyObject, identifier: string) => {
+	// We need to inject all occurences of properties for the new lambda function
+	// for example if there is an `or` operator, where different properties of
+	// the target resource are used, we need to change all occurences of these.
+	// We do this in a recursive way to also cover all nested cases like or [ and, and ]
+	const replaceObject = (object: AnyObject) => {
```
Page-: It seems like the more accurate typing here would be:
```suggestion
	const replaceObject = (object: string | AnyObject | AnyObject[]) => {
```

src/sbvr-api/permissions.ts
```diff
@@ -387,12 +392,60 @@ const $checkPermissions = (
 	});
 };
 
-const constrainedPermissionError = new PermissionError();
-const generateConstrainedAbstractSql = (
+const convertToLambda = (filter: AnyObject, identifier: string) => {
```
Page-: I think renaming to `lamdaAlias` would make the code a little clearer
```suggestion
const convertToLambda = (filter: AnyObject, lambdaAlias: string) => {
```

src/sbvr-api/permissions.ts
afitzek: We don't have such a type yet unfortunately.

PR #341, opened 2020-01-29
Automatically fetch affected ids when possible (postgres only)

Page-:
Example usage: https://github.com/balena-io/open-balena-api/pull/265

Change-type: minor


Page-:
@balena-ci rebase

PR #343, opened 2020-01-31
Convert some internal functions to async await

Page-:


src/config-loader/config-loader.ts
```diff
@@ -52,294 +52,267 @@ export interface Config {
...
+						const result = (await authApiTx.get({
+							resource: 'user',
+							options: {
+								$select: 'id',
+								$filter: {
+									username: user.username,
+								},
+							},
+						})) as Array<{ id: number }>;
+						let userID: number = result[0].id;
```
thgreasi: :grey_exclamation:  We are accessing `result[0].id` before checking the length of the `result`.

src/config-loader/config-loader.ts
```diff
@@ -52,294 +52,267 @@ export interface Config {
...
+								const result = (await authApiTx.get({
+									resource: 'permission',
+									options: {
+										$select: 'id',
+										$filter: {
+											name: permissionName,
+										},
+									},
+								})) as Array<{ id: number }>;
+								if (result.length === 0) {
```
thgreasi: How about destructuring the results as `const [result] = ...` and then early return like
```
if (result != null) {
    return result.id;
}
```
to make the code shallower (instead of checking the length)?

src/config-loader/config-loader.ts
```diff
@@ -52,294 +52,267 @@ export interface Config {
...
-								(model.abstractSql != null || model.modelText != null) &&
-								model.apiRoot != null
-							) {
-								return sbvrUtils
-									.executeModel(tx, model as sbvrUtils.ExecutableModel)
-									.then(() => {
-										const apiRoute = `/${model.apiRoot}/*`;
-										app.options(apiRoute, (_req, res) => res.sendStatus(200));
-										app.all(apiRoute, sbvrUtils.handleODataRequest);
+								})) as sbvrUtils.AnyObject[];
```
thgreasi:
```suggestion
								})) as Array<{ id: number }>;
```

src/config-loader/config-loader.ts
```diff
@@ -52,294 +52,267 @@ export interface Config {
...
-								delete model.migrationsPath;
+					if (model.migrationsPath) {
+						const migrationsPath = resolvePath(model.migrationsPath);
+						delete model.migrationsPath;
 
-								return Bluebird.resolve(fs.promises.readdir(migrationsPath))
-									.map(filename => {
-										const filePath = path.join(migrationsPath, filename);
-										const [migrationKey] = filename.split('-', 1);
+						await Bluebird.resolve(fs.promises.readdir(migrationsPath)).map(
```
thgreasi: How about
```suggestion
						await Bluebird.map(await fs.promises.readdir(migrationsPath), async filename => {
```
similarly to what you did a bit earlier?

src/sbvr-api/permissions.ts
```diff
@@ -692,26 +692,24 @@ export const checkPassword = (
 }> =>
 	getCheckPasswordQuery()({
 		username,
-	}).then((result: AnyObject[]) => {
+	}).then(async (result: AnyObject[]) => {
```
thgreasi:
```suggestion
	}).then(async ([user]: AnyObject[]) => {
```
and we can stop repeating `result[0]`

src/sbvr-api/permissions.ts
```diff
@@ -920,55 +919,54 @@ const getApiKeyActorIdQuery = _.once(() =>
...
+		if (apiKey == null || req.apiKey != null) {
+			return;
+		}
+		let permissions;
+		try {
+			permissions = await getApiKeyPermissions(apiKey);
+		} catch (err) {
 			console.warn('Error with API key:', err);
 			// Ignore errors getting the api key and just use an empty permissions object.
 			return [];
```
thgreasi: I think that this now has to become:
```suggestion
			permissions = [];
```
Otherwise `req.apiKey = {` will no longer be assigned & the `checkApiKey` will no longer be `Bluebird<void>`

src/sbvr-api/permissions.ts
```diff
@@ -1077,33 +1075,30 @@ const getGuestPermissions = memoize(
...
 			},
-		})
-			.then((result: AnyObject[]) => {
-				if (result.length === 0) {
-					throw new Error('No guest user');
-				}
-				return getUserPermissions(result[0].id);
-			})
-			.then(_.uniq),
+		})) as AnyObject[];
```
thgreasi:
```suggestion
		})) as Array<{ id: number }>;
```

src/sbvr-api/permissions.ts
```diff
@@ -1064,9 +1062,9 @@ export const checkPermissionsMiddleware = (
 		});
 
 const getGuestPermissions = memoize(
-	() =>
+	async () => {
```
thgreasi: Does this need to be `Promise.method` as well for compatibility and consistency with `getReqPermissions`?

Page-: It does not as this one is only used internally in a `bluebird.join` and meas all returned types remain the same bluebird rather than native promises

src/sbvr-api/permissions.ts
Page-: Yes you're 100% correct, good spot

src/sbvr-api/permissions.ts
Page-: Done

src/config-loader/config-loader.ts
Page-: Yeah, that's my mistake - tried to eliminate an else

src/config-loader/config-loader.ts
```diff
@@ -52,294 +52,267 @@ export interface Config {
...
+				} = {};
+				_.each(users, user => {
+					if (user.permissions == null) {
+						return;
+					}
+					_.each(user.permissions, permissionName => {
+						if (permissionsCache[permissionName] != null) {
+							return;
+						}
+						permissionsCache[permissionName] = (async () => {
```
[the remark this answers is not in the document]
Page-: Done (also since this is only done once ever at startup the benefits of avoiding the multiple function creation is limited, but still nice to do especially with deduplicating code)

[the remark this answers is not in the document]
thgreasi: Oh I didn't realize that this was only running once :sweat_smile: 
But as you said the code became so much smaller & readable :+1:

PR #346, opened 2020-02-20
✨: Optimize database queries for `canAccess()` permissions

afitzek:
When a permission uses the `canAccess()` function to derive access,
    to another resource, we can optimize the generated SQL queries to not,
    reapply the full permission for resources, that are reached by
    expanding or filtering paths. We achieve this by, applying shortcuts
    in the relationships of the abstract sql model, if we find usages of
    an `canAccess()` function in the relevant permissions, that link the
    resources in the correct direction.
An example:
Let's imagine a model where a `device` belongs to an `application`.
We have the following permissions:
    
```
    example.application.get?belongs_to__user/any(u:u/actor eq @__ACTOR_ID)
    example.device.get?belongs_to__application/canAccess()
```
Given the following odata query:
```
    GET /example/application?$select=id,name&$filter=(owns__device/any(d:d/name eq 'test'))&$expand=owns__device($select=id,name)
```
Will not yield the following SQL statement anymore:
```
    SELECT (
            SELECT coalesce(array_to_json(array_agg("application.owns-device".*)), '[]') AS "owns__device"
            FROM (
                    SELECT "application.owns-device"."id", "application.owns-device"."name"
                    FROM (
                            SELECT "device"."created at", "device"."id", "device"."belongs to-application", "device"."name"
                            FROM "device"
                            WHERE EXISTS (
                                    SELECT 1
                                    FROM "application" AS "device.belongs to-application"
                                    WHERE "device"."belongs to-application" = "device.belongs to-application"."id"
                                    AND EXISTS (
                                            SELECT 1
                                            FROM "user" AS "device.belongs to-application.belongs to-user"
                                            WHERE "device.belongs to-application"."belongs to-user" = "device.belongs to-application.belongs to-user"."id"
                                            AND ("device.belongs to-application.belongs to-user"."actor") IS NOT NULL AND ("device.belongs to-application.belongs to-user"."actor") = ($1)
                                    )
                            )
                    ) AS "application.owns-device"
                    WHERE "application"."id" = "application.owns-device"."belongs to-application"
            ) AS "application.owns-device"
    ) AS "owns__device", "application"."id", "application"."name"
    FROM (
            SELECT "application"."created at", "application"."id", "application"."name", "application"."belongs to-user"
            FROM "application"
            WHERE EXISTS (
                    SELECT 1
                    FROM "user" AS "application.belongs to-user"
                    WHERE "application"."belongs to-user" = "application.belongs to-user"."id"
                    AND ("application.belongs to-user"."actor") IS NOT NULL AND ("application.belongs to-user"."actor") = ($1)
            )
    ) AS "application"
    WHERE EXISTS (
            SELECT 1
            FROM (
                    SELECT "device"."created at", "device"."id", "device"."belongs to-application", "device"."name"
                    FROM "device"
                    WHERE EXISTS (
                            SELECT 1
                            FROM "application" AS "device.belongs to-application"
                            WHERE "device"."belongs to-application" = "device.belongs to-application"."id"
                            AND EXISTS (
                                    SELECT 1
                                    FROM "user" AS "device.belongs to-application.belongs to-user"
                                    WHERE "device.belongs to-application"."belongs to-user" = "device.belongs to-application.belongs to-user"."id"
                                    AND ("device.belongs to-application.belongs to-user"."actor") IS NOT NULL AND ("device.belongs to-application.belongs to-user"."actor") = ($1)
                            )
                    )
            ) AS "application.owns-device"
            WHERE "application"."id" = "application.owns-device"."belongs to-application"
            AND ("application.owns-device"."name") IS NOT NULL AND ("application.owns-device"."name") = ($3)
    )
```
    
but this optimized version:
```
    SELECT (
            SELECT coalesce(array_to_json(array_agg("application.owns-device".*)), '[]') AS "owns__device"
            FROM (
                    SELECT "application.owns-device"."id", "application.owns-device"."name"
                    FROM "device" AS "application.owns-device"
                    WHERE "application"."id" = "application.owns-device"."belongs to-application"
            ) AS "application.owns-device"
    ) AS "owns__device", "application"."id", "application"."name"
    FROM (
            SELECT "application"."created at", "application"."id", "application"."name", "application"."belongs to-user"
            FROM "application"
            WHERE EXISTS (
                    SELECT 1
                    FROM "user" AS "application.belongs to-user"
                    WHERE "application"."belongs to-user" = "application.belongs to-user"."id"
                    AND ("application.belongs to-user"."actor") IS NOT NULL AND ("application.belongs to-user"."actor") = ($1)
            )
    ) AS "application"
    WHERE EXISTS (
            SELECT 1
            FROM "device" AS "application.owns-device"
            WHERE "application"."id" = "application.owns-device"."belongs to-application"
            AND ("application.owns-device"."name") IS NOT NULL AND ("application.owns-device"."name") = ($3)
    )
```
As one can see we don't apply the device access permission anymore in the select and where sub queries.
    
Change-type: minor
Depends-on: https://github.com/balena-io/pinejs/pull/335

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -11,6 +11,7 @@ import {
 	AbstractSqlType,
 	AliasNode,
 	Relationship,
+	RelationshipMapping,
 	SelectNode,
 } from '@resin/abstract-sql-compiler';
 import * as ODataParser from '@resin/odata-parser';
@@ -298,6 +299,8 @@ const namespaceRelationships = (
 		let mapping = relationship.$;
 		if (mapping != null && mapping.length === 2) {
 			mapping = _.cloneDeep(mapping);
+			// we do check the length above, but typescript thinks the second
+			// element could be undefined
 			mapping[1]![0] = `${mapping[1]![0]}$${alias}`;
 			relationships[`${key}$${alias}`] = {
 				$: mapping,
@@ -691,6 +694,202 @@ const getAlias = (name: string) => {
 	}
 	return `permissions${permissionsJSON}`;
 };
+
+const rewriteRelationship = memoizeWeak(
+	(
+		value: Relationship,
+		name: string,
+		abstractSqlModel: AbstractSqlModel,
+		permissionsLookup: PermissionLookup,
+		vocabulary: string,
+	) => {
+		let escapedName = sqlNameToODataName(name);
+		if (abstractSqlModel.tables[name]) {
+			escapedName = sqlNameToODataName(abstractSqlModel.tables[name].name);
+		}
+
+		const originalAbstractSQLModel = sbvrUtils.getAbstractSqlModel({
+			vocabulary,
+		});
+
+		const rewrite = (object: Relationship | RelationshipMapping) => {
+			if ('$' in object && Array.isArray(object.$)) {
+				// object is in the form of
+				// { "$": ["actor", ["actor", "id"]] } or { "$": ["device type"] }
+				// we are only interested in the first case, since this is a relationship
+				// to a different resource
+				const mapping = object.$;
+				if (
+					mapping.length === 2 &&
+					Array.isArray(mapping[1]) &&
+					mapping[1].length === 2 &&
+					typeof mapping[1][0] === 'string'
+				) {
+					// now have ensured that mapping looks like ["actor", ["actor", "id"]]
+					// this relations ship means that:
+					//   mapping[0] is the local field
+					//   mapping[1] is the reference to the other resource, that joins this resource
+					//   mapping[1][0] is the name of the other resource (actor in the example)
+					//   mapping[1][1] is the name of the field on the other resource
+					//
+					// this therefore defines that the local field `actor` needs
+					// to match the `id` of the `actor` resources for the join
+
+					const possibleTargetResourceName = mapping[1][0];
+
+					// Skip this if we already shortcut this connection
+					if (possibleTargetResourceName.endsWith('$bypass')) {
+						return;
+					}
+
+					const targetResourceEscaped = sqlNameToODataName(
+						abstractSqlModel.tables[possibleTargetResourceName]?.name ??
+							possibleTargetResourceName,
+					);
+
+					// This is either a translated or bypassed resource we don't
+					// mess with these
+					if (targetResourceEscaped.includes('$')) {
+						return;
+					}
+
+					let foundCanAccessLink = false;
+
+					try {
+						const odata = memoizedParseOdata(`/${targetResourceEscaped}`);
+
+						const collapsedPermissionFilters = buildODataPermission(
+							permissionsLookup,
+							methodPermissions.GET,
+							vocabulary,
+							targetResourceEscaped,
+							odata,
+						);
+
+						_.set(
+							odata,
+							['tree', 'options', '$filter'],
+							collapsedPermissionFilters,
+						);
+
+						const canAccessFunction: ResourceFunction = function(
+							property: AnyObject,
+						) {
+							// remove method property so that we won't loop back here again at this point
+							delete property.method;
+
+							if (!this.defaultResource) {
+								throw new Error(`No resource selected in AST.`);
+							}
+
+							const targetResourceAST = this.NavigateResources(
+								this.defaultResource,
+								property.name,
+							);
+
+							const targetResourceName = sqlNameToODataName(
+								targetResourceAST.resource.name,
+							);
+							const currentResourceName = sqlNameToODataName(
+								this.defaultResource.name,
+							);
+
+							if (
+								currentResourceName === targetResourceEscaped &&
+								targetResourceName === escapedName
+							) {
+								foundCanAccessLink = true;
+							}
+							// return a true expression to not select the relationship, which might be virtual
+							// this should be a boolean expression, but needs to be a subquery in case it
+							// is wrapped in an `or` or `and`
+							return ['Equals', ['Boolean', true], ['Boolean', true]];
+						};
+
+						// We need execute the abstract SQL compiler to traverse
+						// through the permissions for that resource, using a
+						// special canAccess callback.
+						const odata2AbstractSQL = new OData2AbstractSQL(
+							originalAbstractSQLModel,
+							{
+								canAccess: canAccessFunction,
+							},
+						);
+
+						try {
+							odata2AbstractSQL.match(
+								odata.tree,
+								'GET',
+								[],
+								odata.binds.length,
+							);
+						} catch (e) {
+							throw new ODataParser.SyntaxError(e);
+						}
+						if (foundCanAccessLink) {
+							// store the resource name as it was with a $bypass
+							// suffix in this relationship, this means that the
+							// query generator will use the plain resource instead
+							// of the filtered resource.
+							mapping[1][0] = `${possibleTargetResourceName}$bypass`;
+						}
+					} catch (e) {
+						if (e === constrainedPermissionError) {
+							// ignore
+							return;
+						}
+
+						// TODO: We should investigate in detail why this error
+						// occurse. It might be able to get rid of this.
+						if (e instanceof ODataParser.SyntaxError) {
+							// ignore
+							return;
+						}
+
+						throw e;
+					}
+				}
+			}
+
+			if (Array.isArray(object) || _.isObject(object)) {
+				_.forEach(object, v => {
+					// we want to recurse into the relationship path, but
+					// in case we hit a plain string, we don't need to bother
+					// checking it. This can happen since plain terms also have
+					// relationships to sbvr-types.
+					if (typeof v !== 'string') {
+						rewrite(v as Relationship | RelationshipMapping);
+					}
+				});
+			}
+		};
+
+		rewrite(value);
+	},
+);
+
+const rewriteRelationships = (
+	abstractSqlModel: AbstractSqlModel,
+	relationships: {
+		[resourceName: string]: Relationship;
+	},
+	permissionsLookup: PermissionLookup,
+	vocabulary: string,
+) => {
+	const newRelationships = _.cloneDeep(relationships);
+	_.forOwn(newRelationships, (value, name) =>
+		rewriteRelationship(
+			value,
+			name,
+			abstractSqlModel,
+			permissionsLookup,
+			vocabulary,
+		),
+	);
+
+	return newRelationships;
+};
+
 const stringifiedGetPermissions = JSON.stringify(methodPermissions.GET);
 const getBoundConstrainedMemoizer = memoizeWeak(
 	(abstractSqlModel: AbstractSqlModel) =>
@@ -720,26 +919,6 @@ const getBoundConstrainedMemoizer = memoizeWeak(
 
 				const origRelationships = Object.keys(
 					constrainedAbstractSqlModel.relationships,
-				);
-				constrainedAbstractSqlModel.relationships = new Proxy(
-					constrainedAbstractSqlModel.relationships,
-					{
-						get: (relationships, permissionResourceName: string) => {
-							if (relationships[permissionResourceName]) {
-								return relationships[permissionResourceName];
-							}
-							const alias = getAlias(permissionResourceName);
-							if (!alias) {
-								return;
-							}
-							for (const relationship of origRelationships) {
-								relationships[`${relationship}$${alias}`] =
-									relationships[relationship];
-								namespaceRelationships(relationships[relationship], alias);
-							}
-							return relationships[permissionResourceName];
-						},
-					},
 				);
 
 				_.each(constrainedAbstractSqlModel.tables, (table, resourceName) => {
@@ -802,6 +981,37 @@ const getBoundConstrainedMemoizer = memoizeWeak(
 								),
 							);
 							return permissionsTable;
+						},
+					},
+				);
+
+				// rewrite the relationships of the constraint model
+				// we check if given the current permissions we can direct
+				// expands and filters to unconstraint resources
+				constrainedAbstractSqlModel.relationships = rewriteRelationships(
+					constrainedAbstractSqlModel,
+					constrainedAbstractSqlModel.relationships,
+					permissionsLookup,
+					vocabulary,
+				);
+
+				constrainedAbstractSqlModel.relationships = new Proxy(
+					constrainedAbstractSqlModel.relationships,
+					{
+						get: (relationships, permissionResourceName: string) => {
+							if (relationships[permissionResourceName]) {
+								return relationships[permissionResourceName];
+							}
+							const alias = getAlias(permissionResourceName);
+							if (!alias) {
+								return;
+							}
+							for (const relationship of origRelationships) {
+								relationships[`${relationship}$${alias}`] =
+									relationships[relationship];
+								namespaceRelationships(relationships[relationship], alias);
+							}
+							return relationships[permissionResourceName];
 						},
 					},
 				);
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -263,6 +263,8 @@ const memoizedResolveNavigationResource = memoizeWeak(
 				`'${resourceName}' to '${navigationName}' is a field not a navigation`,
 			);
 		}
+		// we do check the length above, but typescript thinks the second
+		// element could be undefined
 		return sqlNameToODataName(abstractSqlModel.tables[mapping[1]![0]].name);
 	},
 	{ primitive: true },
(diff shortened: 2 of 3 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,161 @@ const getAlias = (name: string) => {
...
+					} catch (e) {
+						if (e === constrainedPermissionError) {
+							// ignore
+							return;
+						}
+					}
+				}
+			}
+
+			if (_.isArray(object) || _.isObject(object)) {
```
Page-:
```suggestion
			if (Array.isArray(object) || _.isObject(object)) {
```

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,161 @@ const getAlias = (name: string) => {
...
+
+						if (foundCanAccessLink) {
+							arr[1][0] = `${arr[1][0]}$bypass`;
+						}
+					} catch (e) {
+						if (e === constrainedPermissionError) {
+							// ignore
+							return;
+						}
+					}
```
Page-: This currently ignores all errors, I expect you want:
```suggestion
					} catch (e) {
						if (e === constrainedPermissionError) {
							// ignore
							return;
						}
						throw e;
					}
```

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,161 @@ const getAlias = (name: string) => {
...
+							) {
+								foundCanAccessLink = true;
+							}
+							// return a true expression to not select the relationship, which might be virtual
+							// this should be a boolean expression, but needs to be a subquery in case it
+							// is wrapped in an `or` or `and`
+							return [
+								'Exists',
+								['SelectQuery', ['Select', [['Alias', ['Number', 1], 'id']]]],
+							];
```
Page-: Can you make this an `[ Equals, Boolean, Boolean ]` please

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,161 @@ const getAlias = (name: string) => {
...
+
+		const rewrite = (object: AnyObject) => {
+			if (_.has(object, '$') && _.isArray(object['$'])) {
+				const arr = object['$'];
+				if (
+					arr.length === 2 &&
+					_.isArray(arr[1]) &&
+					arr[1].length === 2 &&
+					_.isString(arr[1][0])
+				) {
```
Page-:
```suggestion
				if (
					arr.length === 2 &&
					Array.isArray(arr[1]) &&
					arr[1].length === 2 &&
					typeof arr[1][0] === 'string'
				) {
```

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,161 @@ const getAlias = (name: string) => {
...
+		if (abstractSqlModel.tables[name]) {
+			escapedName = sqlNameToODataName(abstractSqlModel.tables[name].name);
+		}
+
+		const originalAbstractSQLModel = sbvrUtils.getAbstractSqlModel({
+			vocabulary,
+		});
+
+		const rewrite = (object: AnyObject) => {
+			if (_.has(object, '$') && _.isArray(object['$'])) {
```
Page-:
```suggestion
			if (_.has(object, '$') && Array.isArray(object['$'])) {
```

Page-: What is the reason for the `_.has` vs other checks?

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,161 @@ const getAlias = (name: string) => {
...
+
+					const targetResource = abstractSqlModel.tables[arr[1][0]];
+					let targetResourceEscaped = sqlNameToODataName(arr[1][0]);
+					if (targetResource) {
+						targetResourceEscaped = sqlNameToODataName(targetResource.name);
+					}
+
+					// This is either a transalted or bypassed resource we don't
+					// mess with these
+					if (targetResourceEscaped.indexOf('$') > 0) {
```
Page-:
```suggestion
					if (targetResourceEscaped.includes('$')) {
```

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,161 @@ const getAlias = (name: string) => {
...
+			if (_.has(object, '$') && _.isArray(object['$'])) {
+				const arr = object['$'];
+				if (
+					arr.length === 2 &&
+					_.isArray(arr[1]) &&
+					arr[1].length === 2 &&
+					_.isString(arr[1][0])
+				) {
+					// Skip this if we already shortcut this connection
+					if (arr[1][0].endsWith('$bypass')) {
```
Page-: Can you destructure/type `arr` please, right now I'm not sure what it actually is and it'd be good to have it documented via the types and var names

src/sbvr-api/permissions.ts
afitzek: I think this was an artifact from a previous version

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,182 @@ const getAlias = (name: string) => {
...
+					const targetResource =
+						abstractSqlModel.tables[possibleTargetResourceName];
+					let targetResourceEscaped = sqlNameToODataName(
+						possibleTargetResourceName,
+					);
+					if (targetResource) {
+						targetResourceEscaped = sqlNameToODataName(targetResource.name);
+					}
+
+					// This is either a transalted or bypassed resource we don't
```
Page-:
```suggestion
					// This is either a translated or bypassed resource we don't
```

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,182 @@ const getAlias = (name: string) => {
...
+					}
+
+					const targetResource =
+						abstractSqlModel.tables[possibleTargetResourceName];
+					let targetResourceEscaped = sqlNameToODataName(
+						possibleTargetResourceName,
+					);
+					if (targetResource) {
+						targetResourceEscaped = sqlNameToODataName(targetResource.name);
+					}
```
Page-:
```suggestion
					const targetResourceEscaped = sqlNameToODataName(
						targetResource?.name ?? possibleTargetResourceName,
					);
```

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,182 @@ const getAlias = (name: string) => {
...
+					if (possibleTargetResourceName.endsWith('$bypass')) {
+						return;
+					}
+
+					const targetResource =
+						abstractSqlModel.tables[possibleTargetResourceName];
+					let targetResourceEscaped = sqlNameToODataName(
+						possibleTargetResourceName,
+					);
+					if (targetResource) {
```
Page-: When would `targetResource` be missing?

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +684,182 @@ const getAlias = (name: string) => {
...
+		let escapedName = sqlNameToODataName(name);
+		if (abstractSqlModel.tables[name]) {
+			escapedName = sqlNameToODataName(abstractSqlModel.tables[name].name);
+		}
+
+		const originalAbstractSQLModel = sbvrUtils.getAbstractSqlModel({
+			vocabulary,
+		});
+
+		const rewrite = (object: AnyObject) => {
```
Page-: Ah, based on the comments I can now see this should be
```suggestion
		const rewrite = (object: Relationship | RelationshipMapping) => {
```
and then the rest of the types should naturally fall out of it

src/sbvr-api/permissions.ts
```diff
@@ -798,6 +954,39 @@ const getBoundConstrainedMemoizer = memoizeWeak(
...
+					{
+						get: (relationships, permissionResourceName: string) => {
+							if (relationships[permissionResourceName]) {
+								return relationships[permissionResourceName];
+							}
+							const alias = getAlias(permissionResourceName);
+							if (!alias) {
+								return;
+							}
+							for (const relationship of origRelationships) {
```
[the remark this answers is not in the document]
afitzek: I'm not totally sure what that code does, since I only moved that piece.

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +685,200 @@ const getAlias = (name: string) => {
...
+							// often until we have full support for translations
+							// in pine
+							return;
+						}
+
+						// we cannot throw an exception here, because this would break
+						// public APIs, by returning different error codes.
+						// We should allow the application to register an error handler
+						// to capture these exceptions.
+						console.log('Error optimizing graph permissions', e);
```
Page-: I think defaulting to throwing a 500 might make sense

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +685,200 @@ const getAlias = (name: string) => {
...
+					} catch (e) {
+						if (e === constrainedPermissionError) {
+							// ignore
+							return;
+						}
+
+						if (e instanceof ODataParser.SyntaxError) {
+							// we can ignore these, because this will occur
+							// often until we have full support for translations
+							// in pine
```
Page-: Why does it occur for translations?

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +685,200 @@ const getAlias = (name: string) => {
...
+
+						try {
+							odata2AbstractSQL.match(
+								odata.tree,
+								'GET',
+								[],
+								odata.binds.length,
+							);
+						} catch (e) {
+							throw new ODataParser.SyntaxError(e);
```
Page-: Isn't this what is thrown anyway? What other types are you seeing here that you wrap?

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +685,200 @@ const getAlias = (name: string) => {
...
+							) {
+								foundCanAccessLink = true;
+							}
+							// return a true expression to not select the relationship, which might be virtual
+							// this should be a boolean expression, but needs to be a subquery in case it
+							// is wrapped in an `or` or `and`
+							return ['Equals', ['Boolean', true], ['Boolean', true]];
+						};
+
+						// We need execute the
```
Page-: This comment looks truncated

src/sbvr-api/permissions.ts
afitzek: It will fail to find resources like `application$resin$bypass`, because they might not exist in the model.

src/sbvr-api/permissions.ts
afitzek: I need to catch this `Error` https://github.com/balena-io-modules/odata-to-abstract-sql/blob/master/src/odata-to-abstract-sql.ts#L1374

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +685,201 @@ const getAlias = (name: string) => {
...
+		}
+
+		const originalAbstractSQLModel = sbvrUtils.getAbstractSqlModel({
+			vocabulary,
+		});
+
+		const rewrite = (object: Relationship | RelationshipMapping) => {
+			if ('$' in object && Array.isArray(object.$)) {
+				// object is in the form of
+				// { "$": ["actor", ["actor", "id"]] } or { "$": ["device type"] }
```
Page-: Can it ever be like `{ "$": ["device type"] }` ? The types say it cannot be but they could be wrong (and should be updated if so)

src/sbvr-api/permissions.ts
Page-: Can you give some examples of resources that are failing to resolve? It'd be interesting to see which ones to understand why

src/sbvr-api/permissions.ts
afitzek: Yes one example would be: 
```
{
 "device": {
  "has": {
    ...
    "uuid": { "$": ["uuid"] },
    ...
  }
  ...
 }
}
```

src/sbvr-api/permissions.ts
afitzek: One resource that is failing there is `application$resin$bypass`, it happens soon when you run the api tests.

src/sbvr-api/permissions.ts
Page-: Ok, lets update the types then to allow for this version

afitzek: done: https://github.com/balena-io-modules/abstract-sql-compiler/pull/83/files

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +685,201 @@ const getAlias = (name: string) => {
...
+					arr[1].length === 2 &&
+					typeof arr[1][0] === 'string'
+				) {
+					// now have ensured that arr looks like ["actor", ["actor", "id"]]
+					// this relations ship means that:
+					//   arr[0] is the local field
+					//   arr[1] is the reference to the other resource, that joins this resource
+					//   arr[1][0] is the name of the other resource (actor in the example)
+					//   arr[1][1] is the name of the field on the other resource
+					//
```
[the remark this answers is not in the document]
afitzek: done: https://github.com/balena-io-modules/abstract-sql-compiler/pull/83/files

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +685,201 @@ const getAlias = (name: string) => {
...
+					// Skip this if we already shortcut this connection
+					if (possibleTargetResourceName.endsWith('$bypass')) {
+						return;
+					}
+
+					const targetResource =
+						abstractSqlModel.tables[possibleTargetResourceName];
+					const targetResourceEscaped = sqlNameToODataName(
+						targetResource?.name ?? possibleTargetResourceName,
+					);
```
Page-: I don't think `targetResource` is used anywhere else right?
```suggestion
					const targetResourceEscaped = sqlNameToODataName(
						abstractSqlModel.tables[possibleTargetResourceName]?.name ?? possibleTargetResourceName,
					);
```

src/sbvr-api/permissions.ts
Page-: Approved

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +685,201 @@ const getAlias = (name: string) => {
...
+		abstractSqlModel: AbstractSqlModel,
+		permissionsLookup: PermissionLookup,
+		vocabulary: string,
+	) => {
+		let escapedName = sqlNameToODataName(name);
+		if (abstractSqlModel.tables[name]) {
+			escapedName = sqlNameToODataName(abstractSqlModel.tables[name].name);
+		}
+
+		const originalAbstractSQLModel = sbvrUtils.getAbstractSqlModel({
```
Page-: Does this `originalAbstractSQLModel` have any `$bypass` info?

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +685,201 @@ const getAlias = (name: string) => {
...
+					}
+				});
+			}
+		};
+
+		rewrite(value);
+	},
+);
+
+const rewriteRelationsships = (
```
Page-:
```suggestion
const rewriteRelationships = (
```

src/sbvr-api/permissions.ts
```diff
@@ -684,6 +685,201 @@ const getAlias = (name: string) => {
...
+
+						// We need execute the abstract SQL compiler to traverse
+						// through the permissions for that resource, using a
+						// special canAccess callback.
+						const odata2AbstractSQL = new OData2AbstractSQL(
+							originalAbstractSQLModel,
+							{
+								canAccess: canAccessFunction,
+							},
+						);
```
Page-: I think creating a new OData2AbstractSQL instance is actually fairly expensive due to have to do some preprocessing on the model (for shortening aliases). This is all done once at startup right and never at runtime? if so we should be fine but we might need to look into optimizing this somehow in future to keep startup times reasonable

src/sbvr-api/permissions.ts
afitzek: No

src/sbvr-api/permissions.ts
afitzek: This is a good point. This is executed also at runtime until the caches are hot. These should be cached per permission set. This should probably be optimized. I think this could be tricky, because we do want the scope of the current function in the canAccess function.

Page-: Yeah I'm thinking we'll just need to be able to hotswap the canAccess function before we run the `.match` and since it's all synchronous we can know that's safe to do although it is a bit ugly

src/sbvr-api/sbvr-utils.ts
```diff
@@ -263,7 +263,9 @@ const memoizedResolveNavigationResource = memoizeWeak(
 				`'${resourceName}' to '${navigationName}' is a field not a navigation`,
 			);
 		}
-		return sqlNameToODataName(abstractSqlModel.tables[mapping[1][0]].name);
+		// we do check the length above, but typescript thinks the second
+		// element could be undefined
+		return sqlNameToODataName(abstractSqlModel.tables[mapping[1]![0]].name);
```
Page-: Does using `if (mapping[1] == null)` above work?

PR #352, opened 2020-03-26
🔒: Only allow one credential to provide permissions for the request

afitzek:
Change-type: major


afitzek:
@balena-ci retest

PR #361, opened 2020-05-05
Update dependencies

Page-:
Update abstract-sql-compiler from 6.13.2 to 6.13.4
Update lf-to-abstract-sql from 3.2.2 to 3.2.3
Update odata-to-abstract-sql from 4.4.1 to 4.4.5
Update sbvr-parser from 0.2.7 to 0.2.8
Update sbvr-types from 2.0.5 to 2.0.7

Change-type: patch


src/sbvr-api/hooks.ts
```diff
@@ -54,26 +54,20 @@ export const rollbackRequestHooks = <T extends InstantiatedHooks<any>>(
...
-			if (hook instanceof SideEffectHook) {
-				return hook.rollback();
-			}
-		},
-	);
+	settleMapSeries(_(hooks).flatMap().compact().value(), (hook) => {
+		if (hook instanceof SideEffectHook) {
+			return hook.rollback();
+		}
+	});
```
dfunckt, line 61: Wow, thank you Prettier, at last. Its previous formatting was so annoying.

PR #363, opened 2020-05-07
Make use of `import type`

Page-:
Change-type: patch

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -1,12 +1,4 @@
-import {
-	Definition,
-	OData2AbstractSQL,
-	odataNameToSqlName,
-	ResourceFunction,
-	sqlNameToODataName,
-} from '@resin/odata-to-abstract-sql';
-
-import {
+import type {
 	AbstractSqlModel,
 	AbstractSqlType,
 	AliasNode,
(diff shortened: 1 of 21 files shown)


Page-:
@balena-ci retest

src/sbvr-api/permissions.ts
```diff
@@ -1,25 +1,29 @@
...
-import {
+import type {
 	AbstractSqlModel,
 	AbstractSqlType,
 	AliasNode,
 	Relationship,
 	RelationshipMapping,
 	SelectNode,
 } from '@resin/abstract-sql-compiler';
+import type * as express from 'express';
```
dfunckt, line 9:
```suggestion
import type * as express from 'Express';
```

like elsewhere.

src/sbvr-api/sbvr-utils.ts
```diff
@@ -57,6 +59,7 @@ import {
 	statusCodeToError,
 	TranslationError,
 	UnauthorizedError,
+	AnyObject,
```
dfunckt, line 62: Doesn't need to happen with with PR, but feels like `AnyObject` should be in `common_types.ts`, definitely not in `errors.ts`.

src/sbvr-api/sbvr-utils.ts
```diff
@@ -73,20 +76,17 @@ import * as controlFlow from './control-flow';
...
 import { version as LF2AbstractSQLVersion } from '@resin/lf-to-abstract-sql/package.json';
 import { version as sbvrTypesVersion } from '@resin/lf-to-abstract-sql/package.json';
-import { Model } from '../config-loader/config-loader';
 import {
 	compileRequest,
 	getAndCheckBindValues,
 	isRuleAffected,
 } from './abstract-sql';
-import { OptionalField, RequiredField } from './common-types';
-import { AnyObject } from './sbvr-utils';
```
dfunckt, line 89: Ugh :/

PR #364, opened 2020-05-08
Make wider use of async/await

Page-:
Change-type: patch

### DIFF
--- a/src/sbvr-api/control-flow.ts
+++ b/src/sbvr-api/control-flow.ts
@@ -17,12 +17,23 @@ export const liftP = <T, U>(fn: (v: T) => Resolvable<U>) => {
 export type MappingFunction = <T, U>(
 	a: T[],
 	fn: (v: T) => Resolvable<U>,
-) => Bluebird<Array<U | Error>>;
+) => Promise<Array<U | Error>>;
 
 // The settle version of `Promise.mapSeries`
-export const settleMapSeries: MappingFunction = (a, fn) => {
-	const runF = Bluebird.method(fn);
-	return Bluebird.mapSeries(a, _.flow(runF, wrap));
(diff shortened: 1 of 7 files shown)


src/sbvr-api/control-flow.ts
```diff
@@ -17,12 +17,23 @@ export const liftP = <T, U>(fn: (v: T) => Resolvable<U>) => {
...
+) => Promise<Array<U | Error>>;
 
 // The settle version of `Promise.mapSeries`
-export const settleMapSeries: MappingFunction = (a, fn) => {
-	const runF = Bluebird.method(fn);
-	return Bluebird.mapSeries(a, _.flow(runF, wrap));
+export const settleMapSeries: MappingFunction = async <T, U>(
+	a: T[],
+	fn: (v: T) => Resolvable<U>,
+) => {
```
dfunckt, line 26: How come you have to declare types both for the var and the arguments? Why isn't `MappingFunction` enough for TS to infer the argument types?

src/server-glue/server.ts
```diff
@@ -84,13 +84,12 @@ if (!process.browser) {
 }
 
 export const initialised = Pinejs.init(app)
-	.then((configLoader) =>
-		Bluebird.all([
+	.then(async (configLoader) => {
+		await Bluebird.all([
```
dfunckt, line 88:
```suggestion
		await Promise.all([
```

src/sbvr-api/control-flow.ts
Page-: It's enough for it to infer the types but we won't have access to `U` for the `const results: Array<U | Error> = [];` below

Page-:
@balena-ci retest

Page-:
@balena-ci retest

PR #365, opened 2020-05-11
Update to @balena/sbvr-parser 1.x

Page-:
Update lf-to-abstract-sql from 3.2.3 to 3.2.6
Update odata-to-abstract-sql from 4.4.5 to 4.4.6
Update sbvr-parser from 0.2.8 to 1.0.1

Change-type: patch


Page-:
@balena-ci retest

PR #368, opened 2020-05-12
Update to @balena/lf-to-abstract-sql 4.x

Page-:
Depends on #367 

Update lf-to-abstract-sql from 3.2.3 to 4.0.0

Change-type: patch


Page-:
@balena-ci retest

PR #371, opened 2020-05-22
13.x

Page-:
* Update to @balena/abstract-sql-compiler 7.x
* Update to @balena/sbvr-types 3.x
* Update to @balena/odata-to-abstract-sql 5.x
* Update to pinejs-client-core 6.x
* Rename to @balena/pinejs
* Switch to returning native promises
* Stop exporting `AnyObject` in errors
* Stop exporting permissions as part of sbvrUtils
* Stop exporting errors as part of sbvrUtils 
* Update to pg 8.x 
* Update @types/node to 12.x


Page-:
@balena-ci retest

Page-:
@balena-ci retest

Page-:
@balena-ci retest

src/sbvr-api/odata-response.ts
```diff
@@ -1,7 +1,7 @@
-declare module '@resin/abstract-sql-compiler' {
+declare module '@balena/abstract-sql-compiler' {
 	interface AbstractSqlTable {
 		fetchProcessingFields?: {
-			[field: string]: (field: any) => Bluebird<any>;
+			[field: string]: NonNullable<typeof sbvrTypes[string]['fetchProcessing']>;
```
thgreasi, line 4: I might be wrong, but I think that this isn't `NonNullable`
See: https://github.com/balena-io-modules/sbvr-types/blob/07981e2ed3fe40d638175dd97de71e3d1474e16e/src/types/date.ts#L12-L17

src/pinejs-session-store/pinejs-session-store.ts
```diff
@@ -30,25 +30,45 @@ const sessionModel = `
 		Necessity: Each session has at most 1 expiry time
 `;
 
+const asCallback = async <T>(
+	callback: undefined | ((err: any, result?: T) => void),
+	promise: Promise<T>,
```
thgreasi, line 35: I would prefer having the promise as the first arg, imo follows the `fact => result` pattern and I'm used always seeing callbacks last, but not a bit deal :)

src/sbvr-api/odata-response.ts
Page-: It is in this case because this property is built up as
```
const getFetchProcessingFields = (table: AbstractSqlTable) => {
	if (table.fetchProcessingFields == null) {
		table.fetchProcessingFields = _(table.fields)
			.filter(
				({ dataType }) =>
					sbvrTypes[dataType] != null &&
					sbvrTypes[dataType].fetchProcessing != null,
			)
			.map(({ fieldName, dataType }) => {
				const odataName = sqlNameToODataName(fieldName);
				return [odataName, sbvrTypes[dataType].fetchProcessing];
			})
			.fromPairs()
			.value();
	}
	return table.fetchProcessingFields!;
};
```
so it filters out any that are null

thgreasi: Scratch that, I thought I saw a `ReturnType` in there...

PR #372, opened 2020-05-22
Convert javascript files to async/await

Page-:
Change-type: patch

### DIFF
--- a/src/data-server/sbvr-server.js
+++ b/src/data-server/sbvr-server.js
@@ -46,14 +46,14 @@ const isServerOnAir = (() => {
 })();
 
 /** @type { import('express').Handler } */
-const serverIsOnAir = (_req, _res, next) =>
-	isServerOnAir().then((onAir) => {
-		if (onAir) {
-			next();
-		} else {
-			next('route');
-		}
-	});
+const serverIsOnAir = async (_req, _res, next) => {
+	const onAir = await isServerOnAir();
+	if (onAir) {
+		next();
+	} else {
+		next('route');
+	}
+};
 
 export let config = {
 	models: [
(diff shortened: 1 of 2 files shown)


src/data-server/sbvr-server.js
```diff
@@ -76,39 +76,40 @@ ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT N
...
-							text: ' ',
 						},
-					});
-				}
-			})
-			.then(() =>
-				devApi.get({
+					},
+				})
+				.then((/** @type { Array<{ [key: string]: any }> } */ result) => {
```
thgreasi, line 99: Missed that `.then` on purpose?
I guess it's the type that made it harder.

src/data-server/sbvr-server.js
```diff
@@ -121,26 +122,26 @@ export function setup(app, sbvrUtils, db) {
...
+						apiRoot: instance.is_of__vocabulary,
+						modelText: instance.model_value,
+					});
 				});
-			})
-			.then(() => isServerOnAir(true))
-			.catch(() => isServerOnAir(false));
+			await isServerOnAir(true);
+		} catch {
+			isServerOnAir(false);
```
thgreasi, line 138: Don't we typically also need an await here?

src/data-server/sbvr-server.js
Page-: Yeah, I can't force the typing if I convert it to async/await (it requires specific brackets that prettier would remove), so it can stay this way until the file is in typescript

src/data-server/sbvr-server.js
Page-: Strictly speaking the `await` isn't needed for either when passing a value, but I'll add it

PR #373, opened 2020-05-22
More async/await conversion

Page-:
Change-type: patch

### DIFF
--- a/src/express-emulator/express.js
+++ b/src/express-emulator/express.js
@@ -63,7 +63,7 @@ const app = (function () {
 			body = '';
 		}
 		if (!handlers[method]) {
-			return Bluebird.reject(404);
+			return Bluebird.reject([404, null, null]);
 		}
 		const req = {
 			// Have a default user for in-browser with all permissions
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -1047,18 +1047,19 @@ const getCheckPasswordQuery = _.once(() =>
 		},
 	}),
 );
-export const checkPassword = (
-	username: string,
-	password: string,
-): Bluebird<{
-	id: number;
-	actor: number;
-	username: string;
-	permissions: string[];
-}> =>
-	getCheckPasswordQuery()({
-		username,
-	}).then(async ([user]: AnyObject[]) => {
+export const checkPassword = Bluebird.method(
+	async (
+		username: string,
+		password: string,
+	): Promise<{
+		id: number;
+		actor: number;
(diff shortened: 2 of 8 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -1151,23 +1153,25 @@ const getUserPermissionsQuery = _.once(() =>
...
-			permissions.map((permission) => permission.name),
-		)
-		.tapCatch((err) => {
+export const getUserPermissions = Bluebird.method(
+	async (userId: number): Promise<string[]> => {
+		if (typeof userId === 'string') {
+			userId = parseInt(userId, 10);
+		}
+		if (!Number.isFinite(userId)) {
+			throw new Error('User ID has to be numeric, got: ' + typeof userId);
```
dfunckt, line 1162: Nit:

```suggestion
			throw new Error(`User ID has to be numeric, got: ${typeof userId}`);
```

Page-: done

src/express-emulator/express.js
```diff
@@ -63,7 +63,7 @@ const app = (function () {
 			body = '';
 		}
 		if (!handlers[method]) {
-			return Bluebird.reject(404);
+			return Bluebird.reject([404, null, null]);
```
Page-, line 66: This now matches up with the rejection value on all the other branches of this promise

src/sbvr-api/permissions.ts
```diff
@@ -1336,31 +1340,34 @@ const checkApiKey = Bluebird.method(
...
 
 			const parts = auth.split(' ');
 			if (parts.length !== 2) {
+				next?.();
 				return;
 			}
 
 			const [scheme, apiKey] = parts;
 			if (scheme.toLowerCase() !== expectedScheme) {
+				next?.();
```
dfunckt, line 1363: All these `next?.()` calls feel fragile -- I think this one specific case is not worth changing away from `try/then`.

src/server-glue/module.ts
```diff
@@ -44,13 +44,13 @@ export const init = (
 	app: Express.Application,
 	config?: string | configLoader.Config,
 ): Bluebird<ReturnType<typeof configLoader.setup>> =>
-	sbvrUtils
-		.setup(app, db)
-		.then(async () => {
+	Bluebird.try(async () => {
```
Page-, line 47: I had to use `Bluebird.try` instead of `Bluebird.method` because `Bluebird.method` types don't work properly with optional params

PR #375, opened 2020-06-17
Refactor exports

Page-:


src/sbvr-api/errors.ts
```diff
@@ -8,10 +10,7 @@ export class SbvrValidationError extends TypedError {}
 
 export class TranslationError extends TypedError {}
 export class ParsingError extends TypedError {}
-
-export interface AnyObject {
-	[index: string]: any;
-}
+export { AnyObject };
```
dfunckt, line 13: I think it's best to not export this one, getting it from common-types everywhere is better IMO.

src/sbvr-api/sbvr-utils.ts
```diff
@@ -59,7 +59,6 @@ import {
 	statusCodeToError,
 	TranslationError,
 	UnauthorizedError,
-	AnyObject,
```
dfunckt, line 62: Github won't let me comment there, but it seems we're re-exporting `./permissions` on line 46 above. Can that be removed?

src/sbvr-api/errors.ts
Page-: This is necessary for backwards compatibility, I'll drop this in the major

src/sbvr-api/sbvr-utils.ts
Page-: It's also a major change, but I do plan to

PR #377, opened 2020-06-19
Optimize rewriting relationships for the canAccess optimization

Page-:
Depends on https://github.com/balena-io-modules/odata-to-abstract-sql/pull/63/files


Page-:
@balena-ci retest

PR #379, opened 2020-07-01
Fetch affected ids only once per request

thgreasi:
Change-type: minor
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/CJ0OxldLkgX3R0raiP-Zht4yk-a

### DIFF
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -1024,7 +1024,30 @@ const getIdField = (
 	>,
 ) => getAbstractSqlModel(request).tables[resolveSynonym(request)].idField;
 
-export const getAffectedIds = async ({
+export const getAffectedIds = async (args: {
+	req: HookReq;
+	request: HookRequest;
+	tx: Db.Tx;
+}): Promise<number[]> => {
+	const { request } = args;
+	if (request.affectedIds) {
+		return request.affectedIds;
+	}
+
+	// We keep the affected ids promise so we only have to fetch them once per request
+	if (request.pendingAffectedIds != null) {
+		return request.pendingAffectedIds;
+	}
+	request.pendingAffectedIds = $getAffectedIds(args);
+
+	// Keep the affected ids and let the promise to be GCed sooner.
+	request.affectedIds = await request.pendingAffectedIds;
+	delete request.pendingAffectedIds;
+
+	return request.affectedIds;
+};
+
+const $getAffectedIds = async ({
 	req,
 	request,
 	tx,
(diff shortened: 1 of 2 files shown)


src/sbvr-api/sbvr-utils.ts
```diff
@@ -1024,7 +1024,24 @@ const getIdField = (
...
+	const { request } = args;
+	if (request.affectedIds) {
+		return request.affectedIds;
+	}
+	// We store the affected ids promise so we only have to fetch it once per request
+	if (request._affectedIdsPromise == null) {
+		request._affectedIdsPromise = getAffectedIdsNoCache(args);
+	}
+
+	return request._affectedIdsPromise;
```
Page-, line 1041: Can you make this
```suggestion
	if (request.pendingAffectedIds == null) {
		request.pendingAffectedIds = $getAffectedIds(args);
	}

	await request.pendingAffectedIds;
	delete request.pendingAffectedIds;
	
	return request.affectedIds
```
Few things
1. using `pending` for the pending promise as it should be a bit clearer
2. deleting the promise once it's resolved so it can be GC'ed immediately
3. using `$getAffectedIds` for the internal version, mainly because it matches how we name memoized functions generally

thgreasi: I thought about point 2 as well, but since it wasn't already in the api I thought there was a reason for this.
Will do :+1:

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1024,7 +1024,27 @@ const getIdField = (
...
+	const { request } = args;
+	if (request.affectedIds) {
+		return request.affectedIds;
+	}
+	// We store the affected ids promise so we only have to fetch them once per request
+	if (request.pendingAffectedIds == null) {
+		request.pendingAffectedIds = $getAffectedIds(args);
+	}
+
+	const affectedIds = await request.pendingAffectedIds;
```
thgreasi, line 1041: Allows us to avoid having a `!` in the `return affectedIds;`, which we would have to do if we did `return request.affectedIds;`.
Feels cleaner as well imo :)

thgreasi: Actually @Page- , while running the api tests I realized that the `request.affectedIds` on `master` is assigned on the re-assigned `request` object (`request = await uriParser.parseOData(/*...*/)`.)
Is that correct?
See: https://github.com/balena-io/pinejs/blob/5c62bbda68129809635e8a040ab81c6c992df71c/src/sbvr-api/sbvr-utils.ts#L1045-L1048

thgreasi:
Updated @Page- and moved the `request.affectedIds` assignment to the new `getAffectedIds()`. Feels cleaner and I also think it fixes an issue in the previous code.

PR #382, opened 2020-07-21
Update pinejs-client-core and use named keys when querying

Page-:
Depends on https://github.com/balena-io-modules/pinejs-client-js/pull/102

Update pinejs-client-core from 6.0.0 to 6.1.0

Change-type: minor


src/config-loader/config-loader.ts
```diff
@@ -54,13 +54,13 @@ const getOrCreate = async (
 	uniqueFields: AnyObject,
 	extraFields?: AnyObject,
```
thgreasi, line 55: Kind reminder that there is also a pinej-client PR about adding a generic `getOrCreate()` method, but I guess we can make it simpler after this.
See: https://github.com/balena-io-modules/pinejs-client-js/pull/100

Page-:
@balena-ci I self-certify!

(LGTM'd above already and I've updated to point at the released pinejs-client-core version)

Page-:
@balena-ci I self-certify!

PR #384, opened 2020-07-29
Improve compatibility with async/await stack traces

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

thgreasi:
Right while I was reviewing this :P

PR #386, opened 2020-08-11
Update pinejs-client-core to add support for $count sugar

Page-:
Update pinejs-client-core from 6.1.0 to 6.5.0
Update typed-error from 3.2.0 to 3.2.1

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #387, opened 2020-08-11
Update pinejs-client-core and make use of improved typings

Page-:
Update pinejs-client-core from 6.5.0 to 6.6.0

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #391, opened 2020-08-17
Separate running the odata request from the expressjs handler

Page-:
This simplifies the implementation of internal api calls via `runURI`
and will allow async stack traces across internal calls with a future
major bump

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #394, opened 2020-08-18
Add the ability to omit metadata from the response

Page-:
Update odata-parser from 2.1.0 to 2.2.0
Update pinejs-client-core from 6.7.0 to 6.7.1

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #398, opened 2020-08-19
Separate the web sql wrapper for async support from the WebSqlTx class

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #400, opened 2020-08-19
Add the ability to mark an existing tx as read-only

Page-:
This is only handled locally by requiring any statements to be
read-only but it's useful for websql and also as a level of
protection against accidental writes

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #401, opened 2020-08-19
Add support for marking hooks as read-only

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #404, opened 2020-08-20
Add the ability to set sideEffects/readOnlyTx for a set of hooks

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #406, opened 2020-08-20
Update typescript to 4.x

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #407, opened 2020-08-20
Use readTransaction for read-only requests

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #411, opened 2020-08-24
Make use of optional assignment syntax

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #416, opened 2020-10-02
Update dependencies

Page-:
Update abstract-sql-compiler from 7.0.2 to 7.0.3
Update lf-to-abstract-sql from 4.0.0 to 4.1.0
Update pinejs-client-core from 6.7.3 to 6.9.1

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #417, opened 2020-10-13
Update dependencies, adding support for negative constants in sbvr

Page-:
Update sbvr-parser from 1.0.1 to 1.1.1

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #418, opened 2020-10-13
Update README now that sbvr-compiler/odata-compiler are included

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #419, opened 2020-10-15
Condense starting a read only transaction to a single statement

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #420, opened 2020-10-22
Fix logger creation for nodejs 14

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #421, opened 2020-10-26
Leave the migrationsPath intact when loading a model

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #422, opened 2020-12-02
Update dependencies

Page-:
Update abstract-sql-compiler from 7.0.3 to 7.0.6
Update lf-to-abstract-sql from 4.1.0 to 4.1.1
Update odata-parser from 2.2.0 to 2.2.1
Update odata-to-abstract-sql from 5.3.0 to 5.3.1
Update sbvr-types from 3.1.0 to 3.1.2
Update pinejs-client-core from 6.9.1 to 6.9.3

Change-type: patch

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,8 +1,0 @@
-{
-	"*.js": [
-		"balena-lint --typescript --fix"
-	],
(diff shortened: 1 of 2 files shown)


Page-:
@balena-ci I self-certify!

PR #423, opened 2020-12-02
Improve compatibility with webpack 5

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #424, opened 2020-12-09
Add support for differentiating check constraint errors

Page-:
Change-type: minor

### DIFF
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -290,6 +290,8 @@ const prettifyConstraintError = (
 			);
 		}
 
+		// Override the error message so we don't leak any internal db info
+		err.message = 'Constraint failed';
 		throw err;
 	}
 };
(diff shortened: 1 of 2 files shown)


src/sbvr-api/sbvr-utils.ts
```diff
@@ -289,8 +289,6 @@ const prettifyConstraintError = (
 				'Data is referenced by ' + sqlNameToODataName(matches[1]) + '.',
 			);
 		}
-
-		throw err;
```
richbayliss, line 293: did you mean to drop this `throw` ?

richbayliss (commented):
Looks good, just a question which may be simple.

src/sbvr-api/sbvr-utils.ts
Page-: Yes in order to avoid leaking info in the case it's a new constraint error that hasn't got special handling, but I've changed it to instead override the error which would be better than having it go to the default `Database error` override message

richbayliss (approved):

PR #425, opened 2020-12-14
Add support for prettifying check constraints with their description

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #428, opened 2020-12-18
Update dependencies

Page-:
Update abstract-sql-compiler from 7.2.0 to 7.4.1
Update odata-to-abstract-sql from 5.3.1 to 5.4.0

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #435, opened 2021-01-05
Disable the creation of a package-lock on each npm install

thgreasi:
Just like we do in pine-client. Atm a no-op since we do not have a pakcage-lock in the repo at all.

Change-type: patch

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false


thgreasi:
@balena-ci retest

thgreasi:
@balena-ci I self-certify!

PR #439, opened 2021-01-15
Update dependencies

Page-:
Update abstract-sql-compiler from 7.8.1 to 7.8.2
Update abstract-sql-to-typescript from 1.1.0 to 1.1.1

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #440, opened 2021-01-18
Optimize generating constrained abstract sql model

Page-:
Instantiating an `OData2AbstractSQL` instance can be relatively costly
and as such reusing a memoized instance can provide a reasonable
performance improvement

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #441, opened 2021-01-20
Add a timeout for pg rollbacks after which the connection is terminated

Page-:
This avoids issues where a postgres connection can hang and the
rollback never finishes and the client is never released from the pool

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #442, opened 2021-01-20
Error/dequeue all in-progress queries when rolling back a transaction

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #445, opened 2021-01-22
Only connect to the database on init and allow specifiy db options to the init call

Page-:


Page-:
@balena-ci I self-certify!

PR #448, opened 2021-01-25
Use 500s for any database errors that are not explicitly handled

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #449, opened 2021-01-25
Use 400s for data type validation errors

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #450, opened 2021-01-25
Update odata-parser to 2.2.2 for optimized number parsing

Page-:
Update odata-parser from 2.2.1 to 2.2.2

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #453, opened 2021-02-01
Update sbvr-types to 3.2.0, adding support for more `Text (Type)` verbs

Page-:
Update abstract-sql-compiler from 7.10.0 to 7.10.1
Update sbvr-types from 3.1.3 to 3.2.0

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #457, opened 2021-02-08
Add support for sbvr date comparison and allow referencing current time

Page-:
Update lf-to-abstract-sql from 4.1.1 to 4.2.0
Update sbvr-parser from 1.1.1 to 1.2.2
Update sbvr-types from 3.2.0 to 3.4.0

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #459, opened 2021-02-15
Gracefully handle `HttpError` thrown from PREPARSE hooks

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #460, opened 2021-02-26
Warn when setting a `request.values` property that will not take effect

Page-:
Change-type: patch

### DIFF
--- a/src/sbvr-api/uri-parser.ts
+++ b/src/sbvr-api/uri-parser.ts
@@ -403,6 +403,17 @@ export const translateUri = <
 	if (!isMetadataEndpoint) {
 		const abstractSqlQuery = memoizedOdata2AbstractSQL(request);
 		request = { ...request };
+		request.values = new Proxy(request.values, {
+			set: (obj: ODataRequest['values'], prop: string, value) => {
+				if (!obj.hasOwnProperty(prop)) {
+					sbvrUtils.api[request.vocabulary].logger.warn(
+						`Assigning a new request.values property '${prop}' however it will be ignored`,
+					);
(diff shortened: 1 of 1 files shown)


src/sbvr-api/uri-parser.ts
```diff
@@ -403,6 +403,17 @@ export const translateUri = <
 	if (!isMetadataEndpoint) {
 		const abstractSqlQuery = memoizedOdata2AbstractSQL(request);
 		request = { ...request };
+		request.values = new Proxy(request.values, {
+			set: (obj: ODataRequest['values'], prop: string, value) => {
+				if (!obj.hasOwnProperty(prop)) {
```
thgreasi, line 408: Does this make the log appear even if I'm setting prop "x" to a dummy value in the POSTPARSE and then set it again to a different value in the PRERUN?

Page-: Nope, because the property will exist already by the time this code is hit

thgreasi: But that we should ideally log in that case as well right? I mean in my test the value changed in the PRERUN didn't persist.
Couldn't we instead of `if (!obj.hasOwnProperty(prop)) {` just be somehow checking which HOOK lifecycle stage we are at?

Page-: This essentially does that since we add the proxy immediately after we've finalized the acceptable `request.values` keys, any additions before this will work and any after this point will be noops

thgreasi (approved):
Thanks for chasing and explaining this so fast :heart:

PR #462, opened 2021-03-01
Update abstract-sql-compiler bring more rules optimizations

Page-:
Update abstract-sql-compiler from 7.10.2 7.11.0
Update lf-to-abstract-sql from 4.2.0 to 4.2.1

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #463, opened 2021-03-08
Update abstract-sql-compiler bring more rules optimizations

Page-:
Update abstract-sql-compiler from 7.11.0 to 7.12.0

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #464, opened 2021-03-12
Block eagerly fetching affected ids in cases it is not possible to do

Page-:
This fixes an issue where trying to eagerly fetch affected ids for
equests that can create entries would return ids for existing entries
that you could have created rather than giving the ids for the entries
that you will create

Change-type: patch

### DIFF
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -891,7 +891,7 @@ export const getAffectedIds = async (
 	}
 	request.pendingAffectedIds = $getAffectedIds(args);
 
-	// Keep the affected ids and let the promise to be GCed sooner.
+	// Store the affected ids and delete the pending promise to allow it to be GCed sooner.
 	request.affectedIds = await request.pendingAffectedIds;
 	delete request.pendingAffectedIds;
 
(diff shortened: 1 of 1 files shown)


src/sbvr-api/sbvr-utils.ts
```diff
@@ -905,9 +905,13 @@ const $getAffectedIds = async ({
 }: HookArgs & {
 	tx: Db.Tx;
 }): Promise<number[]> => {
-	if (request.method === 'GET') {
-		// GET requests don't affect anything so passing one to this method is a mistake
-		throw new Error('Cannot call `getAffectedIds` with a GET request');
+	if (['PATCH', 'DELETE'].includes(request.method)) {
+		// We can only find the affected ids in advance for requests that modify existing requests, if they
```
dfunckt, line 909:
```suggestion
		// We can only find the affected ids in advance for requests that modify existing records/resources/rows?, if they
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -905,9 +905,13 @@ const $getAffectedIds = async ({
 }: HookArgs & {
 	tx: Db.Tx;
 }): Promise<number[]> => {
-	if (request.method === 'GET') {
-		// GET requests don't affect anything so passing one to this method is a mistake
-		throw new Error('Cannot call `getAffectedIds` with a GET request');
+	if (['PATCH', 'DELETE'].includes(request.method)) {
```
dfunckt, line 908:
```suggestion
	if (!['PATCH', 'DELETE'].includes(request.method)) {
```

dfunckt (changes requested):

dfunckt (approved):

PR #468, opened 2021-03-16
Only use the api key if we successfully resolved one

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #469, opened 2021-03-22
Enable strict options by default and only disable specific ones

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #470, opened 2021-04-29
Fix typo in odata-compiler help

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #471, opened 2021-04-29
Update request to query by affected ids once they've been resolved

Page-:
This avoids the need to use a potentially costly filter twice, once to
get the affected ids in advance and then once when running the actual
PATCH/DELETE. This also guarantees there can never be an inconsistency
between the two

Change-type: minor


Page-:
@balena-ci retest

PR #473, opened 2021-05-17
Update sbvr-types to 3.4.3 to support boolean computed terms w/o casting

thgreasi:
Update sbvr-types from 3.4.1 to 3.4.3

Change-type: minor


thgreasi:
@balena-ci I self-certify!

PR #474, opened 2021-05-18
Update abstract-sql-compiler to 7.13.0 to support CurrentTimestamp/Date

thgreasi:
Update abstract-sql-compiler from 7.12.0 to 7.13.0

Change-type: minor


thgreasi:
@balena-ci I self-certify!

PR #476, opened 2021-06-18
Updating to add an expiry date to 'api key'

nitishagar:
Change-Type: minor
Spec: https://jel.ly.fish/e450713f-001c-4470-b571-71724aeb0869

### DIFF
--- a/src/sbvr-api/user.sbvr
+++ b/src/sbvr-api/user.sbvr
@@ -53,6 +53,9 @@ Fact type:  api key has key
 	Necessity: each key is of exactly one api key
 Fact type:  api key has role
 	Note: An 'api key' will inherit all the 'permissions' that the 'role' has.
+	Term Form: api key role
+	Fact type: api key role has expiry date
+		Necessity: Each api key role has at most one expiry date.
 Fact type:  api key has permission
 Fact type:  api key is of actor
 	Necessity: each api key is of exactly one actor
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -1283,6 +1283,16 @@ const $getApiKeyPermissions = (() => {
 																			},
 																		},
 																	},
+																	$or: [
+																		{
+																			khr: { expiry_date: null },
+																		},
+																		{
+																			khr: {
+																				expiry_date: { $gt: { $now: null } },
+																			},
+																		},
+																	],
 																},
 															},
 														},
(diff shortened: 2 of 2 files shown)


nitishagar:
@Page- I am making a change to `pinejs` for the first time. Didn't find anything specific to test this out in the readme.

Any pointers for any additional change required in addition with addition to `.sbvr` here?

dfunckt:
@nitishagar you'll need a migration and possibly enforce expiry when supplied when checking permissions. See this similar commit: https://github.com/balena-io/pinejs/commit/22402c1d4b68e119362e54c23b4a5d4c7312cc5e

Page- (changes requested):
As @dfunckt says you'll need the migrations, and also I'd like to see this added with the code that actually enforces the expiry date

nitishagar:
Thanks @Page- @dfunckt. I have in form replicated query for role and its expiry date.

Two questions: 
- The file name for migration is the convention to base it on the next release version?
- Why do have these deep nested queries right in the `.ts` file?

dfunckt (commented):
👍

src/sbvr-api/user.sbvr
```diff
@@ -53,6 +53,9 @@ Fact type:  api key has key
 	Necessity: each key is of exactly one api key
 Fact type:  api key has role
 	Note: An 'api key' will inherit all the 'permissions' that the 'role' has.
+	Term Form: api key role
+	Fact type: api key role has expiry date
+		Necessity: Each user role has at most one expiry date.
```
fisehara, line 58: The rule: `Necessity: Each user role has at most one expiry date. ` already exists:

```
Term: user
...
Fact type: user has role
	Term Form: user role
	Fact type: user role has expiry date
		Necessity: Each user role has at most one expiry date.
```	
```suggestion
		Necessity: Each api key role has at most one expiry date.
```

fisehara: 👍

migrations/14.24.2-add-api-key-expiry-date.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "api key-has-role" ADD COLUMN "expiry date" TIMESTAMP NULL;
```
Page-, line 1: This should be a migration attached to the user model, ie in https://github.com/balena-io/pinejs/blob/v14.35.1/src/sbvr-api/permissions.ts#L1624-L1653

nitishagar: So this would be some like below, after [L1653](https://github.com/balena-io/pinejs/blob/v14.35.1/src/sbvr-api/permissions.ts#L1624-L1653):
```
'14.24.2-modified-at': `
     ALTER TABLE "api key-has-role" ADD COLUMN "expiry date" TIMESTAMP NULL;
```

Page-: Yes

nitishagar:
Implementing the key expiry on key instead of roles. JF for details: https://jel.ly.fish/e450713f-001c-4470-b571-71724aeb0869

PR #477, opened 2021-06-30
Delete CODEOWNERS

thgreasi:
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/pub/threads/trLcZFnSX9fLZn4LiaIv4xuBbah


thgreasi:
@balena-ci I self-certify!

PR #481, opened 2021-07-06
Update user SBVR model URL

myarmolinsky:
Update user SBVR model URL

Change-type: patch
Signed-off-by: Matthew Yarmolinsky <matthew-timothy@balena.io>


myarmolinsky:
@balena-ci I self certify!

PR #482, opened 2021-07-06
Fix Hooks URL

myarmolinsky:
Fix Hooks URL

Change-type: patch
Signed-off-by: Matthew Yarmolinsky <matthew-timothy@balena.io>


myarmolinsky:
@balena-ci I self certify!

PR #486, opened 2021-07-09
Update dependencies

Page-:
Update odata-parser from 2.2.5 to 2.2.6

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #487, opened 2021-07-13
Reduce duplication for http error constructors

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #489, opened 2021-07-13
Standardize http error to response conversion and response sending

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #490, opened 2021-07-15
Add quotes for single entity curl requests

inspaya:
Add quotes for single entity curl requests

`curl` requests were failing when accessing endpoints with ids e.g. `curl -X GET http://localhost:1337/example/device(3)`. Quoting such endpoints resolved the issue e.g. `curl -X GET "http://localhost:1337/example/device(3)"`

Change-type: patch
Signed-off-by: Muizudeen Kusimo <muizz@balena.io>

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -208,7 +208,7 @@ $ curl -X GET http://localhost:1337/example/device
 The server will respond with an array containing all the devices, if we want to access a specific one, it is sufficient to add the id at the end of the URL we pass.
 
 ```
-$ curl -X GET "http://localhost:1337/example/device(1)"
+$ curl -X GET 'http://localhost:1337/example/device(1)'
 ```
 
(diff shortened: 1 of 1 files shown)


docs/GettingStarted.md
```diff
@@ -228,7 +228,7 @@ What went wrong here? Pine.js is simply preventing us from violating the constra
 To correctly modify the device we can try:
 
 ```
-$ curl -X PUT -d name=testdevice -d note=updatednote -d type=raspberry http://localhost:1337/example/device(1)
+$ curl -X PUT -d name=testdevice -d note=updatednote -d type=raspberry "http://localhost:1337/example/device(1)"
```
Page-, line 231: In general I'd use single-quotes when possible because it also helps rule out issues with eg `$filter` being evaluated as an env var but for these cases it doesn't matter

Page- (approved):

inspaya:
@balena-ci rebase

PR #494, opened 2021-07-19
Export a `handleHttpErrors` function to handle HttpError/subclasses

Page-:
Change-type: minor


Page-:
@balena-ci I self-certify!

PR #495, opened 2021-07-23
Add method for getting transaction level advisory locks in Postgresql

thgreasi:
Change-type: minor

### DIFF
--- a/src/database-layer/db.ts
+++ b/src/database-layer/db.ts
@@ -2,7 +2,7 @@
 import type * as Mysql from 'mysql';
 import type * as Pg from 'pg';
 import type * as PgConnectionString from 'pg-connection-string';
-import type { Resolvable } from '../sbvr-api/common-types';
+import type { Dictionary, Resolvable } from '../sbvr-api/common-types';
 
 import { Engines } from '@balena/abstract-sql-compiler';
 import * as Bluebird from 'bluebird';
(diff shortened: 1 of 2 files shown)


src/database-layer/db.ts
```diff
@@ -529,6 +535,20 @@ if (maybePg != null) {
 				}
 			}
 
+			public override async getTxLevelLock(key1: number, key2: number) {
```
Page-, line 538: I'd prefer to see the first key be pre-registered, eg how open-balena-api uses `ADVISORY_LOCK_NAMESPACES`, you'd pre-register your namespaces and then just provide the namespace key for the first argument

thgreasi: Should that be done by a `registerLockNamespace` method exported by `database-layer/db` or would you imagine it be better if it's added as an extra property in `init`'s `config` object (in `server-glue/module`).
At first a `registerLockNamespace()` that each feature can register it's own namespaces sounded like more appropriate for the feature based separation that we have started following, but on second thought, we will still need coordination in order for two features to not end up with a namespace id clash.

src/database-layer/db.ts
```diff
@@ -108,6 +108,32 @@ export const engines: {
 	[engine: string]: (connectString: string | object) => Database;
 } = {};
 
+const transactionLockNamespaceMap: Dictionary<number> = {};
+
+export function registerTransactionLockNamespace(
+	namespaceKey: string,
+	namespaceId: number,
```
Page-, line 115: This is an integer type in postgres so at least for that needs to be within -2147483648 to +2147483647 (https://www.postgresql.org/docs/13/datatype-numeric.html), I think for now just checking 0-2147483647 would be fine - we can always expand it later as need be if we for some reason want more

src/database-layer/db.ts
```diff
@@ -108,6 +108,32 @@ export const engines: {
...
+		throw new TypeError(`Invalid namespaceId parameter provided for Tx lock`);
+	}
+	if (transactionLockNamespaceMap[namespaceKey] != null) {
+		throw new Error(
+			`Error while registering transaction lock namespace '${namespaceKey}'. Namespace key is already registered.`,
+		);
+	}
+	const existingNamespaceEntry = Object.entries(
+		transactionLockNamespaceMap,
+	).find(([_key, id]) => id === namespaceId);
```
Page-, line 127:
```suggestion
	).find(([, id]) => id === namespaceId);
```

Page- (approved):

thgreasi:
@balena-ci rebase

thgreasi:
@balena-ci rebase

PR #498, opened 2021-09-13
Avoid iterating the sql result if affectedIds is already populated

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #500, opened 2021-09-29
Avoid creating a new readOnlyArgs object when args is already read-only

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #501, opened 2021-09-29
Add support for a `PINEJS_DEBUG` env var that is more strictly checked

Page-:
This means we can warn on/ignore invalid values rather than giving a
potentially surprising result, eg `'0'` being truthy

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #504, opened 2021-10-05
Add the postgres pool error handler even when no PG_SCHEMA is specified

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #506, opened 2021-10-14
Allow passing a function as the cache params to enable custom caching

Page-:
This allows using a different cache than the default of memoizee

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #509, opened 2021-11-15
Use the env pool settings as defaults only if not explicitly provided

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #512, opened 2021-11-30
Add support for circular `canAccess` permissions using a depth argument

Page-:
eg using `resource/canAccess(2)` will allow recursing twice before
bailing on that branch and considering it falsey

Update odata-to-abstract-sql from 5.4.4 to 5.4.6

Change-type: minor


src/sbvr-api/permissions.ts
```diff
@@ -526,31 +527,64 @@ const generateConstrainedAbstractSql = (
...
+			if (type !== 'Real' || !Number.isInteger(depth) || depth < 1) {
+				throw new Error('The depth for `canAccess` must be an integer >= 1');
+			}
+			if (
+				// Make sure we have enough traces yet to have exceeded the depth before checking if they match
+				canAccessTrace.length > depth &&
+				canAccessTrace
+					.slice(-depth)
+					// Check if we've reached the specified depth with all entries being the same resource, and if so we need to terminate the recursion by returning a falsey expression
+					.every((rName) => rName === targetResourceName)
```
thgreasi, line 579: Just curious: what's an example that we need all of them to match?
I was expecting something more like `canAccessTrace[canAccessTrace.length - depth] === targetResourceName`.

thgreasi (commented):
Well, this was more of a learning experience than a review. LGTM :+1:

src/sbvr-api/permissions.ts
Page-: If you were to use `A?A/canAccess(2)` and had `A -> B -> A` then it wouldn't have gone to that depth in a direct circle. That said we probably still want to block indirect circles altogether so I'll have to handle that case differently

src/sbvr-api/permissions.ts
```diff
@@ -526,31 +527,74 @@ const generateConstrainedAbstractSql = (
...
+		if (traceIndex !== -1) {
+			if (canAccessTrace[canAccessTrace.length - 1] !== targetResourceName) {
+				throw new Error(
+					`Indirectly circular 'canAccess()' permissions are not supported, currently permissions for ${resourceName} form an indirect circle by the following path: ${canAccessTrace.join(
+						' -> ',
+					)} -> ${targetResourceName}`,
+				);
+			}
+
+			const { args } = method[1];
```
fisehara, line 570: api.5f57cbfa6e9525866b9883a4d88c2d47.bob.local

fisehara, line 570: how do you know that entry at idx 1 exists?

Page-: Because the type is `[ 'call', { args: any[], method: string } ]` and is checked before we ever enter this function

fisehara: thx. helps me understand the code better 👍

Page-:
@balena-ci I self-certify!

thgreasi:
:tada:

PR #517, opened 2021-12-16
Update dependencies for improved sbvr negation

Page-:
Update abstract-sql-compiler from 7.15.0 to 7.17.0
Update lf-to-abstract-sql from 4.4.0 to 4.4.1
Update sbvr-parser from 1.2.5 to 1.3.0

Change-type: minor


Page-:
@balena-ci I self-certify!

PR #519, opened 2022-01-11
Update dependencies

Page-:
Update @balena/odata-to-abstract-sql from 5.4.6 to 5.4.7

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #520, opened 2022-01-12
Async runtime migrator

fisehara:
Add async migrator

Needed for long lasting / massive data affecting migrations for zero downtime of API.

Depends-on: https://github.com/balena-io/pinejs/pull/556

Change-type: minor


docs/Migrations.md
```diff
@@ -46,4 +46,103 @@ Migrations which are Node modules should export a function which will be called
...
 
-You should return a promise so pinejs can wait for completion/errors.
\ No newline at end of file
+You should return a promise so pinejs can wait for completion/errors.
+
+
+## Async Migrations
+
+Migrations that may lock tables for a long time for example column renames on large tables can be run as async migrations.
+Async migrations are executed in small batches within single transactions on the data and will run in parallel to pine (server). Async migrations are multi-instance save. They are synchronized between instances via a database table `migration status` and the minimum interval between two migration executions is guaranteed across instances.
```
dfunckt, line 55:
```suggestion
Migrations that may lock tables for a long time, eg. column renames on large tables, can be run as async migrations.

Async migrations are executed in small batches within individual transactions and run concurrently while Pine is serving requests. Async migrations are multi-instance safe. They are synchronized between instances via the `migration status` database table and the minimum interval between two migration executions is guaranteed across instances.
```

docs/Migrations.md
```diff
@@ -46,4 +46,103 @@ Migrations which are Node modules should export a function which will be called
...
+When multiple async migrations are specified in one deployment they are all executed competitive but not in parallel: That means, that one migration query is executed at a time but different migration queries can run after each other. Hence, migrations that depend on the result of a previous migration should not be described
```
dfunckt, line 56:
```suggestion

When Pine starts and determines there are multiple async migrations to be run, it starts all migrations concurrently. These migrations however do not run in parallel: only one migration batch is executed at any given time, then another, and so on. Thus, migrations should not depend on the result a previous one.
```

docs/Migrations.md
```diff
@@ -46,4 +46,103 @@ Migrations which are Node modules should export a function which will be called
...
+
+
+## Async Migrations
+
+Migrations that may lock tables for a long time for example column renames on large tables can be run as async migrations.
+Async migrations are executed in small batches within single transactions on the data and will run in parallel to pine (server). Async migrations are multi-instance save. They are synchronized between instances via a database table `migration status` and the minimum interval between two migration executions is guaranteed across instances.
+When multiple async migrations are specified in one deployment they are all executed competitive but not in parallel: That means, that one migration query is executed at a time but different migration queries can run after each other. Hence, migrations that depend on the result of a previous migration should not be described 
+
+Async migrations run forever until the next deployment to guarantee that future data matching the migration condition are also migrated. The status of running migrations can be checked in `migration status` table. The configured execution parameters and the execution metrics are stored and updated after each migration iteration.
```
dfunckt, line 58:
```suggestion
Once an async migration starts, it keeps running forever until the server is shutdown. This is to guarantee that new data inserted or updated at runtime can also be migrated. The status of running migrations can be checked in `migration status` table. The configured execution parameters and the execution metrics are stored and updated after each migration batch completes.
```

docs/Migrations.md
```diff
@@ -46,4 +46,103 @@ Migrations which are Node modules should export a function which will be called
...
+
+## Async Migrations
+
+Migrations that may lock tables for a long time for example column renames on large tables can be run as async migrations.
+Async migrations are executed in small batches within single transactions on the data and will run in parallel to pine (server). Async migrations are multi-instance save. They are synchronized between instances via a database table `migration status` and the minimum interval between two migration executions is guaranteed across instances.
+When multiple async migrations are specified in one deployment they are all executed competitive but not in parallel: That means, that one migration query is executed at a time but different migration queries can run after each other. Hence, migrations that depend on the result of a previous migration should not be described 
+
+Async migrations run forever until the next deployment to guarantee that future data matching the migration condition are also migrated. The status of running migrations can be checked in `migration status` table. The configured execution parameters and the execution metrics are stored and updated after each migration iteration. 
+
+
```
dfunckt, line 60:
```suggestion
```

docs/Migrations.md
```diff
@@ -46,4 +46,103 @@ Migrations which are Node modules should export a function which will be called
...
+Async migrations are stored in the migrations folder specified in the pine configuration / model (same as the synchron / init migrations). The file names
+need to have an infix of type .async. and can be pure SQL query files or .ts / .js files executing a database SQL query.
+The async migration query has to have a LIMIT statement to limit the maximum count of 'locked' rows per transaction and migration step.
```
dfunckt, line 63:
```suggestion
Async migrations are stored in the migrations folder, alongside synchronous migrations. Their file names are significant and must contain the string `.async.` so they are treated as async migrations. They can be pure SQL (with `.sql` extension) or Typescript (with `.ts`/`.js` extensions) files.

The async migration query must have a `LIMIT` statement to limit the maximum number of affected rows per batch.
```

docs/Migrations.md
```diff
@@ -46,4 +46,103 @@ Migrations which are Node modules should export a function which will be called
...
+		return await tx.executeSql(staticSql);
+	},
+	delayMS: 100,
+	backoffDelayMS: 4000,
+	errorThreshold: 15,
+};
+```
+### SQL query file (plain text)
+
+When providing a SQL file only the default values for interval, backoff and error threshold are:
```
dfunckt, line 112:
```suggestion
The default values for interval, backoff and error threshold are: 
```

src/config-loader/config-loader.ts
```diff
@@ -251,18 +255,42 @@ export const setup = (app: Express.Application) => {
...
+									migrationCategory = fileNameParts[1];
+								}
+								/**
+								 *  helper to assign migrations with category level to model
+								 *  example migration file names:
+								 *
+								 *  key0-name.ts  				==> defaults startup migration
+								 *  key1-name1.sql 				==> defaults startup migration
+								 *  key2-name2.async.ts			==> async migraiton (async datafiller)
+								 *  key3-name3.async.sql		==> async migraiton (async datafiller)
```
dfunckt, line 273:
```suggestion
								 *  key2-name2.async.ts			==> async migration (async datafiller)
								 *  key3-name3.async.sql		==> async migration (async datafiller)
```

src/config-loader/config-loader.ts
```diff
@@ -282,14 +310,36 @@ export const setup = (app: Express.Application) => {
...
 			);
 			await loadConfig(configObj);
+			await loadAsyncMigrations(configObj);
 		} catch (err) {
 			console.error('Error loading application config', err, err.stack);
 			process.exit(1);
 		}
 	};
 
+	const loadAsyncMigrations = async (data: Config): Promise<any> => {
```
dfunckt, line 320: This isn't just loading the migrations -- it executes them as well.

```suggestion
	const runAsyncMigrations = async (data: Config): Promise<any> => {
```

src/config-loader/config-loader.ts
```diff
@@ -282,14 +310,36 @@ export const setup = (app: Express.Application) => {
...
+			return Promise.resolve('Async migrations executed');
+		} catch (err) {
+			return Promise.reject(`Error loading async migrations: \n ${err}`);
+		}
+	};
+
 	return {
 		loadConfig,
 		loadApplicationConfig,
+		loadAsyncMigrations,
```
dfunckt, line 343: Why do you export this? Is it used elsewhere apart from above in `loadApplicationConfig`?

src/config-loader/config-loader.ts
```diff
@@ -282,14 +310,36 @@ export const setup = (app: Express.Application) => {
...
+		try {
+			await Promise.all(
+				data.models.map(async (model) => {
+					if (
+						model.migrations != null &&
+						Object.keys(model.migrations).includes(migrationCategories.async)
+					) {
+						await sbvrUtils.executeAsyncModelMigration(
+							model as sbvrUtils.ExecutableModel,
+						);
```
dfunckt, line 330: If migrations run until shutdown, will this promise ever resolve?

src/config-loader/config-loader.ts
fisehara: Thx!

src/config-loader/config-loader.ts
fisehara: No its not exported or used externally. Thx.

src/config-loader/config-loader.ts
fisehara: Thx!

src/config-loader/config-loader.ts
fisehara: The promise resolved as intervals were setup inside of it. Still it confuses and I've taken out the async await of this call and the (for await) loop inside the async migrator. All async migrations must not interfer with other async migrations executed in one deployment, thus the setup of it is save to be not async/await.
Thx!

docs/Migrations.md
fisehara: Thx!

docs/Migrations.md
fisehara: Thx!

docs/Migrations.md
fisehara: Thx

fisehara:
@balena-ci retest

fisehara:
@balena-ci retest

src/migrator/async.ts
```diff
@@ -0,0 +1,279 @@
...
+					try {
+						// clear the interval to avoid retriggering before finisheing this migraiton query
+						clearInterval(asyncScheduler);
+
+						// sync on the last execution time between instances
+						// precondition: All running instances are running on the same time/block
+						// skip execution
+						if (migrationState.last_run_time) {
+							const durationSinceLastRun =
+								Date.now().valueOf() - migrationState.last_run_time.valueOf();
```
Page-, line 200:
```suggestion
								Date.now() - migrationState.last_run_time.getTime();
```

src/migrator/async.ts
```diff
@@ -0,0 +1,279 @@
...
+								(!migrationState.is_backoff &&
+									durationSinceLastRun < migrationState.delayMS)
+							) {
+								// will still execute finally block where the migration lock is released.
+								return;
+							}
+						}
+						// set last run time and run counter only when backoff time sync between
+						// competing instances is in sync
+						migrationState.last_run_time = new Date(Date.now());
```
Page-, line 213:
```suggestion
						migrationState.last_run_time = new Date();
```

src/migrator/async.ts
```diff
@@ -0,0 +1,279 @@
...
+			error_threshold: migratorEnv.asyncMigrationDefaultErrorThreshold,
+			delayMS: migratorEnv.asyncMigrationDefaultDelayMS,
+			backoffDelayMS: migratorEnv.asyncMigrationDefaultBackoffDelayMS,
+			converged_time: undefined,
+			last_error_message: undefined,
+			is_backoff: false,
+			should_stop: false,
+		};
+
+		let asyncRunnerMigratorFn: ((tx: Tx) => Promise<number>) | undefined;
```
Page-, line 137:
```suggestion
		let asyncRunnerMigratorFn: ((tx: Tx) => Promise<number>);
```

src/migrator/utils.ts
```diff
@@ -23,12 +24,49 @@ export type CategorizedMigrations = {
...
+	delayMS?: number | undefined;
+	backoffDelayMS?: number | undefined;
+	errorThreshold?: number | undefined;
+};
+export function isAsyncMigration(
+	migration: AsyncMigration | RunnableMigrations,
+): migration is AsyncMigration {
+	return (
+		(migration as AsyncMigration).type !== undefined &&
+		(migration as AsyncMigration).type === 'async'
```
Page-, line 47: The undefined check is redundant:
```suggestion
		(migration as AsyncMigration).type === 'async'
```

src/migrator/utils.ts
```diff
@@ -158,3 +196,106 @@ WHERE "migration"."model name" = ${1}`,
...
+				migrationStatus['start_time'],
+				migrationStatus['delayMS'],
+				migrationStatus['backoffDelayMS'],
+				migrationStatus['is_backoff'] === true ? 1 : 0,
+				migrationStatus['error_threshold'],
+				migrationStatus['should_stop'] === true ? 1 : 0,
+			],
+		);
+		return result;
+	} catch (err) {
```
Page-, line 222:
```suggestion
	} catch {
```

src/migrator/utils.ts
```diff
@@ -158,3 +196,106 @@ WHERE "migration"."model name" = ${1}`,
...
+SELECT *
+FROM "migration status"
+WHERE "migration status"."migration key" = ${1}
+LIMIT 1;`,
+			[migrationKey],
+		);
+
+		const data = rows[0];
+		if (data == null) {
+			return undefined;
```
Page-, line 279:
```suggestion
			return;
```

src/migrator/utils.ts
```diff
@@ -158,3 +196,106 @@ WHERE "migration"."model name" = ${1}`,
...
+			delayMS: data['delayMS'],
+			backoffDelayMS: data['backoffDelayMS'],
+			converged_time: data['converged_time'],
+			last_error_message: data['last error message'],
+			is_backoff: data['is backoff'] === 1 ? true : false,
+			should_stop: data['should stop'] === 1 ? true : false,
+		};
+	} catch (err) {
+		// we ignore all errors to not spam the api logs
+		return undefined;
```
Page-, line 299:
```suggestion
	} catch {
		// we ignore all errors to not spam the api logs
		return;
```

src/migrator/utils.ts
fisehara: Thx!

src/migrator/utils.ts
fisehara: thx, done.

src/migrator/utils.ts
fisehara: thx, done.

src/migrator/utils.ts
fisehara: thx, done.

src/migrator/async.ts
fisehara: thx

src/migrator/async.ts
```diff
@@ -0,0 +1,279 @@
...
+	) {
+		return;
+	}
+	let runMigrations: RunnableMigrations = {};
+	for (const [key, value] of Object.entries(migrations)) {
+		if (key === MigrationCategories.async && value instanceof Object) {
+			runMigrations = { ...runMigrations, ...value };
+		}
+	}
+	await $run(model, runMigrations);
```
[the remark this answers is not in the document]
fisehara: thx, followed the suggesiton

src/migrator/async.ts
fisehara: thx, followed the suggestion

src/migrator/utils.ts
fisehara: thx, followed the suggestion

Page- (commented):
I still have more review to do but wanted to post what I have thus far

tempalte.migration.async.ts
```diff
@@ -0,0 +1,47 @@
+import { AsyncMigration } from '../../../../../src/migrator/utils';
```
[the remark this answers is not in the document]
fisehara: Oh, this file shouldn't exists at all :(
Thx.

src/migrator/async.ts
```diff
@@ -0,0 +1,283 @@
...
+							tx,
+							{
+								batchSize,
+							},
+							sbvrUtils,
+						)
+					).rowsAffected;
+			} else if (migration.asyncSql && typeof migration.asyncSql === 'string') {
+				const asyncMigraitonSqlStatement = migration.asyncSql?.replace(
+					'%%ASYNC_BATCH_SIZE',
```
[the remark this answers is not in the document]
fisehara: Thx, good point!

src/migrator/async.ts
```diff
@@ -0,0 +1,283 @@
...
+						await (migration.asyncFn as AsyncMigrationFn)(
+							tx,
+							{
+								batchSize,
+							},
+							sbvrUtils,
+						)
+					).rowsAffected;
+			} else if (migration.asyncSql && typeof migration.asyncSql === 'string') {
+				const asyncMigraitonSqlStatement = migration.asyncSql?.replace(
```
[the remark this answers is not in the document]
fisehara: thx

src/migrator/async.ts
```diff
@@ -0,0 +1,283 @@
...
+};
+
+const $run = async (
+	setupTx: Tx,
+	model: ApiRootModel,
+	migrations: RunnableAsyncMigrations,
+): Promise<void> => {
+	const modelName = model.apiRoot;
+
+	const asyncMigrationSetup: any[] = [];
```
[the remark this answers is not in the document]
fisehara: Yes, good catch.

src/sbvr-api/sbvr-utils.ts
```diff
@@ -465,6 +466,37 @@ export const generateModels = (
 	return { vocab, se, lf, abstractSql, sql, odataMetadata };
 };
 
+export const checkModelAlreadyExists = async (
```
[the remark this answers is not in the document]
fisehara: Yes, my rebase apparently went wrong!

src/config-loader/config-loader.ts
```diff
@@ -302,6 +329,14 @@ export const setup = (app: Express.Application) => {
 										);
 										break;
 									case '.sql':
+										if (migrationCategory === 'async') {
+											console.error(
+												`Plain async migration sql statement not supported, needs to be object, skipping: ${path.extname(
```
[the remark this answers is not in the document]
fisehara: Thx

src/migrator/utils.ts
```diff
@@ -23,12 +24,118 @@ export type CategorizedMigrations = {
...
+	asyncFn?: AsyncMigrationFn | undefined;
+	syncFn?: MigrationFn | undefined;
+	asyncSql?: string | undefined;
+	syncSql?: string | undefined;
+	delayMS?: number | undefined;
+	backoffDelayMS?: number | undefined;
+	errorThreshold?: number | undefined;
+	asyncBatchSize?: number | undefined;
+	finalize?: boolean | undefined;
+};
```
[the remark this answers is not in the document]
fisehara: Thanks, that is great! Will follow the suggestion

src/migrator/utils.ts
```diff
@@ -23,12 +24,118 @@ export type CategorizedMigrations = {
...
+export function isAsyncMigration(
+	migration: AsyncMigration | RunnableMigrations,
+): migration is AsyncMigration {
+	return (migration as AsyncMigration).type === 'async';
```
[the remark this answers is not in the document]
fisehara: thx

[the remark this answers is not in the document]
fisehara: thx

src/migrator/utils.ts
```diff
@@ -152,3 +259,100 @@ export const migrationTablesExist = async (tx: Tx) => {
...
+				migrationStatus['migration_key'],
+				migrationStatus['start_time'],
+				migrationStatus['is_backoff'] === true ? 1 : 0,
+				migrationStatus['should_stop'] === true ? 1 : 0,
+			],
+		);
+		return result;
+	} catch (err: any) {
+		// it already exists and we can ignore this error.
+		err = err;
```
Page-, line 283:
```suggestion
```

src/migrator/utils.ts
```diff
@@ -23,12 +24,118 @@ export type CategorizedMigrations = {
...
+		Object.entries(migrations).map(([category, categroyMigrations]) => {
+			if (category in MigrationCategories) {
+				Object.entries(categroyMigrations as Migrations).map(
+					([key, migration]) => {
+						if (isAsyncMigration(migration)) {
+							if (migration.finalize) {
+								if (migration.syncFn) {
+									runnableMigrations = {
+										...runnableMigrations,
+										...{ [key]: migration.syncFn },
```
Page-, line 103:
```suggestion
										[key]: migration.syncFn,
```

src/migrator/utils.ts
```diff
@@ -23,12 +24,118 @@ export type CategorizedMigrations = {
...
+									};
+								} else if (migration.syncSql) {
+									runnableMigrations = {
+										...runnableMigrations,
+										...{ [key]: migration.syncSql },
```
Page-, line 108:
```suggestion
										[key]: migration.syncSql,
```

src/migrator/utils.ts
```diff
@@ -23,12 +24,118 @@ export type CategorizedMigrations = {
...
+									};
+								}
+							}
+						} else {
+							runnableMigrations = {
+								...runnableMigrations,
+								...{ [key]: migration },
```
Page-, line 115:
```suggestion
								[key]: migration,
```

src/migrator/utils.ts
```diff
@@ -23,12 +24,118 @@ export type CategorizedMigrations = {
...
+	migrations: Migrations,
+): RunnableAsyncMigrations => {
+	let runnableMigrations: RunnableAsyncMigrations = {};
+	for (const [key, migration] of Object.entries(
+		migrations[MigrationCategories.async],
+	)) {
+		if (isAsyncMigration(migration)) {
+			runnableMigrations = {
+				...runnableMigrations,
+				...{ [key]: migration },
```
Page-, line 80:
```suggestion
				[key]: migration,
```

src/migrator/utils.ts
```diff
@@ -152,3 +259,100 @@ export const migrationTablesExist = async (tx: Tx) => {
...
+			migration_key: data['migration key'],
+			start_time: data['start time'],
+			last_run_time: data['last run time'],
+			run_counter: data['run counter'],
+			migrated_rows: data['migrated rows'],
+			error_counter: data['error counter'],
+			converged_time: data['converged time'],
+			last_error_message: data['last error message'],
+			is_backoff: data['is backoff'] === 1 ? true : false,
+			should_stop: data['should stop'] === 1 ? true : false,
```
Page-, line 352:
```suggestion
			is_backoff: data['is backoff'] === 1,
			should_stop: data['should stop'] === 1,
```

src/migrator/utils.ts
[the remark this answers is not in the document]
fisehara: @Page- return 'type' in migration fails for the string migration definitions:


```
> const test = 'testest'
> 'key' in test
Uncaught TypeError: Cannot use 'in' operator to search for 'key' in testest
> const test1 = {}
> 'key' in test1
false
```
Going with 
```
(migration as AsyncMigration).type === MigrationCategories.async
```

src/migrator/utils.ts
fisehara: ok

src/migrator/utils.ts
fisehara: thx!

src/migrator/utils.ts
fisehara: thx!

src/migrator/utils.ts
fisehara: thx!

src/migrator/utils.ts
fisehara: thx!

src/migrator/utils.ts
fisehara: thx!

src/migrator/async.ts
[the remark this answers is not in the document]
fisehara: Going for:
```
	const asyncMigrationSetup: Array<{
		key: string;
		initMigrationState: MigrationStatus & {
			error_threshold: number;
			delayMS: number;
			backoffDelayMS: number;
		};
		asyncRunnerMigratorFn: (tx: Tx) => Promise<number>;
	}> = [];
```

src/migrator/utils.ts
```diff
@@ -23,12 +24,137 @@ export type CategorizedMigrations = {
...
+									runnableMigrations = {
+										...runnableMigrations,
+										[key]: migration.syncFn,
+									};
+								} else if (migration.syncSql) {
+									runnableMigrations = {
+										...runnableMigrations,
+										[key]: migration.syncSql,
+									};
+								}
```
Page-, line 129: You should be able to switch this to
```suggestion
								runnableMigrations = {
									...runnableMigrations,
									[key]: migration.syncFn ?? migration.syncSql,
								};
```

src/migrator/utils.ts
```diff
@@ -152,3 +278,95 @@ export const migrationTablesExist = async (tx: Tx) => {
...
+VALUES (${1}, ${2}, ${3});
+`,
+			[
+				migrationStatus['migration_key'],
+				migrationStatus['start_time'],
+				migrationStatus['is_backoff'] ? 1 : 0,
+			],
+		);
+		return result;
+	} catch (err: any) {
```
Page-, line 299:
```suggestion
	} catch {
```

src/migrator/utils.ts
```diff
@@ -152,3 +278,95 @@ export const migrationTablesExist = async (tx: Tx) => {
...
+			start_time: data['start time'],
+			last_run_time: data['last run time'],
+			run_counter: data['run counter'],
+			migrated_rows: data['migrated rows'],
+			error_counter: data['error counter'],
+			converged_time: data['converged time'],
+			last_error_message: data['last error message'],
+			is_backoff: data['is backoff'] === 1,
+		};
+	} catch (err) {
```
Page-, line 368:
```suggestion
	} catch {
```

src/migrator/utils.ts
```diff
@@ -152,3 +278,95 @@ export const migrationTablesExist = async (tx: Tx) => {
...
+		// we ignore all errors to not spam the api logs
```
Page-, line 369: What sort of errors are you seeing?

src/migrator/async.ts
```diff
@@ -0,0 +1,301 @@
...
+							// competing instances is in sync
+							migrationState.last_run_time = new Date();
+							migrationState.run_counter += 1;
+
+							let migratedRows = 0;
+							// here a separate transaction is needed as this migration may fail
+							// when it fails it would break the transaction for managing the migration status
+							await sbvrUtils.db.transaction(async (migrationTx) => {
+								migratedRows =
+									(await asyncRunnerMigratorFn?.(migrationTx)) || 0;
```
Page-, line 248: This should presumably be
```suggestion
									(await asyncRunnerMigratorFn?.(migrationTx)) ?? 0;
```
?

src/migrator/async.ts
```diff
@@ -0,0 +1,301 @@
...
+							});
+
+							migrationState.migrated_rows += migratedRows;
+							if (migratedRows === 0) {
+								// when all rows have been catched up once we only catch up less frequently
+								migrationState.is_backoff = true;
+								if (!migrationState.converged_time) {
+									// only store the first time when migrator converged to all data migrated
+									migrationState.converged_time = new Date();
+								}
```
Page-, line 258: Should this be
```suggestion
								// only store the first time when migrator converged to all data migrated
								migrationState.converged_time ??= new Date();
```
?

src/migrator/async.ts
```diff
@@ -0,0 +1,301 @@
...
+								migrationState.error_counter %
+									initMigrationState.error_threshold ===
+									0
+							) {
+								migrationState.last_error_message = `${err.name} ${err.message}`;
+								(sbvrUtils.api.migrations?.logger.error ?? console.error)(
+									`${key}: ${err.name} ${err.message}`,
+								);
+								migrationState.is_backoff = true;
+							}
```
Page-, line 277: We probably still want to log/do something if `err` is not an `instance of Error` because that'll be pretty surprising/bad

src/migrator/utils.ts
```diff
@@ -23,12 +24,137 @@ export type CategorizedMigrations = {
...
+export function isAsyncMigration(
+	migration: AsyncMigration | RunnableMigrations,
+): migration is AsyncMigration {
+	return (migration as AsyncMigration).type === MigrationCategories.async;
+}
+
+export function areCategorizedMigrations(
+	migrations: Migrations,
+): migrations is CategorizedMigrations {
+	return Object.keys(migrations).some((key) => key in MigrationCategories);
```
Page-, line 71: I think this should be
```suggestion
	return Object.keys(migrations).every((key) => key in MigrationCategories);
```
?
Or maybe something like
```suggestion
	const keys = Object.keys(migrations);
	const containsCategories = keys.some((key) => key in MigrationCategories);
	if (containsCategories && !keys.every((key) => key in MigrationCategories)) {
		throw new Error('Mixing categorized and uncategorized migrations is not supported');
	}
	return containsCategories;
```
to ensure that people don't mix/match migrations (which will presumably be an accident and break things?)

src/migrator/utils.ts
fisehara: I was thinking about the possibility that the transaction is closed b/c the async migrator has faced any other issue. But as this is called only after the transactions has been created, I don't see any real issue. I followed the pattern here, that an async migrator should not fail the running instance but just silentlty fail.

src/migrator/async.ts
fisehara: yes

src/migrator/async.ts
fisehara: yes, thanks

src/migrator/async.ts
fisehara: Yes, that's right. Thanks!
Adding logging for key (migration key) : {err}

src/migrator/utils.ts
fisehara: I'm going for the mixing category protection. Thanks

src/migrator/utils.ts
fisehara: thx

src/migrator/utils.ts
fisehara: thx

src/migrator/utils.ts
fisehara: oh thx a lot!

src/migrator/utils.ts
```diff
@@ -152,3 +278,95 @@ export const migrationTablesExist = async (tx: Tx) => {
...
+				migrationStatus['last_run_time'],
+				migrationStatus['migrated_rows'],
+				migrationStatus['error_counter'],
+				migrationStatus['converged_time'],
+				migrationStatus['last_error_message'],
+				migrationStatus['is_backoff'] ? 1 : 0,
+				migrationStatus['migration_key'],
+			],
+		);
+	} catch (err) {
```
Page-, line 333:
```suggestion
	} catch {
```

src/migrator/utils.ts
Page-: Yeah I agree it shouldn't fail the instance but it's the silencing that I'm questioning really - if something is failing in this section then it probably needs to be dealt with (at least I wouldn't expect any errors here as standard)

fisehara: Thus, a logging information would be fine, right?

src/migrator/utils.ts
fisehara: Jupp, sorry for overlook it.

src/migrator/utils.ts
```diff
@@ -152,3 +278,95 @@ export const migrationTablesExist = async (tx: Tx) => {
...
+		// we ignore all errors to not spam the api logs
```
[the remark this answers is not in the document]
fisehara: Alright, I guess throwing the exeception is fine then.

src/sbvr-api/sbvr-utils.ts
```diff
@@ -588,6 +606,19 @@ export const executeModels = async (
...
+		await Promise.all(
+			execModels.map(async (model) => {
+				if (
+					(model.abstractSql != null || model.modelText != null) &&
+					model.migrations != null &&
+					model.apiRoot != null &&
+					Object.keys(model.migrations).includes(MigrationCategories.async)
+				) {
+					await executeAsyncModelMigration(tx, model);
+				}
```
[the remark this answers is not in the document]
fisehara: Great, thanks. This way it's the same call hierarchy as the syncMigrator.run.
I had to rework the getRunnableAsyncMigrations to execute only asyncMigrations.

src/migrator/async.ts
```diff
@@ -0,0 +1,304 @@
...
+							// using finally as it will also run when return statement is called inside the try block
+							// either success or error release the lock
+							await updateMigrationStatus(tx, migrationState);
+						}
+					});
+				});
+			};
+
+			let asyncScheduler = setInterval(asyncRunner, initMigrationState.delayMS);
```
Page-, line 301: I think this should be a `setTimeout` so that the delay between executions is based on when the previous one stopped rather than when it started. I also think it'll help with a lot of the manual `clearInterval`s that are necessary

src/config-loader/config-loader.ts
```diff
@@ -277,13 +277,21 @@ export const setup = (app: Express.Application) => {
...
 								 *
 								 */
 								const assignMigrationWithCategory = (
 									newMigrationKey: string,
 									newMigration: Migration,
 								) => {
 									const catMigrations = migrations[migrationCategory] || {};
+									if (
+										typeof newMigration === 'object' &&
+										migrationCategory === 'async'
```
Page-, line 291:
```suggestion
										migrationCategory === MigrationCategories.async
```

src/config-loader/config-loader.ts
```diff
@@ -277,13 +277,21 @@ export const setup = (app: Express.Application) => {
...
+									) {
+										newMigration.type = 'async';
```
Page-, line 293:
```suggestion
										newMigration.type = MigrationCategories.async;
```

src/config-loader/config-loader.ts
```diff
@@ -302,6 +310,14 @@ export const setup = (app: Express.Application) => {
 										);
 										break;
 									case '.sql':
+										if (migrationCategory === 'async') {
```
Page-, line 313:
```suggestion
										if (migrationCategory === MigrationCategories.async) {
```

src/migrator/utils.ts
```diff
@@ -23,12 +24,134 @@ export type CategorizedMigrations = {
...
+	[key in `${x}Sql`]?: undefined;
+};
+type AddSql<T extends {}, x extends 'sync' | 'async'> = T & {
+	[key in `${x}Fn`]?: undefined;
+} & {
+	[key in `${x}Sql`]: string;
+};
 
+type BaseAsyncMigration = {
+	type?: 'async';
```
Page-, line 49:
```suggestion
	type?: MigrationCategories.async;
```

src/migrator/async.ts
fisehara: Oh right! Thanks for cathing this.

src/migrator/utils.ts
```diff
@@ -23,12 +24,134 @@ export type CategorizedMigrations = {
...
+};
+
+export const getRunnableAsyncMigrations = (
+	migrations: Migrations,
+): RunnableAsyncMigrations | undefined => {
+	if (migrations[MigrationCategories.async]) {
+		if (
+			Object.values(migrations[MigrationCategories.async]).every((migration) =>
+				isAsyncMigration(migration),
+			)
```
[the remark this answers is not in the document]
fisehara: I'm going to full stop in this condition and throw an error, as the async migration type is mandatory and is put into place by the config-loader

src/config-loader/config-loader.ts
fisehara: thx

src/migrator/utils.ts
fisehara: thx

src/config-loader/config-loader.ts
fisehara: thx

src/config-loader/config-loader.ts
fisehara: thx

src/migrator/utils.ts
```diff
@@ -152,3 +275,96 @@ export const migrationTablesExist = async (tx: Tx) => {
...
+INSERT INTO "migration status" ("migration key", "start time", "is backoff")
+VALUES (${1}, ${2}, ${3});
+`,
+			[
+				migrationStatus['migration_key'],
+				migrationStatus['start_time'],
+				migrationStatus['is_backoff'] ? 1 : 0,
+			],
+		);
+		return result;
```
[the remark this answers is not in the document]
fisehara: done

docs/Migrations.md
```diff
@@ -46,4 +46,112 @@ Migrations which are Node modules should export a function which will be called
...
 You can execute SQL using `tx.executeSql(query)`, however note that the model for which the migration is running will *not* be available under `sbvrUtils.api`, as it has not executed yet.
 
-You should return a promise so pinejs can wait for completion/errors.
\ No newline at end of file
+You should return a promise so pinejs can wait for completion/errors.
+
+
+## Async Migrations
+
+Migrations that may lock tables for a long time, eg. column renames on large tables, can be run as async migrations.
```
thgreasi, line 54: I think column renames only affect the catalog. Maybe we should say something like:
```suggestion
Migrations that may lock tables for a long time, eg. migrating data between columns on large tables, can be run as async migrations.
```

docs/Migrations.md
```diff
@@ -46,4 +46,112 @@ Migrations which are Node modules should export a function which will be called
...
+    - Add new column (with independent sync migration) to contain new data and add code accessing the new column
+	- Old code reads/writes data to/from old column and async migration migrates data from old column to new column
+	- New code reads/writes to/from new column
+	- Async migrator runs forever
+2. Deployment
+	- Finalize async migration => only sync migration part gets executed
+	- Sync migration migrates all left over data from old column to new column
+	- Sync migration deletes old column
+	- Only new code with reads/writes to/from new column 
+	- Disable all old code that reads/writes to old column
```
thgreasi, line 80: In a multi-instance zero-downtime scenario, I think we should do the following (following how we do it today):
```suggestion
* Deployment 1
	- Add new column (with independent sync migration) to contain new data and add code accessing the new column.
	- Update the service's implementation to set both the old & new column on each write.
	- The service's implementation should only read the old column since the async migration still migrates data from old column to new column.
	- Async migrator runs forever.
* Deployment 2
	- Finalize async migration => only sync migration part gets executed.
	- Sync migration migrates all left over data from old column to new column.
	- Update the service's implementation to only read the new column, but still write the old one as well.
* Deployment 3
	- Update the service's implementation to stop settings the old column and remove it from the sbvr.
	- Make the old field NULLable if it isn't.
* Deployment 4
	- Delete the old column with a sync migration.
```

docs/Migrations.md
```diff
@@ -46,4 +46,112 @@ Migrations which are Node modules should export a function which will be called
...
+        `,
+	syncSql: `\
+        UPDATE "device"
+        SET "note" = "device"."name"
+        WHERE id IN (   SELECT id FROM (
+                                                SELECT id FROM "device"
+                                                WHERE  "device"."name" <> "device"."note" OR "device"."note" IS NULL
+                                                ) tmp
+                                );
+        `,
```
thgreasi, line 106: Do we really need the `tmp` subselect?
```suggestion
	asyncSql: `\
	UPDATE "device"
	SET "note" = "device"."name"
	WHERE id IN (
		SELECT id FROM "device"
		WHERE  "device"."name" <> "device"."note" OR "device"."note" IS NULL
		LIMIT %%ASYNC_BATCH_SIZE%%
	);
	`,
	syncSql: `\
	UPDATE "device"
	SET "note" = "device"."name"
	WHERE  "device"."name" <> "device"."note" OR "device"."note" IS NULL;
	`,
```
I think the indentation was weird as well. Mind the mixed spaced & tabs.

docs/Migrations.md
```diff
@@ -46,4 +46,112 @@ Migrations which are Node modules should export a function which will be called
...
+	asyncFn: async (tx: any, options) => {
+		const staticSql = `\
+		UPDATE "device"
+		SET "note" = "device"."name"
+		WHERE id IN (   SELECT id FROM (
+							SELECT id FROM "device"
+							WHERE  "device"."name" <> "device"."note" OR "device"."note" IS NULL
+							LIMIT ${options.batchSize}
+							) tmp
+					);
```
thgreasi, line 130:
```suggestion
		const staticSql = `\
		UPDATE "device"
		SET "note" = "device"."name"
		WHERE id IN (
			SELECT id FROM "device"
			WHERE  "device"."name" <> "device"."note" OR "device"."note" IS NULL
			LIMIT ${options.batchSize}
		);
```

docs/Migrations.md
```diff
@@ -46,4 +46,112 @@ Migrations which are Node modules should export a function which will be called
...
+	syncFn: async (tx: any) => {
+		const staticSql = `\
+		UPDATE "device"
+		SET "note" = "device"."name"
+		WHERE id IN (   SELECT id FROM (
+							SELECT id FROM "device"
+							WHERE  "device"."name" <> "device"."note" OR "device"."note" IS NULL
+							) tmp
+					);
+        `;
```
thgreasi, line 144:
```suggestion
		const staticSql = `\
		UPDATE "device"
		SET "note" = "device"."name"
		WHERE  "device"."name" <> "device"."note" OR "device"."note" IS NULL;
        `;
```

test/fixtures/03-async-migrator/04-migration-errors/migrations/0001-successful-migration.async.ts
```diff
@@ -0,0 +1,35 @@
...
+					);
+        `;
+
+		await tx.executeSql(staticSql);
+	},
+	asyncBatchSize: 1,
+	delayMS: 250,
+	backoffDelayMS: 4000,
+	errorThreshold: 15,
+} as AsyncMigration;
```
thgreasi, line 35: We should be able to get type inference & better type checking if we change this (and all other tests) to:
```suggestion
const migrations: AsyncMigration = {
	asyncFn: async (tx, options) => {
		const staticSql = `\
		UPDATE "device"
		SET "note" = "device"."name"
		WHERE id IN (   SELECT id FROM (
							SELECT id FROM "device"
							WHERE  "device"."name" <> "device"."note" OR "device"."note" IS NULL
							LIMIT ${options.batchSize}
							) tmp
					);
        `;

		return await tx.executeSql(staticSql);
	},
	syncFn: async (tx) => {
		const staticSql = `\
		UPDATE "device"
		SET "note" = "device"."name"
		WHERE id IN (   SELECT id FROM (
							SELECT id FROM "device"
							WHERE  "device"."name" <> "device"."note" OR "device"."note" IS NULL
							) tmp
					);
        `;

		await tx.executeSql(staticSql);
	},
	asyncBatchSize: 1,
	delayMS: 250,
	backoffDelayMS: 4000,
	errorThreshold: 15,
};
export = migration;
```

test/fixtures/03-async-migrator/03-finalize-async/migrations/m0002-data-copy.async.ts
```diff
@@ -0,0 +1,28 @@
+import { AsyncMigration } from '../../../../../src/migrator/utils';
+
+export = {
+	asyncSql: `\
+        UPDATE "device"
+        SET "note" = "device"."name"
```
thgreasi, line 6: Should this be setting a different column or table than `test/fixtures/03-async-migrator/03-finalize-async/migrations/m0001-data-copy.async.ts`?
Should it be using the `deviceb` table like in `test/fixtures/03-async-migrator/02-parallel-migrations/migrations/0002-migrate-testb.async.ts`?

thgreasi (commented):
Will continue on Tue...

src/migrator/migrations.sbvr
```diff
@@ -21,3 +21,52 @@ Term: migration lock
...
+
+Fact Type: migration status is backoff
+
+Fact Type: migration status has converged time 
+	Necessity: each migration status has at most one converged time
+
+Fact Type: migration status has last error message
+	Necessity: each migration status has at most one last error message
+
+Fact Type: migration status should stop
```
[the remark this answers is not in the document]
fisehara: @thgreasi this would be some kind of debugging tool to understand why the async migrations fail.
I'll double check if the same message will be in the logs as well and will then delete the column/field
Thanks!

docs/Migrations.md
fisehara: @thgreasi great thx!

test/fixtures/03-async-migrator/03-finalize-async/migrations/m0002-data-copy.async.ts
fisehara: yeah, thanks

src/migrator/migrations.sbvr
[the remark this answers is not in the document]
fisehara: @thgreasi it will be somewhere in the service logs. Skipping it in the DB.
Thanks for catching it!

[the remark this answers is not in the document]
thgreasi: @fisehara I meant that I don't see the `migration status should stop` being used anywhere.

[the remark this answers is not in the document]
fisehara: @thgreasi it was meant to be a DB queriable information during debugging on production database. But as this information is also logged to the api system logs it's redundant and is not needed to be persisted in the DB.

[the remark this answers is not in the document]
Page-: I think the point is that you can remove it from here :P

src/migrator/async.ts
```diff
@@ -0,0 +1,292 @@
...
+							migrationState.last_run_time = new Date();
+							migrationState.run_count += 1;
+
+							let migratedRows = 0;
+							// here a separate transaction is needed as this migration may fail
+							// when it fails it would break the transaction for managing the migration status
+							await sbvrUtils.db.transaction(async (migrationTx) => {
+								migratedRows =
+									(await asyncRunnerMigratorFn?.(migrationTx)) ?? 0;
+							});
```
Page-, line 244:
```suggestion
							// here a separate transaction is needed as this migration may fail
							// when it fails it would break the transaction for managing the migration status
							const migratedRows = await sbvrUtils.db.transaction(async (migrationTx) => {
								return (await asyncRunnerMigratorFn?.(migrationTx)) ?? 0;
							});
```

src/migrator/migrations.sbvr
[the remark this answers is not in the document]
fisehara: Oh right... I was mixing it up with the last error message, which I deleted already. Yes the 'should stop' can be deleted.

src/migrator/async.ts
```diff
@@ -0,0 +1,292 @@
...
+	 * executed migrations at this point should always contain all sync migrations:
+	 * executed migrations: 		[001,002,004,005]
+	 *
+	 * This will result in only async migration 006 being executed.
+	 *
+	 * The async migrations are meant to be used in separate deployments to make expensive data migrations
+	 * of multiple million row update queries cheaper and with no downtime / long lasting table lock.
+	 * In the end, after each async migration, the next deployment should follow up the data migration
+	 * with a final sync data migrations.
+	 * A async migration will be executed in iterations until no rows are updated anymore (keep row locks short)
```
thgreasi, line 85: NIT:
```suggestion
	 * An async migration will be executed in iterations until no rows are updated anymore (keep row locks short)
```

src/migrator/async.ts
```diff
@@ -0,0 +1,292 @@
...
+	 * 	           FROM tableA
+	 *             WHERE (columnA <> columnB) OR (columnA IS NOT NULL AND columnB IS NULL)
+	 *             LIMIT 1000);
+	 *
+	 * The final sync data migration would look like:
+	 * UPDATE tableA
+	 * SET columnB = columnA
+	 * WHERE (columnA <> columnB) OR (columnA IS NOT NULL AND columnB IS NULL);
+	 *
+	 * And will update remaining rows, which ideally are 0 and therefore now rows are locked for the update
```
thgreasi, line 100: NIT:
```suggestion
	 * And will update remaining rows, which ideally are 0 and therefore no rows are locked for the update
```

src/migrator/utils.ts
```diff
@@ -152,3 +272,92 @@ export const migrationTablesExist = async (tx: Tx) => {
...
+				migrationStatus['start_time'],
+				migrationStatus['is_backing_off'] ? 1 : 0,
+			],
+		);
+	} catch {
+		// it already exists and we can ignore this error.
+	}
+};
+
+// Just update the "modified at" field to show liveness
```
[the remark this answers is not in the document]
fisehara: good catch :) thx

src/migrator/utils.ts
```diff
@@ -23,12 +24,131 @@ export type CategorizedMigrations = {
...
+			Object.values(migrations[MigrationCategories.async]).every((migration) =>
+				isAsyncMigration(migration),
+			)
+		) {
+			return migrations[MigrationCategories.async] as RunnableAsyncMigrations;
+		} else {
+			throw new Error(
+				`All loaded async migrations need to be of type: ${MigrationCategories.async}`,
+			);
+		}
```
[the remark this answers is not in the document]
fisehara: I like that pattern! Thanks for raising @thgreasi :)

src/migrator/async.ts
```diff
@@ -0,0 +1,293 @@
...
+							// skip execution
+							if (migrationState.last_run_time) {
+								const durationSinceLastRun =
+									Date.now() - migrationState.last_run_time.getTime();
+								if (
+									(migrationState.is_backing_off &&
+										durationSinceLastRun < initMigrationState.backoffDelayMS) ||
+									(!migrationState.is_backing_off &&
+										durationSinceLastRun < initMigrationState.delayMS)
+								) {
```
thgreasi, line 229:
```suggestion
								const delayMs = migrationState.is_backing_off ? initMigrationState.backoffDelayMS : initMigrationState.delayMS;
								if (durationSinceLastRun < delayMs) {
```

src/migrator/migrations.sbvr
```diff
@@ -21,3 +21,46 @@ Term: migration lock
...
+	Necessity: each migration status has exactly one migration key
+
+Fact Type: migration status has start time
+	Necessity: each migration status has at most one start time
+
+Fact Type: migration status has last run time
+	Necessity: each migration status has at most one last run time
+
+Fact Type: migration status has run count
+	Necessity: each migration status has at most one run count
```
thgreasi, line 55: I guess that this could be or we should change the typings to mark it as nullable
```suggestion
	Necessity: each migration status has exactly one run count
```

src/migrator/async.ts
fisehara: thx, this is clearer

src/migrator/async.ts
```diff
@@ -0,0 +1,293 @@
...
+			}
+
+			initMigrationState.backoffDelayMS =
+				migration.backoffDelayMS ||
+				migratorEnv.asyncMigrationDefaultBackoffDelayMS;
+			initMigrationState.delayMS =
+				migration.delayMS || migratorEnv.asyncMigrationDefaultDelayMS;
+			initMigrationState.error_threshold =
+				migration.errorThreshold ||
+				migratorEnv.asyncMigrationDefaultErrorThreshold;
```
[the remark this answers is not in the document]
fisehara: Thanks a lot! This is very elegant.

src/migrator/migrations.sbvr
fisehara: right, thx!

src/migrator/utils.ts
```diff
@@ -23,12 +24,134 @@ export type CategorizedMigrations = {
...
+	return containsCategories;
+}
+
+export type Migration = string | MigrationFn | AsyncMigration;
 export class MigrationError extends TypedError {}
 
+export type MigrationStatus = {
+	migration_key: string;
+	start_time: Date;
+	last_run_time: Date;
```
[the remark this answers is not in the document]
fisehara: thx

src/migrator/utils.ts
```diff
@@ -152,3 +275,95 @@ export const migrationTablesExist = async (tx: Tx) => {
...
+			binds`
+INSERT INTO "migration status" ("migration key", "start time", "is backing off")
+SELECT ${1}, ${2}, ${3}
+WHERE NOT EXISTS (SELECT 1 FROM "migration status" WHERE "migration key" = ${4})
+`,
+			[
+				migrationStatus['migration_key'],
+				migrationStatus['start_time'],
+				migrationStatus['is_backing_off'] ? 1 : 0,
+				migrationStatus['migration_key'],
```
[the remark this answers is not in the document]
fisehara: 1. binds`` does only allow a sequence of parameters so that $1, $2, $3, $1 is failling.
2. When I change binds to accept this ordered list of bind keys the tx.executeSql will throw: `DatabaseError: inconsistent types deduced for parameter $1`
This comes from different types deduced from the paramter for the SELECT and the WHERE clause.
@thgreasi so I'm going to just use a duplicate 4th parameter.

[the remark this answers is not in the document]
thgreasi: ok cool :+1:

src/migrator/utils.ts
```diff
@@ -23,12 +24,131 @@ export type CategorizedMigrations = {
...
+): migration is AsyncMigration {
+	return (migration as AsyncMigration).type === MigrationCategories.async;
+}
+
+export function areCategorizedMigrations(
+	migrations: Migrations,
+): migrations is CategorizedMigrations {
+	const keys = Object.keys(migrations);
+	const containsCategories = keys.some((key) => key in MigrationCategories);
+	if (containsCategories && !keys.every((key) => key in MigrationCategories)) {
```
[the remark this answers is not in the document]
thgreasi: That as well, but it's imo closer to how you would phrase it
> in case there are async migration keys in the object, there should not be anything else.

thgreasi (commented):
Code LGTM :+1: 
Self-certifiable unless you also want @Page- 's final stamp :)

pranasziaukas:
FYI @ja-ilija there might soon be some progress on the asynchronous migrator

fisehara:
@pranasziaukas @ja-ilija The only blocker is the failing  Flowzone workflow. When I've fixed it, the PR gets merged :)

fisehara:
@balena-ci I self-certify!

PR #524, opened 2022-01-24
Fix permissions using `canAccess` that would fail if target resource specified no filter

dfunckt:
Change-type: patch

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -474,7 +474,7 @@ const buildODataPermission = (
 	}
 	if (conditionalPerms === true) {
 		// If we have full access then no need to provide a constrained definition
-		return false;
+		return;
 	}
 
 	const permissionFilters = nestedCheck(conditionalPerms, (permissionCheck) => {
@@ -517,6 +517,12 @@ const generateConstrainedAbstractSql = (
 		resourceName,
 		odata,
 	);
+
+	if (collapsePermissionFilters == null) {
+		// If we have full access then there's no need to provide a constrained
+		// definition, just use the table directly.
+		return;
+	}
 
 	_.set(odata, ['tree', 'options', '$filter'], collapsedPermissionFilters);
 
@@ -607,10 +613,9 @@ const generateConstrainedAbstractSql = (
 			parentOdata,
 		);
 
-		if (collapsedParentPermissionFilters === false) {
-			// We reuse a constant permission error here as it will be cached, and
-			// using a single error instance can drastically reduce the memory used
-			throw constrainedPermissionError;
+		if (collapsedParentPermissionFilters == null) {
+			// full access
+			return ['Equals', ['Boolean', true], ['Boolean', true]];
 		}
 
 		rewriteSubPermissionBindings(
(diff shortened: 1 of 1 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -607,24 +607,20 @@ const generateConstrainedAbstractSql = (
...
-			collapsedParentPermissionFilters,
-			this.bindVarsLength + this.extraBindVars.length,
-		);
-		convertToLambda(collapsedParentPermissionFilters, lambdaId);
-		resolvedProperty.lambda = {
-			method: 'any',
-			identifier: lambdaId,
-			expression: collapsedParentPermissionFilters,
-		};
+		if (collapsedParentPermissionFilters != null) {
```
Page-, line 610: I think you can immediately return `['Equals', ['Boolean', true], ['Boolean', true]]` to avoid the extra work and subquery

Page- (commented):
There's another 2 uses of `buildODataPermission` that also need a similar treatment

src/sbvr-api/permissions.ts
```diff
@@ -518,7 +518,11 @@ const generateConstrainedAbstractSql = (
 		odata,
 	);
 
-	_.set(odata, ['tree', 'options', '$filter'], collapsedPermissionFilters);
+	_.set(
+		odata,
+		['tree', 'options', '$filter'],
+		collapsedPermissionFilters || false,
```
Page-, line 524: As a general rule it's much better to use `??` if you're only planning to default null/undefined as it makes the intent much clearer and avoids questions around whether it was intentional to default all falsey values or an oversight, particularly if the types change due to other later changes

src/sbvr-api/permissions.ts
```diff
@@ -518,7 +518,11 @@ const generateConstrainedAbstractSql = (
 		odata,
 	);
 
-	_.set(odata, ['tree', 'options', '$filter'], collapsedPermissionFilters);
+	_.set(
```
Page-, line 521: I think this should instead be
```ts
if (collapsedPermissionFilters == null) {
	return;
}
```
or `return false`

src/sbvr-api/permissions.ts
```diff
@@ -810,7 +813,7 @@ const rewriteRelationship = memoizeWeak(
 						_.set(
```
Page-, line 813: I think this can also have
```ts
if (collapsedPermissionFilters == null) {
	// If we have full access already then there's no need to check for/rewrite based on `canAccess`
	return;
}

**Comment by dfunckt** [comment]
No regressions @Page- 👍 

- https://github.com/balena-io/open-balena-api/pull/937
- https://github.com/balena-io/balena-api/pull/3519

File: src/sbvr-api/permissions.ts [hunk 5]
```diff
@@ -518,6 +518,11 @@ const generateConstrainedAbstractSql = (
 		odata,
 	);
 
+	if (collapsePermissionFilters == null) {
+		// full access
```
Page-, line 522: I'd like to expand on this as
```suggestion
		// If we have full access then there's no need to provide a constrained definition, just use the table directly
```
since it'll help a lot with understanding why full access == return [undefined]

Page- (approved):

PR #525, opened 2022-02-07
Remove bluebird

Page-:


src/sbvr-api/control-flow.ts
```diff
@@ -64,3 +69,32 @@ export const getMappingFn = (headers?: {
...
+
+export const delay = (ms: number) =>
+	new Promise<void>((resolve) => setTimeout(resolve, ms));
+
+export const fromCallback = <T>(
+	resolver: (callback: (err: any, result?: T) => void) => void,
+): Promise<T> =>
+	new Promise<T>((resolve, reject) => {
+		resolver((err, result?: T) => {
+			if (err || !result) {
```
thgreasi, line 81: I didn't expect to have `|| !result` here :thinking: 
Aren't falsy results valid callback result types?

Page-: You're right, I'll stop taking shortcuts on the typing

Page-: Nevermind, I can't type it quite right because it has to be compatible with the places it's passed to which also take shortcuts so I've gone with a cast instead

PR #527, opened 2022-02-07
Add the ability to set the maxLifetimeSeconds for the postgres pool

Page-:
Change-type: minor


Page-:
See: https://github.com/brianc/node-postgres/pull/2698/files

Page-:
@balena-ci I self-certify!

PR #531, opened 2022-03-18
Add logo and header to README

LucianBuzzo:


fisehara:
@balena-ci I self-certify!

README.md
```diff
@@ -1,4 +1,9 @@
-# Pine.js
+<div align="center">
+  <img width="400" height="auto" src="https://raw.githubusercontent.com/balena-io/pinejs/master/pinejs.png">
```
fisehara, line 2:
``` suggestion
<img width="400" height="auto" src="pinejs.png">
```

fisehara (changes requested):

README.md
LucianBuzzo: IIRC this won't work when you view the repo README, but I may be wrong...

fisehara: As long as the view is here in GH it should show up well. Is this needed for landr or for local cloned repo view?

LucianBuzzo: @fisehara Yeh I believe so, it's the approach I've always taken and I think it's because in some circumstances this path won't work.

fisehara: Okay, fine for me then :)

fisehara (approved):

PR #532, opened 2022-03-18
Fixed unneccessary @ts-expect-error

fisehara:
PG Types have added maxUses as key and ts except is not needed any more

Change-type: patch


fisehara:
@balena-ci I self-certify!

PR #537, opened 2022-04-01
Update dependencies

Page-:
Update pinejs-client-core from 6.10.0 to 6.10.1

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #538, opened 2022-04-05
Add an expiry date to 'api key'

nitishagar:
Change-Type: minor

Spec: https://jel.ly.fish/e450713f-001c-4470-b571-71724aeb0869
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/vx-rBrpDBqXtkdtMmSU5MmPYSlb

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -1252,6 +1252,16 @@ const $getApiKeyPermissions = (() => {
 												$alias: 'k',
 												$expr: {
 													k: { key: { '@': 'apiKey' } },
+													$or: [
+														{
+															k: { expiry_date: null },
+														},
+														{
+															k: {
+																expiry_date: { $gt: { $now: null } },
+															},
+														},
+													],
 												},
 											},
 										},
@@ -1279,6 +1289,18 @@ const $getApiKeyPermissions = (() => {
 																				$alias: 'k',
 																				$expr: {
 																					k: { key: { '@': 'apiKey' } },
+																					$or: [
+																						{
+																							k: { expiry_date: null },
+																						},
+																						{
+																							k: {
+																								expiry_date: {
+																									$gt: { $now: null },
+																								},
+																							},
+																						},
+																					],
 																				},
 																			},
 																		},
@@ -1343,6 +1365,12 @@ const getApiKeyActorId = (() => {
 			},
 			options: {
 				$select: 'is_of__actor',
+				$filter: {
+					$or: [
+						{ expiry_date: null },
+						{ expiry_date: { $gt: { $now: null } } },
+					],
+				},
 			},
 		}),
 	);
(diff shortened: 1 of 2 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -1256,6 +1256,16 @@ const $getApiKeyPermissions = (() => {
 											},
 										},
 									},
+									$or: [
```
nitishagar, line 1259: @thgreasi @Page- I am not sure about this specific addition, should we be doing a expiry check while doing a permissions query?

thgreasi: I think that makes sense @nitishagar but you probably need to add the same check in the `is_of__role` part of the query as well.
And on top of that you can also add it to the `getApiKeyActorId()`, so that pine treats the api key as unknown in all cases.

src/sbvr-api/permissions.ts
```diff
@@ -1256,6 +1256,16 @@ const $getApiKeyPermissions = (() => {
...
+										{
+											khr: { expiry_date: null },
```
thgreasi, line 1261: Did you have the chance to give this and the one right after this a try with the API?
B/c I think this filter is in the wrong place and should be erroring, since it filters on the user_has_role.expiry_date & user_has_permission.expiry_date, which do not exist.
You need to move them deeper in the nesting next to the `api_key` where you will be using `k` as the alias.

src/sbvr-api/permissions.ts
```diff
@@ -1670,6 +1698,10 @@ export const config = {
 					ALTER TABLE "role-has-permission"
 					ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
 				`,
+				'14.41.0-modified-at': `
```
nitishagar, line 1701: @thgreasi @Page- Do we need this pinned to latest release?

thgreasi: @nitishagar probably yes and also `modified-at` doesn't sound descriptive of what the migration does. Maybe
```suggestion
				'14.42.0-api-key-expiry-date': `
```

src/sbvr-api/permissions.ts
```diff
@@ -1252,6 +1252,16 @@ const $getApiKeyPermissions = (() => {
 												$alias: 'k',
 												$expr: {
 													k: { key: { '@': 'apiKey' } },
+													$or: [
+														{
+															k: { expiry_date: null },
```
thgreasi, line 1257: Did I ever submitted my review about changing the `khp` to `k` here?
B/c I can't find it in the GH comments atm !?!?!?

src/sbvr-api/permissions.ts
nitishagar: Should we be doing `..-modified-at` like other changes above or `..api-key-expiry-date`?

src/sbvr-api/permissions.ts
nitishagar: My initial changes had that when the `$or` was a level above. Is there something else to add/update?

src/sbvr-api/permissions.ts
thgreasi: We should call it `...-api-key-expiry-date`. The above two the say `modified at` are indeed migrations about that field.

Page- (approved):

PR #541, opened 2022-04-20
Use the db engine from the existing request when getting affected ids

Page-:
In practice the module level `db.engine` will always match the
`request.engine` but this future proofs things and makes it clear where
this value actually comes from

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #545, opened 2022-05-16
Use the modern definition format for permission constrained definitions

Page-:
Change-type: patch


Page-:
@balena-ci retest

Page-:
@balena-ci republish

PR #546, opened 2022-05-16
Generalisation of migrator

fisehara:
Decouple migrator functions

Add basic test for sync migrator


src/migrator/sync.ts
```diff
@@ -0,0 +1,167 @@
...
+
+			const newlyExecutedMigrations = await executeMigrations(
+				tx,
+				pendingMigrations,
+			);
+			await setExecutedMigrations(tx, modelName, [
+				...executedMigrations,
+				...newlyExecutedMigrations,
+			]);
+		} catch (err) {
```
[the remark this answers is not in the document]
fisehara: @Page- thanks! 
Going to add:
```
throw new MigrationError(err);
```

src/migrator/utils.ts
```diff
@@ -0,0 +1,159 @@
...
+	fn: () => Promise<T>,
+): Promise<T> => {
+	try {
+		await tx.executeSql(
+			binds`
+		DELETE FROM "migration lock"
+		WHERE "model name" = ${1}
+		AND "created at" < ${2}`,
+			[modelName, new Date(Date.now() - migratorEnv.lockTimeout)],
+		);
```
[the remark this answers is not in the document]
fisehara: Alright, I'll go for outdented format then, as it's in the other locations.

src/config-loader/config-loader.ts
```diff
@@ -251,18 +255,50 @@ export const setup = (app: Express.Application) => {
...
+								 *  key2-name2.sync.sql 		==> explicit synchrony migration
+								 *
+								 */
+								const assignMigrationWithCategory = (
+									newMigration: Migration,
+								) => {
+									migrations[migrationCategory] = Object.assign(
+										{ [migrationKey]: newMigration },
+										migrations[migrationCategory],
+									);
```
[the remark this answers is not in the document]
fisehara: I have not seen a guard for migration keys reused except the migration table where the keys are stored and will not be executed again, when already marked as executed in the migration table.

fisehara:
@balena-ci retest

src/migrator/sync.ts
```diff
@@ -0,0 +1,177 @@
...
+			(sbvrUtils.api.migrations?.logger.error ?? console.error)(
+				`$run executedMigrations error ${err}`,
```
Page-, line 107: Can you change this message so that it's clearer to a user what actually failed and therefore where they should be looking for the issue in their code

src/migrator/sync.ts
```diff
@@ -0,0 +1,177 @@
...
+		migrations == null ||
+		_.isEmpty(migrations) ||
+		migrations[defaultMigrationCategory] === undefined
+	) {
+		return;
+	}
+
+	let runMigrations: RunnableMigrations = {};
+	for (const [key, value] of Object.entries(migrations)) {
+		if (key === defaultMigrationCategory && value instanceof Object) {
```
Page-, line 60: What's the reason for instanceof vs the more common typeof?
```suggestion
		if (key === defaultMigrationCategory && typeof value === 'object') {
```

src/migrator/sync.ts
```diff
@@ -0,0 +1,177 @@
...
+		});
+	}
+};
+
+export const run = async (tx: Tx, model: ApiRootModel): Promise<void> => {
+	const { migrations } = model;
+	if (
+		migrations == null ||
+		_.isEmpty(migrations) ||
+		migrations[defaultMigrationCategory] === undefined
```
Page-, line 53: Pretty sure this will never run the old form of translations because for those `migrations[defaultMigrationCategory]` will always be undefined
```suggestion
		_.isEmpty(migrations)
```

src/migrator/sync.ts
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	let runMigrations: RunnableMigrations = {};
+	for (const [key, value] of Object.entries(migrations)) {
+		if (key === defaultMigrationCategory && value instanceof Object) {
+			runMigrations = { ...runMigrations, ...value };
+		} else if (!(key in MigrationCategories)) {
+			runMigrations = { ...runMigrations, ...{ [key]: value } };
+		}
+	}
```
Page-, line 65: I think you can simplify this to
```suggestion
	const runMigrations: RunnableMigrations = defaultMigrationCategory in migrations ? migrations[defaultMigrationCategory] : migrations;
```
since you either have specify using the migrations categories or not specify at all - we don't allow mixing the 2 versions in the types and I don't think it makes sense for the code to allow it either (if you want you can also explicitly check that there's no non-migration category entries when using the migration category form)

src/migrator/sync.ts
fisehara: Yep, thx! deleting the defaultMigration check

src/migrator/sync.ts
fisehara: Mo specific reasoning. At some point my IDE (VS Code) was complainingabout checking for tyepof <var> === 'object' and I used instanceof. I didn't changed it back.

src/migrator/sync.ts
fisehara: Need to add a check if migrations[defaultMigrationCategory] is object and not string or function. It's a double ternary, which is 'still readable'

fisehara:
@balena-ci retest

src/migrator/sync.ts
Page-: I expect it was because whilst `typeof x === 'object'` is generally more standard `null` is also a type of object

Page- (approved):

PR #548, opened 2022-05-20
Update dependencies

Page-:
Update @balena/abstract-sql-compiler from 7.19.1 to 7.20.0
Update @balena/odata-to-abstract-sql from 5.4.9 to 5.4.11

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #549, opened 2022-05-26
permissions: Add optional transaction param to user & api key methods

thgreasi:
Change-type: minor
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/H1xvtNzl8i5atMElV51nix3uWGf

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -20,6 +20,7 @@ import type {
 	ODataQuery,
 	SupportedMethod,
 } from '@balena/odata-parser';
+import type { Tx } from '../database-layer/db';
 import type { ApiKey, User } from '../sbvr-api/sbvr-utils';
 import type { AnyObject } from './common-types';
 
@@ -345,7 +346,6 @@ const getPermissionsLookup = env.createCache(
 		return permissionsLookup;
 	},
 	{
-		weak: undefined,
 		normalizer: ([permissions, guestPermissions]) =>
 			// When guestPermissions is present it should always be the same, so we can key by presence not content
 			`${permissions}${guestPermissions == null}`,
@@ -1213,19 +1213,27 @@ const $getUserPermissions = (() => {
 	);
 	return env.createCache(
 		'userPermissions',
-		async (userId: number) => {
-			const permissions = (await getUserPermissionsQuery()({
-				userId,
-			})) as Array<{ name: string }>;
+		async (userId: number, tx?: Tx) => {
+			const permissions = (await getUserPermissionsQuery()(
+				{
+					userId,
+				},
+				undefined,
+				{ tx },
+			)) as Array<{ name: string }>;
 			return permissions.map((permission) => permission.name);
 		},
 		{
 			primitive: true,
 			promise: true,
+			normalizer: ([userId]) => `${userId}`,
 		},
 	);
 })();
-export const getUserPermissions = async (userId: number): Promise<string[]> => {
+export const getUserPermissions = async (
+	userId: number,
+	tx?: Tx,
+): Promise<string[]> => {
 	if (typeof userId === 'string') {
 		userId = parseInt(userId, 10);
 	}
@@ -1233,7 +1241,7 @@ export const getUserPermissions = async (userId: number): Promise<string[]> => {
 		throw new Error(`User ID has to be numeric, got: ${typeof userId}`);
 	}
 	try {
-		return await $getUserPermissions(userId);
+		return await $getUserPermissions(userId, tx);
 	} catch (err) {
 		sbvrUtils.api.Auth.logger.error('Error loading user permissions', err);
 		throw err;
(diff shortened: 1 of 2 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -1213,27 +1214,35 @@ const $getUserPermissions = (() => {
...
+				},
+				undefined,
+				{ tx },
+			)) as Array<{ name: string }>;
 			return permissions.map((permission) => permission.name);
 		},
 		{
 			primitive: true,
 			promise: true,
+			normalizer: ([userId]: [number]) => `${userId}`,
```
Page-, line 1230: I'm pretty sure it should be able to infer the type here from the fn above:
```suggestion
			normalizer: ([userId]) => `${userId}`,
```

src/sbvr-api/permissions.ts
```diff
@@ -1213,27 +1214,35 @@ const $getUserPermissions = (() => {
...
 		},
 	);
 })();
-export const getUserPermissions = async (userId: number): Promise<string[]> => {
+export const getUserPermissions = async (
+	userId: number,
+	tx?: Tx,
```
Page-, line 1236: Does it make sense for anywhere within pinejs to also be passing this in?

src/sbvr-api/permissions.ts
thgreasi: @Page- Unfortunately it wasn't :(
image

I saw that we already use `weak: undefined` for `getPermissionsLookup`, which fixes TS inference of the normalizer.
See: https://github.com/balena-io/pinejs/blob/f5ee762de3fa3f9aa5d2779626ab60cf8e16faf8/src/sbvr-api/permissions.ts#L348

Tried TS 4.7 & using `extends` & `infer` w/o luck.

Options that I found working around the issue:
* Adding `weak: undefined`.
* Making the `opts` argument mandatory.
* Adding two separate overloads for the `createCache()` one with & one w/o the `opts` argument (but that made tslint unhappy about doing it in a silly way rather than marking `opts` as optional.

Of all the above, making the `opts` mandatory felt the less hacky imo, since we anyway pass the options in all invocations in pine. Let me know if your prefer a different one.



See: https://www.typescriptlang.org/play?ts=4.6.4#code/JAFwngDgpgBAYgSwE4GcQEEkHMA8AVAPhgF4Y8YoAPEKAOwBMUYAKAQ2wEYAuGBWgMyhIYAVQCUJIq1pgYAflEwe0sAG4AsAChQkWACUoaTFhT4ipclRoMmbTspkAaGADo37Ezz6Dh4yTBV5RQcwAG0AXQ1NGBiYLVjeWhokflYAY1gAWSgAWwB7BAAvKAB1KFYAawB5CBAEPNpTOApqOkZ4AFdaNLqGogBvLWBgABs6LBAACzkeWg6cgCMhGAAfGFSRlCgo4ZzWSnQsKBmYOcWhHeA9yhOzpaRLiCQoOCgQNOnZ+fvVmBAkDrbIbAJ55HIILYnBZ5PJjaSXegQiB5SHMABurBGgJCYh4aIK9EurBQYG6UJhcNojyQCHBdTRxx40Nh5SpwNoeSQexGRSEcmY-GQRmwPEQqAw2BwcAIzmews8MAM8qaBFxARklzlsIZqBOmCQrDAODsWBx-hUBB2AF94rE+MlUhkYNl8ryar1GlKWtZ2sw3C4PCgQhEJMQpDIBraYmNaBNPqdvss1hstr8uvQoILaFBCVGYNdDoyE+dhGt05m+DmoglrrdE6WYOWs1W808Xm8PnWS79-oC0wwK9nc9FYqDwZCmRTWb8AERTOgz2f0BpQRdrGfL7NcLOYkZgNeNgfN4cJREoZGojFYqA4vEE6uxYmktLklnSfsZ48PmJPWkIelFsylIfoOLYjjEHJcpivJIPygY8AACuwrA5G8QgqmqaA0rG34wFqIw6igepIAaRommaYbqmARBlkelYnqOSB5IKYwAHIoUWWF8FgD42poWjgNAMAAMLpJMLy0O6piWK0NgsP68FUSG5oRiQQxrIM2jDL+dIIAyJy9kCWkgkx45FoZlzAJB3IwScLoFMUZSVO69SeoQoQAOTWdBxRIB5kTAsAADu5QVDwFlDDEfHABpgU6f+enmQCRnDCZYIQklgKWd5PK+XZuQOVALkNNJBCeTlMH+ZZIWVCcTb0TsUVRFo-BdD0rnrFUNJYHwmIwPg3ptLYCnYEGSnhKG4bUcwQz8LQPB4I4Qx5LURE8KJHwSVJZhLZoar4gg9AwP0VpRPwXUID1tCYswbBfCWzgLDwAAirA0JNh6fvRziacFoXhclu1WZyNm+TwfruKNH0AAYACT9IGVrQ7tVpiGdF1XTdd3Fvcj0vW9UAffVQ4-UMFVg-JkMmDD8OI8jWio81mitd0HrrAYACOHTIDmUn9TJPrDVTY0qMplEWjN2hzQtQMrSAY0beJcCSatO1aPtBLHadLWc9zzz0FJt2sPduMwE9MCve9-jEzmpPaDVYV-IDZMgz5QjgyN1P+HDCOjUjKNozrUBczzBurUbJtCHjFsE0TdEk8dLtQbl7uUwGUPe7Tfv05ojNaPn-GaIJsCK1tq0AELErApC-fFAEGclOxjhlDdZQzTPFyJYkSU51SqwLQ1p4posTSp1EkF3m3K1JlepgAZIn2jk3yPD2byvfFW5ZVea7Kd+QF9v-U7be5x3uiT0rtCsQ0G-94NckQ+nCojx9FoT6X08V1XMAL79y+wavAqbpaiuVKuVXelUD5-VqjwG2w5taF07h-FW8sRC0FcgNKwg9H7DxkGLKa5gL491CttQgvxkHX1oLfeWZgmYs3ag0dYABRWS7RMGsKFk-EWeDR7iycDAPmWC5LIKkmgjBhACCS2ANLMgstVonEERw46MAHYAz7HxBICQFDIOoaVPMCR1rd2VpQ3Ras9p3kOlrM6LDBYRxxlHM2+MraUTgXbaBjtDJA3-h7YWNNfYmH9gzQOzMbFDTsXcBx5tLaE2tvHW2i9gbJxgj4rhfi6YByZi1NqbN+ChIYAASQEMsdhgsh6jWDLwqazhFGlJEatMRDQzCSNmvNWRy15E8BqYPfoKij7eGWCUGAGjNExAUIMoR7RDJBB0SQ-uRBDFTyvjfWZNDCD6NiAsy+JiVmlV2hrSxJ1rGsMKT4cJ9Zo7RLjl9BOv1VHHygF4iBFMcEZ0oj7dJQSjk+hOUIM5D1HEx2cUQVxCTvFlK9m8rOASc5500EAA

src/sbvr-api/permissions.ts
thgreasi: @Page- the only places this is called inside pine is from `getGuestPermissions()` (used by `getReqPermissions(req, odataBinds)`) and `checkPassword()` (used by the `login()` middleware).
None of the two callers have an active transaction or do a write before those calls, so it doesn't seem necessary to create a read transaction just to pass it in.

Page- (approved):

PR #551, opened 2022-06-02
Delete default user permissions that are no longer in the config

thgreasi:
Change-type: major
See: https://www.flowdock.com/app/rulemotion/resin-devops/threads/-dlSgJYAMPMNZQ6mpxMlubGKYF6


thgreasi:
We decided that pine should not do that.

PR #553, opened 2022-06-08
Update dependencies

Page-:
Update @balena/odata-parser from 2.2.9 to 2.3.0
Update @balena/odata-to-abstract-sql from 5.4.11 to 5.4.12

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #556, opened 2022-06-24
Refactor checkModelExists to isModelNew

fisehara:
Modules may need to know if a model existed before initialisation of the instance.
Eg. migrations need to know if a model is freshly generated / executed
of if this model already existed and can be target for migrations.

Change-type: minor

### DIFF
--- a/src/migrator/sync.ts
+++ b/src/migrator/sync.ts
@@ -3,7 +3,6 @@ import {
 	MigrationTuple,
 	MigrationError,
 	defaultMigrationCategory,
-	checkModelAlreadyExists,
 	setExecutedMigrations,
 	getExecutedMigrations,
 	lockMigrations,
@@ -24,11 +23,10 @@ export const postRun = async (tx: Tx, model: ApiRootModel): Promise<void> => {
 	}
 
 	const modelName = model.apiRoot;
-
-	const exists = await checkModelAlreadyExists(tx, modelName);
-	if (!exists) {
+	const modelIsNew = await sbvrUtils.isModelNew(tx, modelName);
+	if (modelIsNew) {
 		(sbvrUtils.api.migrations?.logger.info ?? console.info)(
-			'First time executing, running init script',
+			`First time executing '${modelName}', running init script`,
 		);
 
 		await lockMigrations(tx, modelName, async () => {
@@ -69,10 +67,10 @@ const $run = async (
 
 	// migrations only run if the model has been executed before,
 	// to make changes that can't be automatically applied
-	const exists = await checkModelAlreadyExists(tx, modelName);
-	if (!exists) {
+	const modelIsNew = await sbvrUtils.isModelNew(tx, modelName);
+	if (modelIsNew) {
 		(sbvrUtils.api.migrations?.logger.info ?? console.info)(
-			'First time model has executed, skipping migrations',
+			`First time model '${modelName}' has executed, skipping migrations`,
 		);
 
 		return await setExecutedMigrations(tx, modelName, Object.keys(migrations));
--- a/src/migrator/utils.ts
+++ b/src/migrator/utils.ts
@@ -51,7 +51,11 @@ export const lockMigrations = async <T>(
 	tx: Tx,
 	modelName: string,
 	fn: () => Promise<T>,
-): Promise<T> => {
+): Promise<T | undefined> => {
+	if (!(await migrationTablesExist(tx))) {
+		return;
+	}
+
 	try {
 		await tx.executeSql(
 			binds`
@@ -88,26 +92,6 @@ WHERE "model name" = ${1}`,
 	}
 };
 
-export const checkModelAlreadyExists = async (
-	tx: Tx,
-	modelName: string,
-): Promise<boolean> => {
-	const result = await tx.tableList("name = 'migration'");
-	if (result.rows.length === 0) {
-		return false;
-	}
-	const { rows } = await tx.executeSql(
-		binds`
-SELECT 1
-FROM "model"
-WHERE "model"."is of-vocabulary" = ${1}
-LIMIT 1`,
-		[modelName],
-	);
-
-	return rows.length > 0;
-};
-
 export const setExecutedMigrations = async (
 	tx: Tx,
 	modelName: string,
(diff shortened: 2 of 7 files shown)


src/migrator/sync.ts
```diff
@@ -25,10 +24,10 @@ export const postRun = async (tx: Tx, model: ApiRootModel): Promise<void> => {
 
 	const modelName = model.apiRoot;
 
-	const exists = await checkModelAlreadyExists(tx, modelName);
+	const exists = await sbvrUtils.isModelNew(tx, modelName);
 	if (!exists) {
 		(sbvrUtils.api.migrations?.logger.info ?? console.info)(
-			'First time executing, running init script',
+			`First time executing ${modelName}, running init script`,
```
Page-, line 30: I find putting the variable part in quotes can help make it clear that you're quoting some user input
```suggestion
			`First time executing '${modelName}', running init script`,
```

src/migrator/sync.ts
```diff
@@ -69,10 +68,10 @@ const $run = async (
 
 	// migrations only run if the model has been executed before,
 	// to make changes that can't be automatically applied
-	const exists = await checkModelAlreadyExists(tx, modelName);
+	const exists = await sbvrUtils.isModelNew(tx, modelName);
 	if (!exists) {
 		(sbvrUtils.api.migrations?.logger.info ?? console.info)(
-			'First time model has executed, skipping migrations',
+			`First time model ${modelName} has executed, skipping migrations`,
```
Page-, line 74: I find putting the variable part in quotes can help make it clear that you're quoting some user input
```suggestion
			`First time model '${modelName}' has executed, skipping migrations`,
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -330,6 +330,32 @@ const prettifyConstraintError = (
...
+			`SELECT "is of-vocabulary" FROM "model";`,
+		);
+		cachedIsModelNew.models = new Set<string>(
+			rows.map((row) => row['is of-vocabulary']),
+		);
+		cachedIsModelNew.hasRun = true;
+	}
+
+	return cachedIsModelNew.models?.has(modelName) ?? false;
+};
```
Page-, line 357: You can simplify this by using the `undefined` vs defined states as the `hasRun` check, ie
```suggestion
const cachedIsModelNew: Set<string> | undefined } = undefined;

export const isModelNew = async (
	tx: Db.Tx,
	modelName: string,
): Promise<boolean> => {
	const result = await tx.tableList("name = 'model'");
	if (result.rows.length === 0) {
		return false;
	}
	if (cachedIsModelNew == null) {
		const { rows } = await tx.executeSql(
			`SELECT "is of-vocabulary" FROM "model";`,
		);
		cachedIsModelNew = new Set<string>(
			rows.map((row) => row['is of-vocabulary']),
		);
	}

	return cachedIsModelNew.has(modelName);
};
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -330,6 +330,32 @@ const prettifyConstraintError = (
...
+	tx: Db.Tx,
+	modelName: string,
+): Promise<boolean> => {
+	const result = await tx.tableList("name = 'model'");
+	if (result.rows.length === 0) {
+		return false;
+	}
+	if (!cachedIsModelNew.hasRun) {
+		const { rows } = await tx.executeSql(
+			`SELECT "is of-vocabulary" FROM "model";`,
```
Page-, line 348: Since you're switching to checking the `model` resource for existence is it now possible that the model could be created but then an error happen such that the `model` entry is populated but the `migrations` entry is not?

src/migrator/utils.ts
```diff
@@ -158,3 +145,14 @@ WHERE "migration"."model name" = ${1}`,
...
+
+export const migrationTablesExist = async (tx: Tx) => {
+	return (
+		await Promise.all(
+			['migration', 'migration lock'].map(async (tableName) => {
+				const result = await tx.tableList(`name = '${tableName}'`);
+				return result.rows.length > 0;
+			}),
+		)
+	).reduce((prevCheck, currCheck) => prevCheck && currCheck, true);
```
Page-, line 157: It'd make much more sense to use `every` here:
```suggestion
	).every((check) => check);
```
It's also got the benefit that it'll short-circuit return as necessary (although on only two items that doesn't matter, but in general it's nice)

src/sbvr-api/sbvr-utils.ts
Page-: Specifically I was thinking if it's possible for the `INSERT INTO "model" ...` to have been committed and then things fail so `isModelNew(...)` would return `false` but the migrations table wouldn't have an entry and therefore it would try to run all the historical migrations on a new model. I realized though this might not actually be a problem though as long as it's all run in the same transaction?

fisehara: Ah I see. I'll double check this and assume that this will be handled in a single transaction procedure that will fail.
Although using the method should be usable across different transactions for the usage of sync and async migration execution (which will have different transactions)

src/migrator/utils.ts
fisehara: Thx, very elegant!

src/sbvr-api/sbvr-utils.ts
fisehara: @Page-

From the execution order the `isModelNew(...)` right now is executed before the `INSERT INTO "model"` so that the database is evaluated as it was before starting up pine (before calling config-loader).

I've double checked that the isModelNew(...) query needs to run inside the:
https://github.com/balena-io/pinejs/blob/7789e2a9e3da739b33f4b2250527782696255bbc/src/config-loader/config-loader.ts#L93-L94 config-loader transaction. 
Do you agree on leaving a comment on the isModelNew so that it needs to be executed in one same transaction?

src/migrator/utils.ts
```diff
@@ -158,3 +145,14 @@ WHERE "migration"."model name" = ${1}`,
...
+
+export const migrationTablesExist = async (tx: Tx) => {
+	return (
+		await Promise.all(
+			['migration', 'migration lock'].map(async (tableName) => {
+				const result = await tx.tableList(`name = '${tableName}'`);
+				return result.rows.length > 0;
+			}),
+		)
+	).every((check) => check);
```
Page-, line 157: I think it'd be better to do this as a single query, eg:
```suggestion
	const tables = ['migration', 'migration lock'];
	const where = tables.map((tableName) => `name = '${tableName}'`).join(' OR ');
	const result = await tx.tableList(where);
	return result.rows.length === tables.length;
```

src/sbvr-api/sbvr-utils.ts
Page-: I think that's probably the way to go

Page- (approved):

PR #559, opened 2022-07-20
Narrow the scope of the permissions bypass for root permissions

Page-:
This changes the bypass to only occur before guest permissions are
initialized, only when specifically using `root` or `rootRead`
permissions, and also now constrains the model to the specific
permissions passed in rather than bypassing permissions altogether
for the root cases

Change-type: patch

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -1578,6 +1578,7 @@ export const checkPermissionsMiddleware =
 		}
 	};
 
+let guestPermissionsInitialized = false;
 const getGuestPermissions = memoize(
 	async () => {
 		// Get guest user
@@ -1601,6 +1602,7 @@ const getGuestPermissions = memoize(
 		if (guestPermissions.some((p) => DEFAULT_ACTOR_BIND_REGEX.test(p))) {
 			throw new Error('Guest permissions cannot reference actors');
 		}
+		guestPermissionsInitialized = true;
 		return guestPermissions;
 	},
 	{ promise: true },
@@ -1611,7 +1613,18 @@ const getReqPermissions = async (
 	odataBinds: ODataBinds = [],
 ) => {
 	const [guestPermissions] = await Promise.all([
-		getGuestPermissions(),
+		(async () => {
+			if (
+				guestPermissionsInitialized === false &&
+				(req.user === root.user || req.user === rootRead.user)
+			) {
+				// In the case that guest permissions are not initialized yet and the query is being made with root permissions
+				// then we need to bypass `getGuestPermissions` as it will cause an infinite loop back to here.
+				// Therefore to break that loop we just ignore guest permissions.
+				return [];
+			}
+			return await getGuestPermissions();
+		})(),
 		(async () => {
 			// TODO: Remove this extra actor ID lookup making actor non-optional and updating open-balena-api.
 			if (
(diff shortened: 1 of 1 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -1666,21 +1668,18 @@ export const addPermissions = async (
...
+	let permissionsLookup;
 	if (
-		permissions.length > 0 &&
-		$checkPermissions(
-			getPermissionsLookup(permissions),
-			permissionType,
-			vocabulary,
-		) === true
+		guestPermissionsInitialized === false &&
+		(req.user === root.user || req.user === rootRead.user)
```
thgreasi, line 1674: How about instead of this we just change the `getReqPermissions` to never fetch the guest permissions for the root users, (regardless of `guestPermissionsInitialized`)? If we are fine not having the guest permissions once for root users, then I guess we might be fine not having them at all consistently.

src/sbvr-api/permissions.ts
```diff
@@ -1666,21 +1668,18 @@ export const addPermissions = async (
 		}
 	}
 
-	// This bypasses in the root cases, needed for fetching guest permissions to work, it can almost certainly be done better though
-	let permissions = req.user?.permissions ?? [];
-	permissions = permissions.concat(req.apiKey?.permissions ?? []);
+	let permissionsLookup;
```
thgreasi, line 1671: NIT: I think this atm is `any` .

Page-: Yeah, I'm not sure why that doesn't error on the typings though (since we have `strict: true` which defaults `noImplicitAny: true`)? Either way I've explicitly typed it now

src/sbvr-api/permissions.ts
Page-: I think you're right that moving it into `getReqPermissions` makes sense to move it closer to the source, however I think it makes sense to limit the special case as much as possible to avoid on-going differences between eg
```
	passthrough: {
		req: rootRead,
	},
```
and
```
	passthrough: {
		req: {
			user: {
				id: 0,
				actor: 0,
				permissions: ['resource.read'],
			},
		},
	},
```
Where the former would not include guest permissions but the latter would despite otherwise being identical

src/sbvr-api/permissions.ts
```diff
@@ -1611,7 +1613,18 @@ const getReqPermissions = async (
 	odataBinds: ODataBinds = [],
 ) => {
 	const [guestPermissions] = await Promise.all([
-		getGuestPermissions(),
+		(async () => {
+			if (
+				guestPermissionsInitialized === false &&
```
thgreasi, line 1618: For me it worked even without this.
```suggestion
```
Any reason we prefer having it for root users at all??
W/o it this code path will always be consistent on every run.

thgreasi (commented):
LGTM :+1:

src/sbvr-api/permissions.ts
Page-: I did comment on it in https://github.com/balena-io/pinejs/pull/559#discussion_r926744545#559 
>I think it makes sense to limit the special case as much as possible to avoid on-going differences between eg
>```
>	passthrough: {
>		req: rootRead,
>	},
>```
>and
>```
>	passthrough: {
>		req: {
>			user: {
>				id: 0,
>				actor: 0,
>				permissions: ['resource.read'],
>			},
>		},
>	},
>```
>Where the former would not include guest permissions but the latter would despite otherwise being identical

thgreasi (approved):
:clap:

PR #560, opened 2022-07-21
Increase strictness of typings

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

fisehara:
@Page- I don't agree on `err:any` as typescript also documents here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#unknown-on-catch-clause-bindings that `err:unknown` should be preferred used over `err:any`.

The implementation doesn't know the type and would need to test for instanceOf Error to do de-referencing like `e.message`

Page-:
@fisehara the reason for the change is so that we can set the default for catch clauses to `unknown` in the tsconfig.json for future cases whilst keeping the current code the same (but highlighting the use of `any` so we can fix it over time).
In short the behavior of the catch clauses is unchanged, all I've done is expose that behavior

PR #561, opened 2022-07-25
Convert lodash `isObject` usage to native versions

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #562, opened 2022-07-25
Convert lodash flatMap/flatten to native versions

Page-:
Change-type: patch


package.json
```diff
@@ -116,7 +116,7 @@
     "serve-static": "^1.15.0"
   },
   "engines": {
-    "node": ">=10.0.0",
+    "node": ">=12.0.0",
```
Page-, line 119: Between the fact that we've been relying on nodejs 12 types since 13.0.0 (in 2020), have not included nodejs 10 in the test matrix for a long time now, and nodejs 10 has also been EOL for ages I think this makes sense as it clarifies what is actually supported (and even nodejs 12 is itself EOL)

Page-:
@balena-ci I self-certify!

PR #563, opened 2022-07-26
Prefer for..of over forEach

Page-:
Change-type: patch

### DIFF
--- a/src/odata-metadata/odata-metadata-generator.ts
+++ b/src/odata-metadata/odata-metadata-generator.ts
@@ -21,17 +21,14 @@ const forEachUniqueTable = <T>(
 	const usedTableNames: { [tableName: string]: true } = {};
 
 	const result = [];
-	for (const key in model) {
-		if (model.hasOwnProperty(key)) {
-			const table = model[key];
-			if (
-				typeof table !== 'string' &&
-				!table.primitive &&
-				!usedTableNames[table.name]
-			) {
-				usedTableNames[table.name] = true;
-				result.push(callback(key, table));
-			}
+	for (const [key, table] of Object.entries(model)) {
+		if (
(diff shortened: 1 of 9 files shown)


src/odata-metadata/odata-metadata-generator.ts
```diff
@@ -21,17 +21,14 @@ const forEachUniqueTable = <T>(
...
-			const table = model[key];
-			if (
-				typeof table !== 'string' &&
-				!table.primitive &&
-				!usedTableNames[table.name]
-			) {
-				usedTableNames[table.name] = true;
-				result.push(callback(key, table));
-			}
+	for (const [key, table] of Object.entries(model)) {
```
thgreasi, line 24: :+1:

src/sbvr-api/uri-parser.ts
```diff
@@ -96,12 +96,12 @@ export const memoizedParseOdata = (() => {
 				// Try to cache based on parameter aliases if there might be some
 				const parameterAliases = new URLSearchParams();
 				const queryParams = new URLSearchParams(url.slice(queryParamsIndex));
-				Array.from(queryParams.entries()).forEach(([key, value]) => {
+				for (const [key, value] of Array.from(queryParams.entries())) {
```
thgreasi, line 99:
```suggestion
				for (const [key, value] of queryParams.entries()) {
```

thgreasi (commented):
:+1:

src/sbvr-api/uri-parser.ts
Page-: I think that's unsafe to do because we use `queryParams.delete(key);` below and that'll affect the iterator if we haven't already pulled it all into an array

thgreasi (approved):

PR #565, opened 2022-07-28
Differentiate typings between parsed and processed odata requests

Page-:
This allows us to make properties on the process odata request required
whereas previously they had to be optional in order to be compatible,
as of now this only affects the `engine` property but in future further
properties of the processed request could be made required. And even at
this point it will block passing a parsed only request object to a
function expecting a processed request object

Change-type: patch

### DIFF
--- a/src/sbvr-api/abstract-sql.ts
+++ b/src/sbvr-api/abstract-sql.ts
@@ -173,7 +173,10 @@ const checkModifiedFields = (
 };
 export const isRuleAffected = (
 	rule: AbstractSQLCompiler.SqlRule,
-	request?: ODataRequest,
+	request?: Pick<
+		ODataRequest,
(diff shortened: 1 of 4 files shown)


fisehara (commented):
LGTM 

Is there a chance to get a test case for: 
> ... even at this point it will block passing a parsed only request object to a
function expecting a processed request object

Page-:
It'll block at the typings level which makes it much harder to test (there are ways to do tests on typings but they're quite fiddly)

Page-:
@balena-ci I self-certify!

fisehara:
> It'll block at the typings level which makes it much harder to test (there are ways to do tests on typings but they're quite fiddly)

@Page- In the pinejs test framework, the test case could test for not starting up a particular pinejs configuration that violates the typing. This is what I'm going to do to check the async migration mutual exclusive typings.

PR #571, opened 2022-09-08
Remove prepublishOnly lint step

Page-:
It is already run as part of the tests which are required by CI before it'll publish, and avoids issues when publishing without dev deps

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #574, opened 2022-09-08
Remove docker hub credentials from flow zone job

fisehara:
Change-type: patch


fisehara:
@balena-ci I self-certify!

PR #576, opened 2022-09-09
Add headers data-type to HookReq

nitishagar:
Change-Type: patch

See:
- https://github.com/balena-io/balena-api/pull/3773

### DIFF
--- a/src/sbvr-api/hooks.ts
+++ b/src/sbvr-api/hooks.ts
@@ -24,6 +24,7 @@ export interface HookReq {
 	method: string;
 	url: string;
 	query: AnyObject;
+	headers: AnyObject;
 	params: AnyObject;
 	body: AnyObject;
 	custom?: AnyObject;


src/sbvr-api/hooks.ts
```diff
@@ -24,6 +24,7 @@ export interface HookReq {
 	method: string;
 	url: string;
 	query: AnyObject;
+	headers: AnyObject;
```
Page-, line 27: We don't add any headers to the `HookReq` do we so what's the reasoning for this change?

nitishagar: We are looking to add file-upload via hook. Currently, I am trying out `busboy` to implement this which needs `headers` info to identify `Content-Type` See: https://github.com/mscdex/busboy/blob/master/lib/index.js#L12

Implementation on `balena-api` side: https://github.com/balena-io/balena-api/pull/3773/files#diff-6147e3c1761df71fd40952dbcbac388a45f80b8c318f367ef41048351a2c0c99R81

@Page- Let me know your thoughts.

PR #581, opened 2022-11-01
Docs: Fix TypeScript getting started example

joshbwlng:
Change-type: patch
Signed-off-by: Josh Bowling <josh@monarci.com>


joshbwlng:
@balena-ci I self certify!

PR #586, opened 2022-11-15
Update TypeScript to 4.9.3

thgreasi:
Change-type: patch


thgreasi:
@balena-ci I self-certify!

PR #590, opened 2022-11-16
Support `$count: { $filter: {...} }` notation in $filter & $orderby

thgreasi:
Update pinejs-client-core from 6.10.2 to 6.12.2

Change-type: minor


thgreasi:
@balena-ci I self-certify!

PR #594, opened 2022-11-20
Update @balena/abstract-sql-compiler from 7.20.0 to 7.21.0

joshbwlng:
Change-type: patch
Signed-off-by: Josh Bowling <josh@monarci.com>


joshbwlng:
@balena-ci I self certify!

PR #595, opened 2022-11-22
Use minimized aliases when not in debug mode for smaller queries

Page-:
Update @balena/odata-to-abstract-sql from 5.5.0 to 5.6.0

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #601, opened 2022-11-25
Use `affectedIds` to narrow the number of rows checked by rules

Ereski:
Change-type: minor
Signed-off-by: Carol Schulze <carol@balena.io>

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,10 +24,10 @@
     "prettify": "balena-lint -e js -e ts --fix src build typings Gruntfile.ts"
   },
   "dependencies": {
-    "@balena/abstract-sql-compiler": "^7.26.0",
+    "@balena/abstract-sql-compiler": "^8.0.0",
     "@balena/abstract-sql-to-typescript": "^1.4.2",
     "@balena/env-parsing": "^1.1.5",
-    "@balena/lf-to-abstract-sql": "^4.7.0",
+    "@balena/lf-to-abstract-sql": "^5.0.0",
     "@balena/odata-parser": "^2.4.2",
     "@balena/odata-to-abstract-sql": "^5.9.2",
     "@balena/sbvr-parser": "^1.4.3",
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -372,12 +372,60 @@ export const isModelNew = async (
 	return !cachedIsModelNew.has(modelName);
 };
 
+const bindsForAffectedIds = (
+	bindings: AbstractSQLCompiler.Binding[],
+	request?: Pick<
+		uriParser.ODataRequest,
+		| 'vocabulary'
+		| 'abstractSqlModel'
+		| 'method'
+		| 'resourceName'
+		| 'affectedIds'
+	>,
+) => {
+	if (request?.affectedIds == null) {
+		return {};
+	}
+
+	const tableName =
+		getAbstractSqlModel(request).tables[resolveSynonym(request)].name;
+
+	// If we're deleting the affected IDs then we can't narrow our rule to
+	// those IDs that are now missing
+	const isDelete = request.method === 'DELETE';
+
+	const odataBinds: { [key: string]: any } = {};
+	for (const bind of bindings) {
+		if (
+			bind.length !== 2 ||
+			bind[0] !== 'Bind' ||
+			typeof bind[1] !== 'string'
+		) {
+			continue;
+		}
+
+		const bindName = bind[1];
+		if (!isDelete && bindName === tableName) {
+			odataBinds[bindName] = ['Text', `{${request.affectedIds}}`];
+		} else {
+			odataBinds[bindName] = ['Text', '{}'];
+		}
+	}
+
+	return odataBinds;
+};
+
 export const validateModel = async (
 	tx: Db.Tx,
 	modelName: string,
 	request?: Pick<
 		uriParser.ODataRequest,
-		'abstractSqlQuery' | 'modifiedFields' | 'method' | 'vocabulary'
+		| 'abstractSqlQuery'
+		| 'modifiedFields'
+		| 'method'
+		| 'vocabulary'
+		| 'resourceName'
+		| 'affectedIds'
 	>,
 ): Promise<void> => {
 	const { sql } = models[modelName];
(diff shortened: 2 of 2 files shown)


src/sbvr-api/sbvr-utils.ts
```diff
@@ -394,7 +442,7 @@ export const validateModel = async (
 			const values = await getAndCheckBindValues(
 				{
 					vocabulary: modelName,
-					odataBinds: [],
+					odataBinds: bindsForAffectedIds(rule.bindings, request) as any,
```
Page-, line 445: If there needs to be a cast I'd rather it be in the `bindsForAffectedIds` function directly

src/sbvr-api/sbvr-utils.ts
```diff
@@ -372,12 +372,60 @@ export const isModelNew = async (
...
+	bindings: AbstractSQLCompiler.Binding[],
+	request?: Pick<
+		uriParser.ODataRequest,
+		| 'vocabulary'
+		| 'abstractSqlModel'
+		| 'method'
+		| 'resourceName'
+		| 'affectedIds'
+	>,
+) => {
```
Page-, line 385:
```suggestion
): ODataBinds => {
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -372,12 +372,60 @@ export const isModelNew = async (
...
+	if (!request || !request.affectedIds) {
```
Page-, line 386: I assume this would work:
```suggestion
	if (request?.affectedIds == null) {
```
?

package.json
```diff
@@ -87,11 +87,11 @@
     "mocha": "^10.2.0",
     "raw-loader": "^4.0.2",
     "require-npm4-to-publish": "^1.0.0",
-    "supertest": "^6.3.3",
+    "supertest": "^6.3.1",
```
Page-, line 90: Can you revert these downgrades please

src/sbvr-api/sbvr-utils.ts
Ereski: @Page- `ODataBinds` doesn't actually cover the case needed for this particular function, which is map and not an array: https://github.com/balena-io/pinejs/pull/601/files#diff-824816fb30b8d1b6c2ec737540770325de0f51a1d9d8fca9e44bedcae76007d0R397. Even though this form is supported just fine: https://github.com/balena-io/pinejs/blob/master/src/sbvr-api/abstract-sql.ts#L117-L128.

IIRC the problem here is that `bindsForAffectedIds` can't know which index are the ones used by the narrowing snippet so that ID is bound by name as supported by the asql compiler and pine.

`ODataBinds` is defined by `odata-parser` so maybe it's worth expanding the type definition there first?

src/sbvr-api/sbvr-utils.ts
Ereski: It does, though I'm puzzled as to why `===` leads to type errors

src/sbvr-api/sbvr-utils.ts
Page-: @Ereski `ODataBinds` is defined as
```ts
export interface ODataBinds extends Array<any> {
	[key: string]: any;
}
```
which is both an array and a dictionary object so it should be fine

Ereski: @Page- It's not because it requires the usual symbols present in an array `extends Array`. Can't take an existing array and add extra entries either because of [isArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray). Would have to be something closer to:

```ts
export type ODataBinds =
  Array<any>
  | {
    [key: string]: any;
  }
```

But what would the ramifications be?

Page-: @Ereski it's specifically both an array and an object - where is the `isArray` check that's causing you problems if you go that route?

Ereski: @Page- It's an array with arbitrary extra fields. But in this case in particular it should be *either* an array or an object, not both. I'm going to explain the whole logic here to show the typing issue better and as a sanity check to myself too.

In `getAndCheckBindValues`, values to be bound are passed through `request.odataBinds`, which has type `OdataBinds` and is what you posted:

```ts
export interface ODataBinds extends Array<any> {
	[key: string]: any;
}
```

`getAndCheckBindValues` retrieves the actual value for each binding in the rule's SQL from `request.odataBinds`, and those values are used when running the rule: https://github.com/balena-io/pinejs/blob/7352cf516f188b21485647fb944faa649ee9c081/src/sbvr-api/sbvr-utils.ts#L442-L451. It additionally checks if those values are valid.

`getAndCheckBindValues` accepts multiple shapes for `odataBinds` because there is more than one way to name a binding: https://github.com/balena-io-modules/abstract-sql-compiler/blob/master/src/AbstractSQLCompiler.ts#L141. Where

- `['Bind', string, string]` defines a reference to a qualified table.column if I'm reading this correctly: https://github.com/balena-io/pinejs/blob/master/src/sbvr-api/abstract-sql.ts#L84-L105
- `['Bind', number]` defines a positional binding and `request.odataBinds` must be an array: https://github.com/balena-io/pinejs/blob/master/src/sbvr-api/abstract-sql.ts#L106-L116
- `['Bind', string]` defines a named binding and `request.odataBinds` must be an object: https://github.com/balena-io/pinejs/blob/master/src/sbvr-api/abstract-sql.ts#L117-L127

I made an error in my previous comment and `isArray` doesn't matter here, but the problem is the same: an array is not an arbitrary object.

In the case of narrowing, bindings fall in the last category above, using named bindings. The name here being the name of the table which that binding narrows. In this case, `request.odataBinds` *must not* be an array for one reason: an `Array<any>` implies predefined symbols with predefined behaviors and predefined types. If, by coincidence, the name of a table clashes (and possibly overwrites) one of those symbols then the `Array` contract is broken.

In this particular setup that's not a fatal problem for almost all cases because all `getAndCheckBindValues` does is check the array's size and index into it. But it would still be a bug and a potential problem in the future. The only exception AFAIK would be if the name of a table was `length`, which would cause an exception.

Having `odataBinds` be a pure map avoids that `length` and similar errors, and redefining `ODataBinds` to be either an array or a map (instead of a possibly mangled array) resolves everything given the appropriate type checks.

Page- (approved):

PR #602, opened 2022-11-29
Add config for multer with constraints for size and routes

nitishagar:
Change-Type: minor


thgreasi (commented):
Should definitely also be a minor since it's adding a feature

PR #603, opened 2022-12-01
Update eventemitter to 5.x

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #604, opened 2022-12-01
Proposal: v15

Page-:


fisehara:
@Page- would it make sense to switch over to odata v4 result.

`{ d: [] }` => `{value: []}`

src/migrator/utils.ts
```diff
@@ -12,6 +12,23 @@ import { delay } from '../sbvr-api/control-flow';
 
 // tslint:disable-next-line:no-var-requires
 export const modelText = require('./migrations.sbvr');
+export const migrations: Migrations = {
```
Page-, line 15: TODO: migrate `is backing off` boolean as well

Page-:
> would it make sense to switch over to odata v4 result.

In general yes, although I expect there's probably more required than just switching `d` to `value` to make it actually match OData v4 so I'd rather do it as a concerted effort rather than making it at first glance look like it'll match and then throw people off later. For reference it was following a (presumably draft) version of the spec that there's essentially no references left for but is at least mentioned a bit in https://www.odata.org/blog/format-efficiency-take-2-really-clean-json/ and https://www.odata.org/blog/json-light-sample-payloads/

Page-:
> Are we going to drop backwards node12 compatibility and jump to node16 engine requirement directly?

that's the plan

src/sbvr-api/permissions.ts
```diff
@@ -1477,11 +1462,6 @@ export const resolveAuthHeader = async (
 	// TODO: Consider making tx the second argument in the next major
```
Page-, line 1462: @thgreasi ?

src/sbvr-api/permissions.ts
```diff
@@ -1527,11 +1507,6 @@ export const resolveApiKey = async (
...
```
Page-, line 1507: @thgreasi ?

src/sbvr-api/permissions.ts
```diff
@@ -862,9 +862,7 @@ const rewriteRelationship = memoizeWeak(
 								foundCanAccessLink = true;
 							}
 							// return a true expression to not select the relationship, which might be virtual
-							// this should be a boolean expression, but needs to be a subquery in case it
-							// is wrapped in an `or` or `and`
-							return ['Equals', ['Boolean', true], ['Boolean', true]];
+							return ['Boolean', true];
```
thgreasi, line 865: Just curious, which changes was that allowed this?

src/migrator/utils.ts
```diff
@@ -425,7 +439,8 @@ LIMIT 1;`,
 			migrated_row_count: data['migrated row count'],
 			error_count: data['error count'],
 			converged_time: data['converged time'],
-			is_backing_off: data['is backing off'] === 1,
+			is_backing_off:
+				data['is backing off'] === true || data['is backing off'] === 1,
```
thgreasi, line 443: Afaict that's just for being able to do this major bump right?
Should this and the above `typeof data.executed_migrations === 'string'` have a `TODO` to drop it on a next major?

src/migrator/utils.ts
```diff
@@ -12,6 +12,29 @@ import { delay } from '../sbvr-api/control-flow';
...
+					ALTER TABLE "migration status"
+					MODIFY "is backing off" JSON NOT NULL;`);
+				break;
+			case 'postgres':
+				await tx.executeSql(`\
+					ALTER TABLE "migration"
+					ALTER COLUMN "executed migrations" SET DATA TYPE JSONB USING b::JSONB;`);
+				await tx.executeSql(`\
+					ALTER TABLE "migration status"
+					ALTER COLUMN "is backing off" SET DATA TYPE JSONB USING b::JSONB;`);
```
thgreasi, line 32: Shouldn't this be BOOLEAN?

src/migrator/utils.ts
```diff
@@ -12,6 +12,29 @@ import { delay } from '../sbvr-api/control-flow';
...
+export const migrations: Migrations = {
+	'15.0.0-data-types': async (tx, { db }) => {
+		switch (db.engine) {
+			case 'mysql':
+				await tx.executeSql(`\
+					ALTER TABLE "migration"
+					MODIFY "executed migrations" JSON NOT NULL;`);
+				await tx.executeSql(`\
+					ALTER TABLE "migration status"
+					MODIFY "is backing off" JSON NOT NULL;`);
```
thgreasi, line 24: Shouldn't this be BOOLEAN?

src/http-transactions/transactions.js
```diff
@@ -24,6 +25,23 @@ ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT N
...
+						case 'mysql':
+							await tx.executeSql(`\
+								ALTER TABLE "lock"
+								MODIFY "is exclusive" BOOLEAN NOT NULL,
+								MODIFY "is under lock" BOOLEAN NOT NULL;`);
+							break;
+						case 'postgres':
+							await tx.executeSql(`\
+								ALTER TABLE "lock"
+								ALTER COLUMN "is exclusive" SET DATA TYPE BOOLEAN USING b::BOOLEAN,
```
thgreasi, line 39: What is this `b` in all PG migrations that have a `USING`? Should it be `"is exclusive"`?

src/sbvr-api/permissions.ts
thgreasi: @Page- that's just imo that `req, tx, paramName` makes more sense, since both req & tx provide context, while `param` is the real arguments. I'm fine with it as is though.

src/sbvr-api/permissions.ts
thgreasi: Same as below
> that's just imo that `req, tx, paramName` makes more sense, since both req & tx provide context, while `param` is the real arguments. I'm fine with it as is though.

src/sbvr-api/permissions.ts
```diff
@@ -1477,11 +1462,6 @@ export const resolveAuthHeader = async (
 	// TODO: Consider making tx the second argument in the next major
 	tx?: Tx,
 ): Promise<PermissionReq['apiKey']> => {
-	// TODO-MAJOR: remove this check
```
thgreasi, line 1480: I'm fine with it, but really curious about the reasoning :thinking: 
Was there a case that mid-request we decided we want to re-evaluate the cached value?

src/sbvr-api/permissions.ts
Page-: It's allowed by using true booleans, ie `['Boolean', true]` will now generate `TRUE` rather than `1` which means `WHERE TRUE` is valid vs `WHERE 1` which is not

src/http-transactions/transactions.js
Page-: Fixed

src/migrator/utils.ts
Page-: Fixed

src/sbvr-api/permissions.ts
Page-: This function (and the ones it calls) do not assign `req.apiKey` at all so it's weird for them to look at it's value and potentially refuse to check a different api key that has been passed because some external property has been set. In practice it works out because of how we have always used it thus far but it doesn't make sense in itself - it makes more sense for whatever assigns `req.apiKey` to check for it already existing if that is important

src/migrator/utils.ts
Page-: It's also necessary for sqlite/websql where BOOLEAN is just an alias for INTEGER

thgreasi (approved):

package.json
```diff
@@ -113,8 +113,8 @@
     "serve-static": "^1.15.0"
   },
   "engines": {
-    "node": ">=12.0.0",
-    "npm": ">=6.0.0"
+    "node": ">=16.13.0",
```
fisehara, line 116: We should directly bump it to node 18

fisehara (commented):
LGTM

PR #608, opened 2022-12-08
Update @balena/lf-to-abstract-sql to 4.6.0

Page-:
Update @balena/lf-to-abstract-sql from 4.5.6 to 4.6.0

Change-type: patch


Page-:
@balena-ci I self-certify!

fisehara:
Update @balena/lf-to-abstract-sql from 4.5.6 to 4.6.0 was already performed

PR #610, opened 2022-12-08
Fix async migrations being able to stop unexpectedly in some failure cases

Page-:
Change-type: patch


fisehara:
@Page- good to merge nad proceed?

fisehara (approved):

PR #615, opened 2022-12-13
Record async migration execution time

fisehara:
Needs a migration of existing `migration status` tables to introduce `execution time` column.

Change-type: minor


test/03-async-migrator.test.ts
```diff
@@ -202,6 +202,29 @@ describe('03 Async Migrations', async function () {
 			expect(result[0]?.migrated_row_count - firstRowsMigrated).to.equal(1);
 			expect(Date.now().valueOf() - startTime).to.be.greaterThan(4000); // backOff time from migrator
 		});
+
+		it.only('should record last execution time ms of the last migration run', async function () {
```
Page-, line 206:
```suggestion
		it('should record last execution time ms of the last migration run', async function () {
```

src/migrator/sync.ts
```diff
@@ -146,6 +159,10 @@ export const config: Config = {
 					ALTER TABLE "migration lock"
 					ADD COLUMN IF NOT EXISTS "modified at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
 				`,
+				'14.55.0-last-execution-time': `
+					ALTER TABLE "migration status"
+					ADD COLUMN IF NOT EXISTS "last execution time ms" INTEGER NULL;
```
Page-, line 164: This would also need to drop the `start time` column, and that potentially makes it a breaking change

fisehara: @Page- 
I was unsure how to handle it. Basically PineJS doesn't know about old instances that reference the `start time`. My non favourite solution is to keep the column as is. My preferred solution would be a 2 staged deployment. 

Basically it's not breaking the code it's breaking old-still running instances which require `start time` to be existing.
How was this handled in the past?

PR #616, opened 2022-12-15
Utilize @balena/env-parsing module

fisehara:
Change-type: patch


src/database-layer/db.ts
```diff
@@ -469,7 +470,7 @@ if (maybePg != null) {
 		const PG_CHECK_CONSTRAINT_VIOLATION = '23514';
 		const PG_EXCLUSION_CONSTRAINT_VIOLATION = '23P01';
 
-		const { PG_SCHEMA } = process.env;
+		const PG_SCHEMA = optionalVar('PG_SCHEMA', null);
```
Page-, line 473:
```suggestion
		const PG_SCHEMA = optionalVar('PG_SCHEMA', undefined);
```

src/server-glue/module.ts
```diff
@@ -57,7 +58,7 @@ export const init = async <T extends string>(
 		await cfgLoader.loadConfig(migrator.config);
 
 		const promises: Array<Promise<void>> = [];
-		if (process.env.SBVR_SERVER_ENABLED) {
+		if (optionalVar('SBVR_SERVER_ENABLED')) {
```
Page-, line 61: I think these changes will interfere with the DEFINE plugin: https://github.com/balena-io/pinejs/blob/v14.54.1/build/browser.ts#L18-L24

fisehara: Oh I see...

https://webpack.js.org/plugins/define-plugin/: 
> Note that because the plugin does a direct text replacement, the value given to it must include actual quotes inside of the string itself.

Thus, it's fine for not text-replaced process.env vars... :(

PR #617, opened 2022-12-15
Introduce control means for async migration execution

fisehara:
Introduce env var PINEJS_ASYNC_MIGRATION_ENABLED to control async migration execution on startup. Introduce SIGUSR2 listener to toggle migration execution during runtime.

Change-type: minor

### DIFF
--- a/src/migrator/async.ts
+++ b/src/migrator/async.ts
@@ -27,6 +27,14 @@ import {
 	MigrationStatus,
 	BaseAsyncMigration,
 } from './utils';
+import { booleanToEnabledString } from '../config-loader/env';
+
+// log the startup condition of the async migration
+(sbvrUtils.api?.migrations?.logger.info ?? console.info)(
+	`Async migration execution is ${booleanToEnabledString(
+		migratorEnv.asyncMigrationIsEnabled,
+	)}`,
+);
 
 export const run = async (tx: Tx, model: ApiRootModel): Promise<void> => {
 	const { migrations } = model;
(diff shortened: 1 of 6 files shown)


src/migrator/async.ts
```diff
@@ -28,6 +29,33 @@ import {
 	BaseAsyncMigration,
 } from './utils';
 
+const ASYNC_MIGRATION_ENV_VAR_NAME = 'PINEJS_ASYNC_MIGRATION_ENABLED';
```
Page-, line 32: Can you stick this env var related stuff in https://github.com/balena-io/pinejs/blob/v14.54.1/src/config-loader/env.ts please

src/migrator/async.ts
```diff
@@ -28,6 +29,33 @@ import {
...
+ * When the process receives a SIGUSR2 signal the async migrations will toggle to be enabled or disabled.
+ */
+process.on('SIGUSR2', () => {
+	(sbvrUtils.api.migrations?.logger.info ?? console.info)(
+		`Received SIGUSR2 to toggle async migration execution from ${booleanToEnabledString(
+			asyncMigrationIsEnabled,
+		)} to ${booleanToEnabledString(!asyncMigrationIsEnabled)}`,
+	);
+	asyncMigrationIsEnabled = !asyncMigrationIsEnabled;
+});
```
Page-, line 57: Can you instead export the ability to toggle this so that users of pinejs can choose whatever toggle mechanism they want, hardcoding SIGUSR2 into pinejs would stop any users of pinejs using that for other purposes

src/migrator/async.ts
fisehara: Yes

src/migrator/async.ts
fisehara: @Page- 
Okay so you mean that it's exported and the pinejs user will call this with a signal string 'SIGWHATEVER' or would never call it and then it's not listening to a signal at all.

fisehara: As the migrator configuration is already exposed through pine.env.migrator I'll follow this and let the user of pinejs toggle the 
pine.env.migrator. asyncMigrationIsEnabled on their behalf. If they use a SIGUSR2 or whatever logic to toggle it.
WDYT @Page-

Page- (approved):

fisehara:
I self-certify!

PR #618, opened 2022-12-19
Simplify `checkApiKey` by moving checks out to more relevant places

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #620, opened 2022-12-22
Fix pre-commit husky hook

fisehara:
pre-commit file has no executable flag.

Change-type: patch


fisehara:
@balena-ci I self-certify!

PR #621, opened 2022-12-23
Use postgres advisory locks for migration locking.

fisehara:
Avoid starvation of sync migrations while
aggressive async migrations take the model lock
over and over again.

Change-type: minor

### DIFF
--- a/src/database-layer/db.ts
+++ b/src/database-layer/db.ts
@@ -116,8 +116,11 @@ const validateTransactionLockParameter = (
 	value: number,
 	parameterName: string,
 ) => {
-	// PG also support negative values, but we initially restrict this to non-negative only.
-	if (!Number.isInteger(value) || value < 0 || types.integer.max < value) {
+	if (
+		!Number.isInteger(value) ||
+		value < types.integer.min ||
+		types.integer.max < value
+	) {
 		throw new TypeError(
 			`Invalid parameter '${parameterName}' provided for transaction lock`,
 		);
@@ -125,6 +128,13 @@ const validateTransactionLockParameter = (
 };
 
 const transactionLockNamespaceMap: Dictionary<number> = {};
+
+/**
+ *
+ * @param namespaceKey 	Any string representing a namespace
+ * @param namespaceId 	For application level locks positive values should be used
+ * 						as internal / low level locks are recommended to use negative values
+ */
 
 export function registerTransactionLockNamespace(
 	namespaceKey: string,
@@ -390,7 +400,8 @@ export abstract class Tx {
 	public async getTxLevelLock(
 		_namespaceKey: string,
 		_key: number,
-	): Promise<void> {
+		_blocking: boolean,
+	): Promise<boolean> {
 		throw new Error(
 			'The getTxLevelLock method is not implemented for the current engine.',
 		);
@@ -619,7 +630,11 @@ if (maybePg != null) {
 				}
 			}
 
-			public override async getTxLevelLock(namespaceKey: string, key: number) {
+			public override async getTxLevelLock(
+				namespaceKey: string,
+				key: number,
+				blocking: boolean = true,
+			) {
 				validateTransactionLockParameter(key, 'key');
 				const namespaceId = transactionLockNamespaceMap[namespaceKey];
 				if (namespaceId == null) {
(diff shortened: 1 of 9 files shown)


src/migrator/utils.ts
```diff
@@ -215,6 +226,73 @@ WHERE "model name" = ${1}`,
...
+		// pg_advisory lock expects an BigInt or two Int as lock identifier
+		// Therefore the model name is hashed and the first 8 bytes are taken as the Integer representation.
+		const modelKey: string = createHash('shake128', { outputLength: 8 })
+			.update('resin')
+			.digest()
+			.readBigInt64BE()
+			.toString();
+		getLockStatement = blockingWait
+			? `pg_advisory_xact_lock(${modelKey})`
+			: `pg_try_advisory_xact_lock(${modelKey})`;
```
Page-, line 257: Can you use `tx.getTxLevelLock(...)` instead please, and extend it to allow for non-blocking operation as well

src/migrator/sync.ts
```diff
@@ -30,7 +30,7 @@ export const postRun = async (tx: Tx, model: ApiRootModel): Promise<void> => {
 			`First time executing '${modelName}', running init script`,
 		);
 
-		await lockMigrations(tx, modelName, async () => {
+		await lockMigrations(tx, modelName, true, async () => {
```
Page-, line 33: I think this is reaching the point where it probably makes sense to use an options object in order to have named parameters, ie
```suggestion
		await lockMigrations({tx, modelName, blocking: true}, async () => {
```

src/migrator/utils.ts
fisehara: Oh sure, thanks for the pointer

src/migrator/sync.ts
fisehara: Thanks!

src/database-layer/db.ts
```diff
@@ -619,18 +630,37 @@ if (maybePg != null) {
...
-					namespaceId,
-					key,
-				]);
+				try {
+					if (blocking) {
+						const result = await this.executeSql(
+							`SELECT pg_advisory_xact_lock($1, $2);`,
+							[namespaceId, key],
+						);
+						return result.rows?.[0]?.['pg_advisory_xact_lock'] === '';
```
Page-, line 651: If the statement returns at all that means it has to be successful right? And since the return type of `pg_advisory_xact_lock` is `void` we shouldn't really be looking at the contents
```suggestion
						return true;
```

src/database-layer/db.ts
```diff
@@ -619,18 +630,37 @@ if (maybePg != null) {
...
+					} else {
+						const result = await this.executeSql(
+							`SELECT pg_try_advisory_xact_lock($1, $2);`,
+							[namespaceId, key],
+						);
+						return result.rows?.[0]?.['pg_try_advisory_xact_lock'] === true;
```
Page-, line 657: I'm sure most (all?) of these `?` can be removed:
```suggestion
						return result.rows[0].pg_try_advisory_xact_lock === true;
```

src/database-layer/db.ts
fisehara: Fair point, will do so.

src/database-layer/db.ts
fisehara: Thx!

src/database-layer/db.ts
```diff
@@ -619,18 +630,37 @@ if (maybePg != null) {
...
+					} else {
+						const result = await this.executeSql(
+							`SELECT pg_try_advisory_xact_lock($1, $2);`,
+							[namespaceId, key],
+						);
+						return result.rows[0].pg_try_advisory_xact_lock === true;
+					}
+				} catch (error) {
+					throw new Error(
+						`getTxLevelLock error during getting lock from postgres db layer ${error}`,
```
Page-, line 661: It's minor but we almost always use `err` for these and so the full form of `error` feels kind of out of place
```suggestion
				} catch (err) {
					throw new Error(
						`getTxLevelLock error during getting lock from postgres db layer ${err}`,
```

src/database-layer/db.ts
```diff
@@ -619,18 +630,37 @@ if (maybePg != null) {
...
+							namespaceId,
+							key,
+						]);
+						return true;
+					} else {
+						const result = await this.executeSql(
+							`SELECT pg_try_advisory_xact_lock($1, $2);`,
+							[namespaceId, key],
+						);
+						return result.rows[0].pg_try_advisory_xact_lock === true;
```
Page-, line 657: Minor but slightly shortens things
```suggestion
						const { rows } = await this.executeSql(
							`SELECT pg_try_advisory_xact_lock($1, $2);`,
							[namespaceId, key],
						);
						return rows[0].pg_try_advisory_xact_lock === true;
```

Page- (commented):
Looks good, ideally just fix the very minor things and then you can self-cert of ping for an approve=merge

src/database-layer/db.ts
fisehara: Thanks for the reminder!

fisehara:
I self-certify!

PR #624, opened 2023-01-13
Filter for `SideEffectHook`s before entering the async rollback loop

Page-:
This avoids unnecessary work by adding an async boundary for the filtering part, meaning we can avoid any async boundaries in the case of no side effecting hooks (the likely case)

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #628, opened 2023-01-25
db: Make the base getTxLevelLock blocking by default

thgreasi:
That's already how it works b/c the only concrete implementation default to true, so this is more as a way to make the base class typings mark this as an optional parameter having a default value.
Otherwise users would get a  TS error after updating.

Change-type: patch
See: https://github.com/balena-io/open-balena-api/actions/runs/4004587439/jobs/6873921096


fisehara (approved):
Nice spot!

PR #641, opened 2023-02-20
Fixes setting only finalized async migration as executed

fisehara:
Change-type: patch

### DIFF
--- a/src/migrator/async.ts
+++ b/src/migrator/async.ts
@@ -14,7 +14,6 @@ type InitialMigrationStatus = MigrationStatus &
 import {
 	MigrationTuple,
 	AsyncMigrationFn,
-	setExecutedMigrations,
 	getExecutedMigrations,
 	migratorEnv,
 	lockMigrations,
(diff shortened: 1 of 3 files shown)


src/passport-pinejs/passport-pinejs.ts
```diff
@@ -54,7 +54,7 @@ const setup: ConfigLoader.SetupFunction = async (app: Express.Application) => {
 		passport.use(new LocalStrategy(checkPassword));
 
 		login = (fn) => (req, res, next) =>
-			passport.authenticate('local', (err, user?) => {
+			passport.authenticate('local', (err: any, user?: any) => {
```
Page-, line 57: Why this change?

src/migrator/async.ts
```diff
@@ -64,19 +63,10 @@ const $run = async (
...
-			'First time model has executed, skipping async migrations',
-		);
+	const executedMigrations = await getExecutedMigrations(setupTx, modelName);
 
-		return await setExecutedMigrations(setupTx, modelName, [
-			...executedMigrations,
-			...Object.keys(migrations),
-		]);
-	}
+	// if the model is new, the sync migration parts (marked by finalize=true) are already marked in the sync migration runner.
```
Page-, line 69: And we know that it'll definitely mark them as executed before we reach this branch?

fisehara: Yes definitively, the sync migrations are executed guaranteed before the async migrations. Any async migration that is finalized is executed in the sync migration block and will be marked as executed there.

src/passport-pinejs/passport-pinejs.ts
fisehara: TSC complains.

fisehara: I'll remove it from this PR and make another explicitly for this TSC complains

PR #642, opened 2023-02-20
Fisehara/fix implict any type

fisehara:


src/passport-pinejs/passport-pinejs.ts
```diff
@@ -54,7 +54,7 @@ const setup: ConfigLoader.SetupFunction = async (app: Express.Application) => {
 		passport.use(new LocalStrategy(checkPassword));
 
 		login = (fn) => (req, res, next) =>
-			passport.authenticate('local', (err, user?) => {
+			passport.authenticate('local', (err: any, user?: any) => {
```
Page-, line 57: It'd be better to use the correct typing (no idea why tsc doesn't infer it automatically):
```ts
			passport.authenticate('local', ((err, user) => {
				if (err || user == null || user === false) {
					fn(err, user, req, res, next);
					return;
				}
				req.login(user, (error) => {
					fn(error, user, req, res, next);
				});
			}) as Passport.AuthenticateCallback)(req, res, next);
```
And then propogate the `| false | null` to
```ts
export let login: (
	fn: (
		err: any,
		user: {} | false | null | undefined,
		req: Express.Request,
		res: Express.Response,
		next: Express.NextFunction,
	) => void,
) => Express.RequestHandler;

fisehara: Thanks for the guidance!

Page- (approved):

PR #649, opened 2023-03-15
Actually compute computed terms only once.

fisehara:
Afterwards use as Alias or ReferencedFields

Change-type: patch


fisehara:
obsolete will be implemented in odata-to-abstract-sql solely

PR #654, opened 2023-04-05
Add scheduled migrations

joshbwlng:
Change-type: minor

---

WIP

---

Currently working on something like this:
- Add a new `scheduled job` resource to the `migration` model
- Add hooks for creating/managing `scheduled job`s
- Schedule executions using `node-schedule` on creation (we also use this for jobs in `open-balena-api`)
- Also re-schedule existing but not yet executed `scheduled job`s during system setup using `node-schedule`
- Add ability to register job handlers (similar to hooks), each with a unique name
  - `scheduled job` resources must specify the `name` of their handler and `parameters` required by the handler
  - `parameters` is a free-form json object stored in the db, retrieved by pine and passed to the handler on execution
- Add retry/backoff logic for failed executions
  - Should be configurable with sane defaults?
  - Very useful when working with external services that could be down or rate limit requests

Note: Will try breaking away from using existing migration functionality as it now feels forced and unnecessary.


src/sbvr-api/sbvr-utils.ts
```diff
@@ -641,6 +643,22 @@ export const executeModels = async (
 						}
 					}
 				}
+				if (model.modelName === 'migrations') {
+					sbvrUtils.addPureHook('POST', 'migrations', 'scheduled_migration', {
```
joshbwlng, line 647: TODO: Move this to possibly `src/migrator/hooks.ts`.

src/migrator/sync.ts
```diff
@@ -52,6 +54,14 @@ export const run = async (tx: Tx, model: ApiRootModel): Promise<void> => {
 	return $run(tx, model, runnableMigrations);
 };
 
+export const schedule = async (date: Date, key: string, callback: () => Promise<void>): Promise<void> => {
```
joshbwlng, line 57: TODO: `callback` should instead probably be a `Migration` instead of a generic function.

PR #655, opened 2023-04-07
Add scheduled tasks

joshbwlng:
Change-type: minor

---

**[WIP]** Add a new `tasks` vocabulary with `job` as its primary resource. Consumers can first register new job handlers with `addTaskHandler()` (similar to `addPureHook()`), and then schedule tasks for future execution by creating `task` resources normally. Task executions are cancelled upon their resource counterpart being deleted. Limited retry/backoff logic included on execution failure. During startup, any pending jobs are either executed immediately or scheduled for future execution depending on their `start_time` date.

**Note:** Originally started with https://github.com/balena-io/pinejs/pull/654 which uses some existing migration functionality behind the scenes, but decided to try something that was completely decoupled from migrations. I'm looking for something to schedule and execute tasks, not migrations. While we could force the use of migrations, I'm not convinced it's the right decision - don't need async migrations/etc and wouldn't want to pollute migration tables with non-migration task records.

---

ToDo/Issues:
- Scope task handlers to models?
- Allow for retry/backoff value overrides
- Needs queueing of some sort? Worried about what happens when too many tasks are scheduled to execute at the same time.

---

Example usage can be found in `test/07-tasks.test.ts` and `test/fixtures/07-tasks/task-handlers.ts`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "lodash": "^4.17.21",
     "memoizee": "^0.4.15",
     "pinejs-client-core": "^6.13.0",
+    "node-schedule": "^2.1.1",
     "randomstring": "^1.2.3",
     "typed-error": "^3.2.2"
   },
@@ -71,6 +72,7 @@
     "@types/mocha": "^10.0.1",
     "@types/on-finished": "^2.3.1",
     "@types/request": "^2.48.8",
+    "@types/node-schedule": "^2.1.0",
     "@types/supertest": "^2.0.12",
     "@types/terser-webpack-plugin": "^5.2.0",
     "@types/type-is": "^1.6.3",
--- a/src/config-loader/env.ts
+++ b/src/config-loader/env.ts
@@ -146,3 +146,9 @@ export const migrator = {
 	 */
 	asyncMigrationIsEnabled: boolVar('PINEJS_ASYNC_MIGRATION_ENABLED', true),
 };
+
+export const tasks = {
+	pollIntervalMS: process.env.TASKS_POLL_INTERVAL_MS
+		? parseInt(process.env.TASKS_POLL_INTERVAL_MS, 10)
+		: 2000,
+};
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -42,6 +42,7 @@ import { generateODataMetadata } from '../odata-metadata/odata-metadata-generato
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const devModel = require('./dev.sbvr');
+import * as tasks from './tasks';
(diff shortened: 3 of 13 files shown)


fisehara:
@joshbwlng Can we somehow track the actor who caused the job to be scheduled? I think when this is an async data manipulation we still need to track the user for all data changes / transactions that are performed 'later'

joshbwlng:
@fisehara Good point. How about using `req.user.id` in the create hook, something like:
```typescript
export const setup = async (tx: Db.Tx) => {
	// Validate and schedule new scheduled jobs for future execution.
	addPureHook('POST', apiRoot, 'scheduled_job', {
		POSTPARSE: async ({ req, request }) => {
			// Set defaults.
			request.values.is_executed_by__user = req.user.id;
			request.values.status = 'pending';
...
```

fisehara:
@joshbwlng to not limit it to a user calling the API we should go and fetch the actorId like we do for the history tables in balena-api:

```
const getActor = (req: sbvrUtils.HookReq) => {
  const actor = req.user?.actor ?? req.apiKey?.actor;
  if (actor == null) {
	  const err = new Error(
		  `Scheduling job for ${resource} with missing actor on req is not allowed`,
	  );
	  errors.captureException(err, err.message, { req });
	  throw err;
  }
  return actor;
};
```

joshbwlng:
@fisehara Nice, thanks, added a new field to the jobs sbvr and some bits to set the actor using basically the example you provided.

fisehara:
@joshbwlng I have to review it again, but can you please add test cases for the scheduled jobs already? This will demonstrate how they will be used.

src/sbvr-api/jobs.ts
```diff
@@ -0,0 +1,261 @@
...
+ * @returns Actor ID
+ * @throws BadRequestError if actor is missing
+ */
+function getActor(req: sbvrUtils.HookReq): number {
+	const actor = req.user?.actor ?? req.apiKey?.actor;
+	if (actor == null) {
+		throw new errors.BadRequestError(
+			'Scheduling job with missing actor on req is not allowed',
+		);
+	}
```
joshbwlng, line 51: @fisehara This currently always throws here in tests as `req.user` is always undefined. I've tried a number of things including setting `passthrough.req.user` on the pineTest client and `await null;` in the jobs preparse/postparse hooks. Any ideas?

joshbwlng: Nevermind, I did some digging and found out how to create and pass an apikey as part of the request through PineTest (which also works as my hooks checks for the actor from both `req.user` and `req.apiKey`).

joshbwlng:
Moved to https://github.com/balena-io/pinejs/pull/739

PR #656, opened 2023-04-18
Update select sql typing

joshbwlng:
Change-type: patch

---

Recent type changes in `abstract-sql-compiler` broke pinejs typescript builds, updating typings to get builds working again.


fisehara:
LGTM

PR #657, opened 2023-04-19
Fix deepFreeze

fisehara:
Actually deep freezing the `constrainedAbstractSqlModel`

Change-type: patch


src/sbvr-api/permissions.ts
```diff
@@ -715,19 +715,18 @@ const onceGetter = <T, U extends keyof T>(
...
-
 	for (const prop of Object.getOwnPropertyNames(obj)) {
-		// We skip the definition because we know it's a property we've defined that will throw an error in some cases
+		const value = obj[prop];
 		if (
-			prop !== 'definition' &&
-			obj.hasOwnProperty(prop) &&
-			obj[prop] !== null &&
-			!['object', 'function'].includes(typeof obj[prop])
+			// `synonyms` is apparently changed later on and cannot be deepFroozen
```
Page-, line 721: Where is `synonyms` modified?

src/sbvr-api/permissions.ts
```diff
@@ -715,19 +715,18 @@ const onceGetter = <T, U extends keyof T>(
 };
 
 const deepFreezeExceptDefinition = (obj: AnyObject) => {
```
Page-, line 717: `ExceptDefinition` isn't accurate to what this is attempting to do

src/sbvr-api/permissions.ts
```diff
@@ -1036,9 +1035,9 @@ const getBoundConstrainedMemoizer = memoizeWeak(
 
 							const table = tables[`${resourceName}$bypass`];
 
-							const permissionsTable = (tables[permissionResourceName] = {
```
Page-, line 1039: If you don't assign to `tables[permissionResourceName]` then all these changes to `permissionsTable` are essentially noops?

src/sbvr-api/permissions.ts
fisehara: Yep corrected.

src/sbvr-api/permissions.ts
```diff
@@ -714,20 +714,20 @@ const onceGetter = <T, U extends keyof T>(
...
-		// We skip the definition because we know it's a property we've defined that will throw an error in some cases
+		const value = obj[prop];
 		if (
-			prop !== 'definition' &&
-			obj.hasOwnProperty(prop) &&
-			obj[prop] !== null &&
-			!['object', 'function'].includes(typeof obj[prop])
+			// `synonyms` is changed later on in the constrainedAbstractSqlModel.symbols get proxy
+			// `tables` is changed later on in the constrainedAbstractSqlModel.tables get proxy
+			!['synonyms', 'tables'].includes(prop) &&
```
Page-, line 723: Does the `_.cloneDeep` in https://github.com/balena-io/pinejs/blob/6675183ce12ff63202c374cee003ca55daa61bf8/src/sbvr-api/permissions.ts#L963 not remove the deep-freezing? Because the cloning is intended to avoid hitting issues with modifying the original model which is deep frozen, but if it's not removing the freezing as part of the cloning then that probably needs dealing with

fisehara: @Page- from my understanding the `constrainedAbstractSqlModel` should be deep frozen after the deepClone was made.
The freezing part applies to the `constrainedAbstractSqlModel` and at runtime the Proxies manipulate the dee frozen Object by writing parts to it through the proxies.

So I don't see that the _.cloneDeep is part of the freeze, please correct me here!

Page- (approved):

PR #658, opened 2023-04-20
Narrow the registered express HTTP methods

fisehara:
app.all registers too many not needed HTTP methods. As documented here:
https://expressjs.com/en/4x/api.html#app.METHOD

Change-type: patch

### DIFF
--- a/src/config-loader/config-loader.ts
+++ b/src/config-loader/config-loader.ts
@@ -195,8 +195,15 @@ export const setup = (app: Express.Application) => {
 						);
 
 						const apiRoute = `/${model.apiRoot}/*`;
-						app.options(apiRoute, (_req, res) => res.status(200).end());
-						app.all(apiRoute, sbvrUtils.handleODataRequest);
+						app
+							.route(apiRoute)
+							.options((_req, res) => res.status(200).end())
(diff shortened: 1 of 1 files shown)


src/config-loader/config-loader.ts
```diff
@@ -195,8 +195,15 @@ export const setup = (app: Express.Application) => {
...
-						app.all(apiRoute, sbvrUtils.handleODataRequest);
+						app
+							.route(apiRoute)
+							.get(sbvrUtils.handleODataRequest)
+							.put(sbvrUtils.handleODataRequest)
+							.post(sbvrUtils.handleODataRequest)
+							.patch(sbvrUtils.handleODataRequest)
+							.merge(sbvrUtils.handleODataRequest)
+							.delete(sbvrUtils.handleODataRequest)
+							.options((_req, res) => res.status(200).end());
```
Page-, line 206: I'd prefer to put `options` first in the list as it is a pre-flight request and hence will always be the first request to come in (if it does) so it feels like the right place to put it/is where I would expect it to be and look for it first

fisehara: I see, will put it on top of the list. Thanks

fisehara:
I self-certify!

PR #659, opened 2023-04-27
Support returning a number from async migrations rather than db result

Page-:
This also deprecates the form of returning a db result

Change-type: minor


Page-:
LGTM

PR #675, opened 2023-06-28
$batch

myarmolinsky:
MVP: $batch

Connects-to: https://github.com/balena-io/balena-api/pull/4501
Post-MVP: `atomicityGroup`, `if`, `dependsOn`, relative path `url`s, referencing with `$`, `nextLink`, `multipart` (see: https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_MultipartBatchFormat), `respond-async`
Change-type: major

See: https://balena.zulipchat.com/#narrow/stream/345746-aspect.2Fproduct/topic/Prompt.20When.20Restarting.20Containers


src/sbvr-api/sbvr-utils.ts
```diff
@@ -1094,7 +1097,7 @@ const $getAffectedIds = async ({
 	tx,
 }: HookArgs & {
 	tx: Db.Tx;
-}): Promise<number[]> => {
```
Page-, line 1097: Why did you change this?

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1158,6 +1162,84 @@ const runODataRequest = (req: Express.Request, vocabulary: string) => {
 		api[vocabulary].logger.log('Parsing', req.method, req.url);
 	}
 
+	if (req.url === `/${vocabulary}/$batch`) {
+		const { requests } = req.body as { requests: uriParser.UnparsedRequest[] };
```
Page-, line 1166: Can you separate the contents of this `if` into a separate function, eg `parseBatch`

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1205,17 +1287,22 @@ const runODataRequest = (req: Express.Request, vocabulary: string) => {
 			await runHooks('PREPARSE', reqHooks, { req, tx: req.tx });
 			let requests: uriParser.UnparsedRequest[];
 			// Check if it is a single request or a batch
-			if (req.batch != null && req.batch.length > 0) {
-				requests = req.batch;
+			if (req.url === `/${vocabulary}/$batch`) {
+				requests = req.batch!;
```
Page-, line 1291: I expect `req.batch` can be removed entirely now that the batch is just json in `req.body`

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1205,17 +1287,22 @@ const runODataRequest = (req: Express.Request, vocabulary: string) => {
...
 
 			const prepareRequest = async (
 				parsedRequest: uriParser.ParsedODataRequest &
 					Partial<Pick<uriParser.ODataRequest, 'engine' | 'translateVersions'>>,
 			): Promise<uriParser.ODataRequest> => {
+				if (models[parsedRequest.vocabulary] == null) {
+					throw new BadRequestError(
+						'Unknown vocabulary: ' + parsedRequest.vocabulary,
+					);
+				}
```
Page-, line 1305: I think this can be merged in a separate PR to shrink this one a bit

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1352,15 +1442,15 @@ export const handleODataRequest: Express.Handler = async (req, res, next) => {
 
 			// Otherwise its a multipart request and we reply with the appropriate multipart response
 		} else {
-			(res.status(200) as any).sendMulti(
-				responses.map((response) => {
+			res.status(200).json({
+				responses: responses.map((response) => {
 					if (response instanceof HttpError) {
-						response = httpErrorToResponse(response);
+						return httpErrorToResponse(response);
```
Page-, line 1448: This can be merged in a separate PR

src/sbvr-api/uri-parser.ts
```diff
@@ -60,8 +64,8 @@ export interface ODataRequest extends ParsedODataRequest {
 	modifiedFields?: ReturnType<
 		AbstractSQLCompiler.EngineInstance['getModifiedFields']
 	>;
-	affectedIds?: number[];
-	pendingAffectedIds?: Promise<number[]>;
+	affectedIds?: string[];
+	pendingAffectedIds?: Promise<string[]>;
```
Page-, line 68: Why did you change these? Technically they could be strings but by default they'll always be numbers (you have to do some sbvr magic to get pinejs to use non-number id fields)

src/sbvr-api/uri-parser.ts
```diff
@@ -263,15 +267,27 @@ export const metadataEndpoints = ['$metadata', '$serviceroot'];
...
 ): Promise<ParsedODataRequest[]>;
 export async function parseOData(
 	b: UnparsedRequest,
+	headers?: {
+		[key: string]: string | string[] | undefined;
+	},
 ): Promise<ParsedODataRequest | ParsedODataRequest[]>;
 export async function parseOData(
 	b: UnparsedRequest,
+	headers?: {
```
Page-, line 288: It might be worth naming this as `defaultHeaders` to be a bit clearer what it does

src/sbvr-api/uri-parser.ts
```diff
@@ -292,12 +308,14 @@ export async function parseOData(
 			const odata = memoizedParseOdata(url);
 
 			return {
+				id: b.id,
+				headers: { ...headers, ...b.headers } ?? headers,
```
Page-, line 312: Can the `?? headers` ever take effect?
```suggestion
				headers: { ...headers, ...b.headers },
```

src/sbvr-api/sbvr-utils.ts
myarmolinsky: According to the odata spec, id should be a string, not a number. See: https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#_Toc38457781:~:text=The%20value%20of%20id%20is%20a%20string%20containing%20the%20request%20identifier%20of%20the%20individual%20request

That is why I changed the typing for anything related to `id`s from `number` to `string`. Aside from just being a specified typing from the spec, it also enables us to use ids such as `"device"` to identify i.e. a device request and then refer to it in a subsequent request such as via `dependsOn: ["device"]`.

This is also the explanation for the following comment: https://github.com/balena-io/pinejs/pull/675/files/b5f6e0e7b5428c2642790f78c748a70d68998ea9#r1265570856

src/sbvr-api/uri-parser.ts
myarmolinsky: See: https://github.com/balena-io/pinejs/pull/675/files/b5f6e0e7b5428c2642790f78c748a70d68998ea9#r1266631186

src/sbvr-api/sbvr-utils.ts
myarmolinsky: Pushed in this commit: https://github.com/balena-io/pinejs/pull/675/commits/e01ace83e102b68fc0bef5911821a01ef8965122

Is this what you had in mind?

src/sbvr-api/sbvr-utils.ts
myarmolinsky: I assume you mean as a PR preceding this one? See: https://github.com/balena-io/pinejs/pull/678

src/sbvr-api/sbvr-utils.ts
myarmolinsky: See: https://github.com/balena-io/pinejs/pull/679

src/sbvr-api/uri-parser.ts
```diff
@@ -29,15 +29,19 @@ import * as sbvrUtils from './sbvr-utils';
 export type OdataBinds = ODataBinds;
 
 export interface UnparsedRequest {
+	id?: string;
```
fisehara, line 32: I think this needs to become a `requestId` to identify the request from a batch body

src/sbvr-api/uri-parser.ts
```diff
@@ -47,7 +51,7 @@ export interface ParsedODataRequest {
 	odataQuery: ODataQuery;
 	odataBinds: OdataBinds;
 	custom: AnyObject;
-	id?: number | undefined;
+	id?: string | undefined;
```
fisehara, line 54:
```suggestion
	id?: number | undefined;
    requestId?: string
```

src/sbvr-api/uri-parser.ts
```diff
@@ -292,12 +308,14 @@ export async function parseOData(
 			const odata = memoizedParseOdata(url);
 
 			return {
+				id: b.id,
```
fisehara, line 311:
```suggestion
				requestId: b.id,
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1108,6 +1111,7 @@ const $getAffectedIds = async ({
 	const parsedRequest: uriParser.ParsedODataRequest &
 		Partial<Pick<uriParser.ODataRequest, 'engine' | 'translateVersions'>> =
 		await uriParser.parseOData({
+			id: request.id,
```
fisehara, line 1114:
```suggestion
			requestId: request.id,
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1357,15 +1469,15 @@ export const handleODataRequest: Express.Handler = async (req, res, next) => {
 
 			// Otherwise its a multipart request and we reply with the appropriate multipart response
 		} else {
-			(res.status(200) as any).sendMulti(
-				responses.map((response) => {
+			res.status(200).json({
+				responses: responses.map((response) => {
 					if (response instanceof HttpError) {
```
fisehara, line 1474:
```suggestion
				responses: responses.map((response) => {
                    if (response.requestId) {
                        response.id = response.requestId;
                        delete response.requestId;
                    }
					if (response instanceof HttpError) {
```

fisehara (commented):
I would change statusCode and body as it's adding a little noise to this PR but it's fine.
The requestId for a $batch request are a different Id then the pineJs entity IDs.

src/sbvr-api/sbvr-utils.ts
myarmolinsky: @fisehara So even though the spec specifies that the property is called `id`, I need to name it something else (i.e. `requestId`) in pinejs?

src/sbvr-api/uri-parser.ts
```diff
@@ -29,15 +29,19 @@ import * as sbvrUtils from './sbvr-utils';
...
+	body?: any;
 	headers?: { [header: string]: string };
 	changeSet?: UnparsedRequest[];
 	_isChangeSet?: boolean;
 }
 
 export interface ParsedODataRequest {
+	headers?: {
+		[key: string]: string | string[] | undefined;
+	};
```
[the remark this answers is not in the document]
myarmolinsky: It's the `IncomingHttpHeaders` type from `http`, I'll use that as the typing for all these `headers`

src/sbvr-api/uri-parser.ts
```diff
@@ -263,15 +267,27 @@ export const metadataEndpoints = ['$metadata', '$serviceroot'];
 
 export async function parseOData(
 	b: UnparsedRequest & { _isChangeSet?: false },
+	headers?: {
+		[key: string]: string | string[] | undefined;
```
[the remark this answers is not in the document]
myarmolinsky: Changed it to use `IncomingHttpHeaders` from `http`

src/sbvr-api/uri-parser.ts
```diff
@@ -263,15 +267,27 @@ export const metadataEndpoints = ['$metadata', '$serviceroot'];
...
+	},
 ): Promise<ParsedODataRequest>;
 export async function parseOData(
 	b: UnparsedRequest & { _isChangeSet: true },
+	headers?: {
+		[key: string]: string | string[] | undefined;
```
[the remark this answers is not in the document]
myarmolinsky: Changed it to use `IncomingHttpHeaders` from `http`

src/sbvr-api/uri-parser.ts
```diff
@@ -263,15 +267,27 @@ export const metadataEndpoints = ['$metadata', '$serviceroot'];
...
+	},
 ): Promise<ParsedODataRequest[]>;
 export async function parseOData(
 	b: UnparsedRequest,
+	headers?: {
+		[key: string]: string | string[] | undefined;
```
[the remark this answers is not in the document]
myarmolinsky: Changed it to use `IncomingHttpHeaders` from `http`

src/sbvr-api/uri-parser.ts
```diff
@@ -263,15 +267,27 @@ export const metadataEndpoints = ['$metadata', '$serviceroot'];
...
+	},
 ): Promise<ParsedODataRequest | ParsedODataRequest[]>;
 export async function parseOData(
 	b: UnparsedRequest,
+	batchHeaders?: {
+		[key: string]: string | string[] | undefined;
```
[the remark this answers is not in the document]
myarmolinsky: Changed it to use `IncomingHttpHeaders` from `http`

test/06-batch.test.ts
```diff
@@ -0,0 +1,537 @@
...
+							id: '1',
+							method: 'POST',
+							url: '/university/student',
+							body: {
+								matrix_number: 100004,
+								name: faker.name.firstName(),
+								last_name: faker.name.lastName(),
+								studies_at__campus: 'bar',
+							},
+						},
```
fisehara, line 481: For readability it could be a defined validBatchPostRequest object, as it's 5 times the same.


```suggestion
						validBatchPostRequest
```

test/06-batch.test.ts
```diff
@@ -0,0 +1,537 @@
...
+
+	after(async () => {
+		testDeInit(pineServer);
+	});
+
+	describe('Basic', () => {
+		it('check /ping route is OK', async () => {
+			await supertest(testLocalServer).get('/ping').expect(200, 'OK');
+		});
+	});
```
fisehara, line 31: Not needed, as this basic test is covered in other units
```suggestion
```

test/06-batch.test.ts
```diff
@@ -0,0 +1,537 @@
...
+			await supertest(testLocalServer)
+				.post('/university/$batch')
+				.send({ requests: 'test' })
+				.expect(
+					400,
+					'"Batch requests must include an array of requests in the body via the \\"requests\\" property"',
+				);
+		});
+
+		// TODO: Seems we have default `continue-on-error` = `false`, but the docs specify `true`. Do we want to continue like this?
```
fisehara, line 118: I think, as long as request cannot depend on each other, pine can execute all request in parallel. If one fails, pine just return back one failed request and other successful ones.

I don't see an implementation of dependsOn, so request are all independent and it's more a request bundling and not a chained execution of requests.
https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_BatchRequest

test/06-batch.test.ts
```diff
@@ -0,0 +1,537 @@
...
+						},
+					],
+				})
+				.expect(
+					400,
+					'"Authorization may only be passed to the main batch request"',
+				);
+		});
+	});
+});
```
fisehara, line 537: I'm missing test cases for:

- request id request matches with returned response and Id. eg: mutliple gets for different resources, matching returned resources
- Expectation of two concurrent requests like 2 POST request sending the same matrix_number?
- At least some tests for $filter, $expand, ... Potentially a map of test parameters executed as test cases using the same structure [{testSet1}, {testSet2}, ...].foreach(it( run the test))

test/fixtures/06-batch/config.ts
```diff
@@ -0,0 +1,42 @@
...
+import { v1AbstractSqlModel, v1Translations } from './translations/v1';
+
+export const abstractSql = getAbstractSqlModelFromFile(modelFile);
+
+abstractSql.tables['student'].fields.push({
+	fieldName: 'computed field',
+	dataType: 'Text',
+	required: false,
+	computed: ['EmbeddedText', 'latest_computed_field'] as AbstractSqlQuery,
+});
```
fisehara, line 18: Do you need the computed field for the batch tests?

test/fixtures/06-batch/translations/v1/index.ts
```diff
@@ -0,0 +1,24 @@
...
+export const v1AbstractSqlModel = getAbstractSqlModelFromFile(
+	__dirname + '/university.sbvr',
+);
+
+v1AbstractSqlModel.tables['student'].fields.push({
+	fieldName: 'computed field',
+	dataType: 'Text',
+	required: false,
+	computed: ['EmbeddedText', 'v1_computed_field'] as AbstractSqlQuery,
+});
```
fisehara, line 16: Do you need this computed field for the batch tests?

src/sbvr-api/sbvr-utils.ts
```diff
@@ -133,7 +135,8 @@ export interface ApiKey extends Actor {
 }
 
 export interface Response {
-	statusCode: number;
+	id?: string;
+	status: number;
```
fisehara, line 139: I still think this is uneccesary noise. 
We can continue with the statusCode for now. It's an exported interface and we don't know the implications of changing it here. Very likely open-balena-api or balena-api may import the interface and may fail on it. 

It's not neccesary to rename the property to implement batch processing, no?

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1567,22 +1682,29 @@ const prepareResponse = async (
...
+			response = await respondDelete(req, request, result, tx);
+			break;
 		case 'OPTIONS':
-			return await respondOptions(req, request, result, tx);
+			response = await respondOptions(req, request, result, tx);
+			break;
 		default:
 			throw new MethodNotAllowedError();
 	}
+	return { ...response, id: request.batchRequestId };
```
fisehara, line 1707: As we don't want to return the id: as undefined when we are in the single request processing, it should better not return the id key at all. 
```suggestion
	return { ...response, id: request.batchRequestId ?? null }; // or
    return { ...response, ... ( request.batchRequestId ? {id: request.batchRequestId} :  null) }; or
    
    // Potentially the preferd one, as it's not needing the `...` object expand operator that could be heavy depending on the return object size.
    if (request.batchRequestId != null ) {
        response['id'] = request.batchRequestId
    }
    return response;
```

src/sbvr-api/sbvr-utils.ts
fisehara: It's your decision to evaluate the impact of the change, it makes sense, but maybe not now.

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1153,11 +1157,101 @@ const $getAffectedIds = async ({
...
+
+	const urls = new Set<string | undefined>(
+		requests.map((request) => request.url),
+	);
+	if (urls.has(undefined)) {
+		throw new BadRequestError('Requests of a batch request must have a "url"');
+	}
+	if (urls.has('/university/$batch')) {
+		throw new BadRequestError('Batch requests cannot contain batch requests');
+	}
```
fisehara, line 1235: Debugging or should it be? :
```suggestion
	if (urls.has('$batch')) {
		throw new BadRequestError('Batch requests cannot contain batch requests');
	}
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1205,11 +1299,20 @@ const runODataRequest = (req: Express.Request, vocabulary: string) => {
...
+				await Promise.all(
+					req.body.requests.map(
+						async (request: HookReq) =>
+							await runHooks('PREPARSE', reqHooks, {
+								req: request,
+								tx: req.tx,
+							}),
+					),
+				);
+				requests = req.body.requests;
```
fisehara, line 1312: This will execute all requests in one database transaction. Is this desired or should every request run in it's own database transaction?
The background is, if a database transaction fails the qhole transaction is rolled back, thus all batch requests will be rolled back.

@Page- What do you think, should every request in a batch run in its own db tx?

fisehara (commented):
Potentially there needs more test coverage to test the correctnes of delivered data back and forth when using batch requests.

Is there any limit on allowed batch requests in one batch?

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1273,7 +1376,13 @@ const runODataRequest = (req: Express.Request, vocabulary: string) => {
 
 			// Parse the OData requests
 			const results = await mappingFn(requests, async (requestPart) => {
-				const parsedRequest = await uriParser.parseOData(requestPart);
+				const parsedRequest = await uriParser.parseOData(
```
fisehara, line 1379: If the requests are independent, should they be executed in parallel on the DB or should pinejs limit the requests to be executed sequencially to do some kind of rate limiting?

src/sbvr-api/sbvr-utils.ts
myarmolinsky: Don't remember but I suspect it's neither, just a mistake 🤦

src/sbvr-api/sbvr-utils.ts
myarmolinsky: Yes, that is part of the point/idea of implementing $batch @fisehara , to run all requests in a batch on one transaction. Part of the following roadmap item: https://roadmap.balena.io/posts/19/allow-changes-to-variables-to-be-batched-and-submitted-all-at-once

src/sbvr-api/sbvr-utils.ts
myarmolinsky: Requests should be executed sequentially. If a request along the way fails, the batch request should fail. However, there will be no rollback if the requests are not atomic
> should pinejs limit the requests to be executed sequencially to do some kind of rate limiting?

I don't know enough about pine and the API to answer this well, your opinion given what I stated above would be better

test/06-batch.test.ts
myarmolinsky: - The 'successful request should have `responses` in its body' test sort of tests for the first bullet but I'll add a separate tests for multiple requests
- The requests should be sequential, so this shouldn't need testing, right?
- What exactly would this be for? This feature should not change how those work. Tests for $filter, $expand, etc. should live in their own test suite, no?

src/sbvr-api/uri-parser.ts
```diff
@@ -29,15 +29,19 @@ import * as sbvrUtils from './sbvr-utils';
 export type OdataBinds = ODataBinds;
 
 export interface UnparsedRequest {
+	id?: string;
 	method: string;
 	url: string;
-	data?: any;
+	body?: any;
```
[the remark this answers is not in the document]
Page-: This predates the odata json $batch but now that it exists I think it makes sense to match that format for our internal data structure as it should simplify a whole bunch of things

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1192,11 +1196,103 @@ export const getModel = (vocabulary: string) => {
...
+		);
+	}
+};
+
 const runODataRequest = (req: Express.Request, vocabulary: string) => {
 	if (env.DEBUG) {
 		api[vocabulary].logger.log('Parsing', req.method, req.url);
 	}
 
+	if (req.url.startsWith(`/${vocabulary}/$batch`)) {
```
Page-, line 1292: I think it'd be good to store this check somewhere as I've seen equivalent `.startsWith` checks in a few places

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1192,11 +1196,103 @@ export const getModel = (vocabulary: string) => {
...
+			throw new BadRequestError(
+				'GET and DELETE requests of a batch request must not have a body',
+			);
+		}
+	}
+
+	const urls = new Set<string | undefined>(
+		requests.map((request) => request.url),
+	);
+	if (urls.has(undefined)) {
```
Page-, line 1269: Should this also check for `null`?

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1192,11 +1196,103 @@ export const getModel = (vocabulary: string) => {
...
+		throw new BadRequestError('Requests of a batch request must have a "url"');
+	}
+	const containsBatch =
+		Array.from(urls).filter((url) => !!url?.includes('/$batch')).length > 0;
+	if (containsBatch) {
+		throw new BadRequestError('Batch requests cannot contain batch requests');
+	}
+	const urlModels = new Set(
+		Array.from(urls.values()).map((url: string) => url.split('/')[1]),
```
Page-, line 1278: You can limit the splitting to stop on the first match:
```suggestion
		Array.from(urls.values()).map((url: string) => url.split('/', 1)[1]),
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1192,11 +1196,103 @@ export const getModel = (vocabulary: string) => {
...
+	}
+
+	const urls = new Set<string | undefined>(
+		requests.map((request) => request.url),
+	);
+	if (urls.has(undefined)) {
+		throw new BadRequestError('Requests of a batch request must have a "url"');
+	}
+	const containsBatch =
+		Array.from(urls).filter((url) => !!url?.includes('/$batch')).length > 0;
```
Page-, line 1273:
```suggestion
		Array.from(urls).some((url) => !!url?.includes('/$batch'));
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1192,11 +1196,103 @@ export const getModel = (vocabulary: string) => {
...
+		throw new BadRequestError(
+			'All requests in a batch request must have unique string ids',
+		);
+	}
+
+	for (const request of requests) {
+		if (
+			request.headers != null &&
+			request.headers['content-type'] == null &&
+			(req.headers == null || req.headers['content-type'] == null)
```
Page-, line 1237:
```suggestion
			req.headers?.['content-type'] == null
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1192,11 +1196,103 @@ export const getModel = (vocabulary: string) => {
...
+		) != null
+	) {
+		throw new BadRequestError(
+			'Authorization may only be passed to the main batch request',
+		);
+	}
+	const ids = new Set<string>(
+		requests
+			.map((request) => request.id)
+			.filter((id) => typeof id === 'string') as string[],
```
Page-, line 1225:
```suggestion
			.filter((id): id is string => typeof id === 'string'),
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1192,11 +1196,103 @@ export const getModel = (vocabulary: string) => {
...
+			(request) =>
+				request.headers?.authorization != null ||
+				request.url?.includes('apikey='),
+		) != null
+	) {
+		throw new BadRequestError(
+			'Authorization may only be passed to the main batch request',
+		);
+	}
+	const ids = new Set<string>(
```
Page-, line 1222: The `string` should be inferred if a `string[]` is passed in?
```suggestion
	const ids = new Set(
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1192,11 +1196,103 @@ export const getModel = (vocabulary: string) => {
...
+	if (req.headers != null && req.headers['content-type'] == null) {
+		throw new BadRequestError(
+			'Headers in a batch request must include a "content-type" header if they are provided',
+		);
+	}
+	if (
+		requests.find(
+			(request) =>
+				request.headers?.authorization != null ||
+				request.url?.includes('apikey='),
```
Page-, line 1215: Technically params for authentication can be customized so this isn't perfect - I'm not sure if it's something we can account for but worth being aware of

PR #676, opened 2023-07-07
Mark the bin scripts as executable

thgreasi:
This should fix
> npx -p @balena/pinejs abstract-sql-compiler compile .materialized-config.json .materialized-schema.sql

which atm fails w/

> sh: 1: abstract-sql-compiler: Permission denied

Change-type: patch


thgreasi:
I self-certify!

thgreasi:
I self-certify!

Page- (approved):

PR #682, opened 2023-08-22
Balena lint v7

thgreasi:


src/database-layer/db.ts
```diff
@@ -497,6 +505,7 @@ if (maybePg != null) {
 			const p = new pg.Pool(config);
 			if (PG_SCHEMA != null) {
 				p.on('connect', (client) => {
+					// eslint-disable-next-line @typescript-eslint/no-floating-promises
```
thgreasi, line 508: @Page- wouldn't it be better if this had an  `await`?

src/database-layer/db.ts
```diff
@@ -686,12 +697,14 @@ if (maybePg != null) {
...
 				const client = await pool.connect();
 				const tx = new PostgresTx(client, false, stackTraceErr);
+				// eslint-disable-next-line @typescript-eslint/no-floating-promises
 				tx.executeSql('START TRANSACTION;');
 				return tx;
 			}),
 			readTransaction: createTransaction(async (stackTraceErr) => {
 				const client = await replica.connect();
 				const tx = new PostgresTx(client, false, stackTraceErr);
+				// eslint-disable-next-line @typescript-eslint/no-floating-promises
```
thgreasi, line 707: @Page- wouldn't it be better if all the `tx.executeSql` in the file had an  `await`?
Or at least the `START TRANSACTION` ones?

src/bin/abstract-sql-compiler.ts
```diff
@@ -11,6 +11,7 @@ import { program } from 'commander';
 
 const runCompile = (inputFile: string, outputFile?: string) => {
 	const { generateSqlModel } =
+		// eslint-disable-next-line @typescript-eslint/no-var-requires
```
fisehara, line 14: I'm wondering if we want to replace the requires with await import in that case?
```suggestion
	const { generateSqlModel } = (await import(
		'../sbvr-api/sbvr-utils'
	)) as typeof SbvrUtils;
```

fisehara (commented):
@thgreasi 

I'm wondering if we should replace the `const x = require ()` with `const x = (await import())`

src/bin/abstract-sql-compiler.ts
thgreasi: @fisehara Doing that would require marking the function as an async one, which I guessed that @Page- would have already done that if possible. After trying it though it seems that it works fine when running the commends from the bC api scripts (unless I'm missing anything) so great idea :+1:

thgreasi: Here is an extra PR for you to review and try :)
See: https://github.com/balena-io/pinejs/pull/683

src/passport-pinejs/passport-pinejs.ts
```diff
@@ -66,7 +67,11 @@ const setup: ConfigLoader.SetupFunction = async (app: Express.Application) => {
 
 		logout = (req, _res, next) => {
 			req.logout((error) => {
-				error ? next(error) : next();
+				if (error) {
+					next(error);
+					return;
```
fisehara, line 72:
```suggestion
					return next(error);
```

thgreasi:
I self-certify

thgreasi:
I self-certify!

PR #683, opened 2023-08-24
bin: Use async imports in compile-schema and generate-types cli commands

thgreasi:
Change-type: patch
See: https://github.com/balena-io/pinejs/pull/682#discussion_r1304255389


fisehara (approved):
Thx

PR #687, opened 2023-08-29
Update TypeScript to 5.2.2

thgreasi:
Depends-on: https://github.com/balena-io/pinejs/pull/682
Change-type: patch


thgreasi:
I self-certify!

PR #688, opened 2023-08-29
Fix failure on delete requests with invalid ID

otaviojacobi:
Change-type: patch

### DIFF
--- a/test/06-webresource.test.ts
+++ b/test/06-webresource.test.ts
@@ -216,7 +216,7 @@ describe('06 webresources tests', function () {
 					const uniqueFilename = `${randomUUID()}_${filename}`;
 
 					await supertest(testLocalServer)
-						.patch(`/${resourceName}/organization(42)`)
+						.patch(`/${resourceName}/organization(4242)`)
 						.field('name', 'john')
 						.attach(resourcePath, newFilePath, {
 							filename: uniqueFilename,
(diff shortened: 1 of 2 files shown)


test/06-webresource.test.ts
```diff
@@ -227,6 +227,12 @@ describe('06 webresources tests', function () {
 					expect(await isEventuallyDeleted(uniqueFilename)).to.be.true;
 				});
 
+				it(`does not fail if delete on ${resourcePath} has no affected ids`, async () => {
```
fisehara, line 230: Should this now also test for a patch / post with empty webresource fields?

otaviojacobi: @fisehara this is already tested in a test I wrote previously. The issue didn't happen in there because there was affected IDs. I will add one for a parch on a inexistant id without any webresource which would also trigger the issue

otaviojacobi: AH! I think I understand what you mean. You mean when `getWebResourcesKeysFromRequest` is called from another point in the code. Let me check

otaviojacobi: So the explanation is that PATCH would never cause this problem here because the code would have jumped out of the execution in case there was no webresource on the actual request being done (an optimisation that ended up fixing this for that case). But I also added two more tests for patches in order to future proof this. The issue only existed on DELETE requests because these will never have a body and then we can't do the same optmization.

fisehara: Did this problem also evolve on a new Post on the tests? @otaviojacobi

otaviojacobi: @fisehara a new post runs a different hook which does not have any of the problems, also the tests already present for post cover all meanigfull scenarios that I can see

otaviojacobi: Ah @fisehara there is actually a way to reuse the function we created on the POST hook, let me do it aswell

otaviojacobi: Small change to use the generic function and added a POST test! @fisehara

otaviojacobi:
LGTM!

PR #689, opened 2023-08-30
Fix $expand on parent resource with child webresource

otaviojacobi:
Change-type: patch

Context: When serving webresources we may need aditional postprocessing (e.g. signing the URL). This was previously done with a PRERESPONSE hook on the model that had a webresource. 

The problem with this approach is that one model that does not contain a webresource can be $expand on a field that contains a webresource and then the GET hook won't be setup and postprocessing (presigning) won't occur.

This PR solves the issue by presigning a level deeper into the stack. When we fetch a resource from the DB that is a webresource, we postprocess as a webresource will never be served without the postprocessing.

### DIFF
--- a/src/sbvr-api/odata-response.ts
+++ b/src/sbvr-api/odata-response.ts
@@ -12,6 +12,9 @@ declare module '@balena/abstract-sql-compiler' {
 		localFields?: {
 			[odataName: string]: true;
 		};
+		webresourceFields?: {
+			[odataName: string]: true;
+		};
 	}
 }
 
@@ -21,6 +24,7 @@ import { sqlNameToODataName } from '@balena/odata-to-abstract-sql';
 import sbvrTypes, { SbvrType } from '@balena/sbvr-types';
 import * as _ from 'lodash';
 import { resolveNavigationResource, resolveSynonym } from './sbvr-utils';
+import { getWebresourceHandler } from '../webresource-handler';
 
 const checkForExpansion = async (
 	vocab: string,
@@ -98,6 +102,20 @@ const getLocalFields = (table: AbstractSqlTable) => {
 	}
 	return table.localFields;
 };
+
+const getWebResourceFields = (table: AbstractSqlTable) => {
+	if (table.webresourceFields == null) {
+		table.webresourceFields = {};
+		for (const { fieldName, dataType } of table.fields) {
+			if (dataType === 'WebResource') {
+				const odataName = sqlNameToODataName(fieldName);
+				table.webresourceFields[odataName] = true;
+			}
+		}
+	}
+	return table.webresourceFields;
+};
+
 const getFetchProcessingFields = (table: AbstractSqlTable) => {
 	return (table.fetchProcessingFields ??= _(table.fields)
 		.filter(
@@ -140,6 +158,7 @@ export const process = async (
 		vocabulary: vocab,
 		resourceName,
 	});
+	const configuredWebResourceHandler = getWebresourceHandler();
 	const table = abstractSqlModel.tables[sqlResourceName];
 
 	const fieldNames = Object.keys(rows[0]);
@@ -157,6 +176,11 @@ export const process = async (
 			!Object.prototype.hasOwnProperty.call(localFields, fieldName),
 	);
 
+	const webresourceFields = getWebResourceFields(table);
+	const requiredSigningFields = fieldNames.filter((fieldName) =>
+		Object.prototype.hasOwnProperty.call(webresourceFields, fieldName),
+	);
+
 	const odataIdField = sqlNameToODataName(table.idField);
 	for (const row of rows) {
 		for (const fieldName of processedFields) {
@@ -167,6 +191,23 @@ export const process = async (
 				uri: resourceURI(vocab, resourceName, row[odataIdField]),
 			};
 		}
+	}
+
+	if (
+		requiredSigningFields.length > 0 &&
+		configuredWebResourceHandler != null
+	) {
+		await Promise.all(
+			rows.map(async (row) => {
+				requiredSigningFields.map(async (fieldName) => {
+					if (row[fieldName] != null) {
+						row[fieldName] = await configuredWebResourceHandler.onPreRespond(
+							row[fieldName],
+						);
+					}
+				});
+			}),
+		);
 	}
 
 	if (expandableFields.length > 0) {
@@ -195,5 +236,6 @@ export const prepareModel = (abstractSqlModel: AbstractSqlModel) => {
 	_.forEach(abstractSqlModel.tables, (table) => {
 		getLocalFields(table);
 		getFetchProcessingFields(table);
+		getWebResourceFields(table);
 	});
 };
--- a/src/webresource-handler/index.ts
+++ b/src/webresource-handler/index.ts
@@ -56,6 +56,14 @@ const getLogger = (vocab?: string): Console => {
 	return console;
 };
 
+let configuredWebResourceHandler: WebResourceHandler | undefined;
+export const setupWebresourceHandler = (handler: WebResourceHandler): void => {
+	configuredWebResourceHandler = handler;
+};
+export const getWebresourceHandler = (): WebResourceHandler | undefined => {
+	return configuredWebResourceHandler;
+};
+
 const isFileInValidPath = async (
 	fieldname: string,
 	req: Express.Request,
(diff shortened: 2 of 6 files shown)


src/webresource-handler/index.ts
```diff
@@ -229,37 +248,6 @@ const getCreateWebResourceHooks = (
...
-		PRERESPOND: async ({ request, response }) => {
-			const fields = getWebResourceFields(request, false);
-
-			if (fields.length === 0) {
-				return;
-			}
-
-			if (typeof response.body === 'object' && response.body.d != null) {
-				const transformedItems = await Promise.all(
-					response.body.d.map(async (responseItem: any) => {
```
otaviojacobi, line 246: And we remove one `any` for free

src/sbvr-api/odata-response.ts
```diff
@@ -158,9 +162,17 @@ export const process = async (
...
 
 	const odataIdField = sqlNameToODataName(table.idField);
+	const webResourceHandler = getWebresourceHandler();
 	for (const row of rows) {
 		for (const fieldName of processedFields) {
 			row[fieldName] = fetchProcessingFields[fieldName](row[fieldName]);
+			if (
+				webResourceHandler &&
+				isWebResourceField(fieldName, table) &&
+				row[fieldName]?.href != null
```
fisehara, line 172: Checking `row[fieldName]?.href != null` should be encapsuled in a webresource function as it is webResource specific knowledge

src/sbvr-api/odata-response.ts
```diff
@@ -158,9 +162,17 @@ export const process = async (
 	);
 
 	const odataIdField = sqlNameToODataName(table.idField);
+	const webResourceHandler = getWebresourceHandler();
 	for (const row of rows) {
 		for (const fieldName of processedFields) {
 			row[fieldName] = fetchProcessingFields[fieldName](row[fieldName]);
+			if (
+				webResourceHandler &&
```
fisehara, line 170: Potentially check the webResourceHandler before the loop and create a different function that is called in the loop.

src/sbvr-api/odata-response.ts
```diff
@@ -158,9 +162,17 @@ export const process = async (
...
+				isWebResourceField(fieldName, table) &&
```
fisehara, line 171: I don't know about the performance, but the table is static for each iteration, potentially the isWebResourceFiled can be wrapped into a table specific function and in the loop the tableSpecific function is executed?

fisehara: isWebResourceFieled could be optimised into a 
`entityContainsWebResourceField` to select a table being a webResource containing table, instead of iterating over all fields on the table for the specific fieldname?

fisehara (commented):
Please add test cases to do regular expands with this change. I know that they are missing from the test set, but changing the expand code should ensure that the expanding works. In specific chained expands.

src/sbvr-api/odata-response.ts
```diff
@@ -107,11 +109,24 @@ const getFetchProcessingFields = (table: AbstractSqlTable) => {
...
-					.fetchProcessing,
-			];
+			const sbvrDataType = dataType as keyof typeof sbvrTypes;
+			let fetchProcessingFn = (sbvrTypes[sbvrDataType] as SbvrType)
+				.fetchProcessing;
+
+			// fetchProcessing for WebResources needs to be extended in order
+			// to support presigning a URL immedeately after it is fetched from the database
+			// as no URL can be served without the signing
+			if (sbvrDataType === 'WebResource') {
```
otaviojacobi, line 119: @fisehara I tried to solve performance issues by going one step deeper into the stack here. Now instead of having to figure it out if a row/fieldname is a webresource we bind the `fetchProcessing` function incoming from sbvrTypes with the onPreRespond hook. This operation happens before the expensive MxN loop on process and should be fairly cheap, and avoid us having to do any extra checks in there

src/sbvr-api/odata-response.ts
```diff
@@ -158,9 +173,17 @@ export const process = async (
 	);
 
 	const odataIdField = sqlNameToODataName(table.idField);
+	const fetchProcessingPromises: Array<Promise<void>> = [];
 	for (const row of rows) {
 		for (const fieldName of processedFields) {
 			row[fieldName] = fetchProcessingFields[fieldName](row[fieldName]);
+			if (row[fieldName] instanceof Promise) {
```
otaviojacobi, line 180: @fisehara  We also could just do `await` on the fetchProcessingFields as the presigning promises should be really fast because of the caching on the signing but I think it is very weird to have nested for loop with await inside, so I optmised with `Promise.all` the impact for anything that is not a webresource is basically a if check on instance that I would expect to be fairly fast

otaviojacobi:
@Page- @fisehara I tried to fix possile performance problems the best way I knew it, please let me know if you have other/better ideas

src/sbvr-api/odata-response.ts
```diff
@@ -107,11 +109,24 @@ const getFetchProcessingFields = (table: AbstractSqlTable) => {
...
+			const sbvrDataType = dataType as keyof typeof sbvrTypes;
+			let fetchProcessingFn = (sbvrTypes[sbvrDataType] as SbvrType)
+				.fetchProcessing;
+
+			// fetchProcessing for WebResources needs to be extended in order
+			// to support presigning a URL immedeately after it is fetched from the database
+			// as no URL can be served without the signing
+			if (sbvrDataType === 'WebResource') {
+				const originalFetchProcessing = fetchProcessingFn;
+				fetchProcessingFn = (data: WebResourceType) => {
```
Page-, line 121: Is there any reason why this is not built in to the data type's `fetchProcessing` itself?

src/sbvr-api/odata-response.ts
```diff
@@ -158,9 +173,17 @@ export const process = async (
 	);
 
 	const odataIdField = sqlNameToODataName(table.idField);
+	const fetchProcessingPromises: Array<Promise<void>> = [];
```
Page-, line 176: If this is being handled in pinejs like this then I think I'd prefer it be a separate pass like `expandableFields` has where it's done in parallel for only the single field (and only if such a field is included). Obviously if there are benchmarks proving one way or the other to be faster then that'd win but afaik we don't atm

src/sbvr-api/odata-response.ts
otaviojacobi: @Page- Adding it directly into sbvr-types would require we changed the prototype of `fetchProcessing` to either receive two parameters (the data and the handler) or to have fetchProcessing to receive one parameter data (type `any`) that put together both the webResourceHandler and the data. Either way, it felt that this signing logic did not belong into something that should just handle sbvr types, so that is why I baked into pinejs (and either approach we go it would still require us to check if the type is a WebResource and then forward the webresourceHandler for signing)

src/sbvr-api/odata-response.ts
otaviojacobi: I see. With that we can actually stop doing the function that changes the prerespond hook aswell. Will fix it

otaviojacobi: @Page- Fixed it as I think it is what you suggested. The good thing about this approach is that it does not change anything on the request time of requests without a webresource, which I think was your goal here. Let me know what do you think :+1:

src/sbvr-api/odata-response.ts
```diff
@@ -169,6 +192,20 @@ export const process = async (
...
+					if (configuredWebResourceHandler != null && row[fieldName] != null) {
+						row[fieldName] = await configuredWebResourceHandler.onPreRespond(
+							row[fieldName],
+						);
+					}
+				});
+			}),
+		);
+	}
+
```
fisehara, line 208: This should be in line with the loop before, as it otherwise iterates again through all rows. 
Also it should only run if `configuredWebResourceHandler != null` as this is static to the loop.

```suggestion
	if (requiredSigningFields.length > 0) {
		await Promise.all(
			rows.map(async (row) => {
				requiredSigningFields.map(async (fieldName) => {
					if (configuredWebResourceHandler != null && row[fieldName] != null) {
						row[fieldName] = await configuredWebResourceHandler.onPreRespond(
							row[fieldName],
						);
					}
				});
			}),
		);
	}

```

test/06-webresource.test.ts
```diff
@@ -801,6 +801,39 @@ describe('06 webresources tests', function () {
 					await Promise.all(promises);
 				});
 
+				it('is able to expand an organization with its resources', async () => {
+					const { body: org } = await supertest(testLocalServer)
```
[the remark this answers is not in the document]
otaviojacobi: nice, done!

src/sbvr-api/odata-response.ts
otaviojacobi: "This should be in line with the loop before, as it otherwise iterates again through all rows." I think this is exactly the behavior we want. 
"Also it should only run if configuredWebResourceHandler != null as this is static to the loop." Done!

src/sbvr-api/odata-response.ts
```diff
@@ -21,6 +24,7 @@ import { sqlNameToODataName } from '@balena/odata-to-abstract-sql';
 import sbvrTypes, { SbvrType } from '@balena/sbvr-types';
 import * as _ from 'lodash';
 import { resolveNavigationResource, resolveSynonym } from './sbvr-utils';
+import { configuredWebResourceHandler } from '../webresource-handler';
```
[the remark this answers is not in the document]
otaviojacobi: done!

fisehara (approved):

PR #691, opened 2023-08-31
Add checksum calculation to WebResource

otaviojacobi:
Change-type: minor

### DIFF
--- a/src/webresource-handler/handlers/S3Handler.ts
+++ b/src/webresource-handler/handlers/S3Handler.ts
@@ -16,9 +16,11 @@ import {
 import { Upload } from '@aws-sdk/lib-storage';
 import { getSignedUrl } from '@aws-sdk/s3-request-presigner';
 
-import { randomUUID } from 'crypto';
+import { randomUUID, createHash } from 'crypto';
+import type { Hash } from 'crypto';
 import type { WebResourceType as WebResource } from '@balena/sbvr-types';
 import * as memoize from 'memoizee';
+import { Readable, Transform } from 'stream';
 
 export interface S3HandlerProps {
 	region: string;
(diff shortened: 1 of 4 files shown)


src/webresource-handler/handlers/S3Handler.ts
```diff
@@ -29,12 +30,55 @@ export interface S3HandlerProps {
...
+			if (this.calculateChecksums) {
+				this.stream.on('data', (chunk: Buffer) => {
+					hash.update(chunk);
+					this.passthrough.write(chunk);
+				});
+
+				this.stream.on('end', () => {
+					resolve(hash.digest('hex'));
+					this.passthrough.end();
+				});
```
pranasziaukas, line 58: Can this fail? If so, would we like to handle such failure at this point in code?

I am mostly referring to `error` events -- https://nodejs.org/dist/latest-v18.x/docs/api/stream.html

src/webresource-handler/handlers/S3Handler.ts
```diff
@@ -61,6 +105,7 @@ export class S3Handler implements WebResourceHandler {
 		this.maxFileSize = config.maxSize ?? 52428800;
 		this.bucket = config.bucket;
 		this.client = new S3Client(this.config);
+		this.calculateChecksums = config.calculateChecksums ?? true;
```
pranasziaukas, line 108: What are the implications on enabling checksum calculations by default? For example, the performance overhead for large files, many files, etc.

pranasziaukas (commented):
AFAICT, the PR implements a new `ChecksumCalculator` class to optionally compute checksums during S3 uploads. This feature is integrated into the `S3Handler` upload process and is reflected in the upload response type. The functionality is covered with reasonable unit tests.

Everything makes sense to me, nice job 👍

src/webresource-handler/handlers/S3Handler.ts
otaviojacobi: Very nice catch @pranasziaukas 
In order to catch all the events and properly pipe them, I updated the code to make use of a pipe, therefore properly piping all the events without having to manually enumerate them.

Now basically we implement the `Transform` interface from node stream api where we basically pipe the data with a side effect which is calculating the checksum. .update() and .digest() methods should not throw, so we are fine just calling them without try/catches. Let me know what do you think!

src/webresource-handler/handlers/S3Handler.ts
otaviojacobi: @pranasziaukas now that's a great question I would also want to know the answer. I suppose that the main risk associated is with higher CPU usage as more and more entities on our DB use webresources. I guess that if proven to be a problem we could change the implementation to be on resource level. For example, we would want checksum calculation for release assets but it might be overkill for logo images... 

For now I would say we are safe with the current behaviour (we are not doing anything like storing the whole file in memory etc) but we do have to pipe all the file content. 

tl;dr: Idk how this will behave at larger scale, but we have alternatives to reduce impact if it happens

otaviojacobi:
Moving to draft.... there is a better way using s3 to calculate the hash for us

otaviojacobi:
FWIW We can't use native s3 checksum calculations: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html#large-object-checksums

The native implementation of checksum is not a checksum of the entire file but rather it will calculate a sha256 (or any algorithm...) of each uploaded chunk, concatenate all the generated hashes and then sha256 (or any other algorithm...) encode this whole thing and append how many chunks were added...

this make for a worse overall experience as 
1) the checksums will vary if we change our upload chunk size
2) if you open your terminal and `sha256sum` the file, the value won't be the same stored in our DB/S3

otaviojacobi:
Closing in favor of adding checksums as a DB field only to resources where it is relevant and allowing customers to post the checksum themselves.

PR #719, opened 2023-12-18
Improve handling of requests trying to access non-existent resoruces

Page-:


otaviojacobi (approved):
@Page- I think you just may need to re-run the linter on db.ts file

PR #722, opened 2023-12-18
Webresource fixes

Page-:


src/webresource-handler/index.ts
```diff
@@ -81,22 +81,30 @@ const isFileInValidPath = async (
 		url: req.url,
 		method: req.method,
 	});
-	const resourceName = sbvrUtils.resolveSynonym(odataRequest);
+	const sqlResourceName = sbvrUtils.resolveSynonym(odataRequest);
```
Page-, line 84: `resolveSynonym` returns the sql name so I renamed this for clarity

src/webresource-handler/index.ts
```diff
@@ -81,22 +81,30 @@ const isFileInValidPath = async (
...
+	if (table == null) {
+		return false;
+	}
 
 	const permission = req.method === 'POST' ? 'create' : 'update';
 	const vocab = model.versions[model.versions.length - 1];
 
 	// Checks if it has permissions on both the original resourceName or any synonym
-	const hasPermissions =
-		(await checkPermissions(req, permission, resourceName, vocab)) ||
```
Page-, line 91: I removed the permission check on `resourceName` since it's a sql name and as such won't (or at least shouldn't) have permissions associated

src/webresource-handler/index.ts
```diff
@@ -81,22 +81,30 @@ const isFileInValidPath = async (
...
+		vocab,
+	);
 
 	if (!hasPermissions) {
 		return false;
 	}
 
-	const sqlResourceName = odataNameToSqlName(resourceName);
-	const fields = model.abstractSql.tables[sqlResourceName].fields;
+	const fields = table.fields;
```
otaviojacobi, line 107: NIT: Maybe we just remove the whole fields variable and on the `.some` below we do 
```ts
return table.fields.some(...)
```

I did create the variable before because `model.abstractSql.tables[sqlResourceName].fields;` was a bit too long to inline, but `table.fields` should be ok and it is clear to read

PR #723, opened 2023-12-21
Fix not found on canAccess

otaviojacobi:


src/sbvr-api/sbvr-utils.ts
```diff
@@ -1270,7 +1270,11 @@ const runODataRequest = (req: Express.Request, vocabulary: string) => {
 				>;
 				// Add/check the relevant permissions
 				try {
-					if (abstractSqlModel.tables[resolveSynonym($request)] == null) {
+					const resolvedResourceName = resolveSynonym($request);
+					if (
+						!resolvedResourceName.includes('#canAccess') &&
```
Page-, line 1275: Specifically this should be `endsWith` right? And I assume that would also be more performant

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1270,7 +1270,11 @@ const runODataRequest = (req: Express.Request, vocabulary: string) => {
...
+						abstractSqlModel.tables[resolvedResourceName] == null
```
Page-, line 1276: I'd put this check first because I'd expect it to pass the majority of the time and therefore avoid running the second check most often

test/07-permissions.test.ts
```diff
@@ -17,7 +17,7 @@ describe('07 permissions tests', function () {
 		testDeInit(pineServer);
 	});
 
-	describe('university vocabulary with guest permissions', () => {
+	describe('university vocabulary with admin permissions', () => {
```
Page-, line 20: This change should ideally be in the tests commit

src/server-glue/module.ts
```diff
@@ -12,6 +12,7 @@ import { PINEJS_ADVISORY_LOCK } from '../config-loader/env';
 
 export * as dbModule from '../database-layer/db';
 export { PinejsSessionStore } from '../pinejs-session-store/pinejs-session-store';
+export { mountLoginRouter } from '../passport-pinejs/mount-login-router';
```
Page-, line 15: This is technically a minor change because it adds/exports a new feature

Page- (approved):

PR #734, opened 2024-01-17
Change NotFound for Unauthorized when requesting a resource that does…

otaviojacobi:
… not exists

Change-type: patch


Page- (approved):
Might be good to add a comment that we 401 instead of 404 to not differentiate between it not existing and not having access

PR #739, opened 2024-02-06
Add async tasks

joshbwlng:
Change-type: minor

---

Spec: https://balena.fibery.io/Inputs/Research/PineJS-Async-Tasks-483

Testing locally:
```bash
docker-compose -f docker-compose.npm-test.yml up -d
export DATABASE_URL=postgres://docker:docker@localhost:5431/postgres
export PINEJS_QUEUE_CONCURRENCY=1
TZ=UTC npm run mocha test/08-tasks.test.ts
```

---

Follow-ups:
- Add support for `FOR UPDATE SKIP LOCKED` to abstractSql and update the polling query for tasks to be written in abstractSql
- Add generic `.on('notification')` handler for listening for db events, such as new task inserts

---

Depends on:
- https://github.com/balena-io-modules/sbvr-types/pull/106
- https://github.com/balena-io/pinejs/pull/775
- https://github.com/balena-io-modules/abstract-sql-to-typescript/pull/76
- https://github.com/balena-io-modules/pinejs-client-js/pull/167


test/fixtures/08-tasks/task-handlers.ts
```diff
@@ -0,0 +1,60 @@
+import type { FromSchema } from 'json-schema-to-ts';
+import { sbvrUtils } from '../../../src/server-glue/module';
+
+// Define JSON schema for accepted parameters
+const createDeviceParamsSchema = {
```
joshbwlng, line 5: Note: I also investigated using [zod](https://www.npmjs.com/package/zod) to define parameter object shapes, but decided to stick with JSON Schema as its a commonly used standard. Consumers can still define their shapes with zod and then convert to JSON Schema using something like [zod-to-json-schema](https://www.npmjs.com/package/json-schema-to-zod) when defining task handlers.

test/fixtures/08-tasks/task-handlers.ts
```diff
@@ -0,0 +1,62 @@
...
+	sbvrUtils.addTaskHandler(
+		'create_device',
+		async (options) => {
+			try {
+				const params = options.params as CreateDeviceParams;
+				await options.api.post({
+					apiPrefix: '/example/',
+					resource: 'device',
+					passthrough: {
+						tx: options.tx,
```
dfunckt, line 32: My understanding is this tx comes from poll() (or start()?) and it always commits in the end. This means that, in general, handlers can do some work and then fail in some way but the work they did will not be rolled back. Is this intentional?

joshbwlng: @dfunckt Thanks for checking out the PR! Initially the decision was to expect handlers to do some of the lifting here, but I see what you mean and agree. I've pushed a (somewhat hastily put together) commit that should achieve this, and asserts as much in a test. Let me know what you think!

dfunckt: Thanks @joshbwlng. This looks like it’ll work nicely. The only thing I’m not sure of is if Postgres will allow a new transaction (the handlerTx) on the same connection that already has a transaction open (the “logistics” one).

joshbwlng: I'm honestly not sure either, which is why I've kept it in its own commit for now. Will come back to this when I have more time.

joshbwlng: @dfunckt Pushed again (this time with a squash), using a savepoint instead. Each savepoint is unique to the task being executed and depending on the result we can either release the savepoint or roll back to it. Much cleaner in my opinion.

src/tasks/worker.ts
```diff
@@ -0,0 +1,252 @@
...
+					startedOnTime,
+					'failed',
+					`Invalid parameter set: ${ajv.errorsText(handler.validate.errors)}`,
+				);
+				return;
+			}
+
+			// Execute handler
+			await tx.executeSql(`SAVEPOINT task_${task.id}`);
+			const result = await handler.fn({
```
dfunckt, line 84: Can `handler.fn()` throw? If it does, then the SAVEPOINT will not be released (but I don't know if that's a problem in practice).

test/fixtures/08-tasks/task-handlers.ts
joshbwlng: We do currently nest transactions in our migration code [here](https://github.com/balena-io/pinejs/blob/master/src/migrator/async.ts#L236C1-L236C12), so going back to the way I had it before at least is a proper option. I switched to savepoints (a) to experiment and (b) because we don't use the transaction for "logistics" before passing it off to the handler or executing savepoint release/rollback.

src/tasks/worker.ts
joshbwlng: The idea came to during the shaping meeting (after you left I believe) was to initially require handlers to always catch, which is why the code kills the process in the catch here in `execute()`.

src/tasks/worker.ts
```diff
@@ -0,0 +1,263 @@
...
+				console.error('Failed polling for tasks:', err);
+			})
+			.finally(() => {
+				setTimeout(() => this.poll(), executed ? 0 : this.interval);
+			});
+	}
+
+	// Start listening and polling for tasks
+	public start(): void {
+		sbvrUtils.db.on?.(
```
Page-, line 242: Why is the `?.` needed? I feel like if it's ever undefined then the entire task system would break and that's not something we'd want to silently ignore

src/tasks/worker.ts
```diff
@@ -0,0 +1,263 @@
...
+			'notification',
+			async (msg) => {
+				if (this.canExecute()) {
+					await sbvrUtils.db.transaction(async (tx) => {
+						const result = await tx.executeSql(
+							`SELECT ${selectColumns} FROM task AS t WHERE id = $1 FOR UPDATE SKIP LOCKED`,
```
Page-, line 248: Is this compliant with postgres+mysql+sqlite? If not we should have a warning if you're trying to use with an unsupported db engine

src/tasks/worker.ts
```diff
@@ -0,0 +1,263 @@
...
+	// Poll for tasks to execute
+	private poll(): void {
+		let executed = false;
+		const handlerNames = Object.keys(this.handlers);
+		const binds = handlerNames.map((_, index) => `$${index + 1}`).join(', ');
+		sbvrUtils.db
+			.transaction(async (tx) => {
+				if (!this.canExecute()) {
+					return;
+				}
```
Page-, line 199: This should be outside of the transaction imo to avoid the unnecessary db interaction. Presumably you've put it inside to be in a promise chain but I think it'd be better to do something like
```ts
(async () => {
	try {
		let executed = false;
		const handlerNames = Object.keys(this.handlers);
		const binds = handlerNames.map((_, index) => `$${index + 1}`).join(', ');
		if (!this.canExecute()) {
			return;
		}
		await sbvrUtils.db.transaction(async (tx) => {
			...
		});
	} catch (err) {
		...
	} finally {
		...
	}
})();
```
to be able to use async/await as usual and put the try/catch/finally around everything

src/tasks/worker.ts
```diff
@@ -0,0 +1,263 @@
...
+						task.is_scheduled_with__cron_expression,
+					priority: task.priority,
+				},
+			});
+		}
+	}
+
+	// Calculate next attempt time using exponential backoff
+	private getNextAttemptTime(attempt: number): Date | null {
+		const delay = Math.ceil(Math.exp(Math.min(10, attempt))) * 1000;
```
Page-, line 186: I'm pretty sure this is not doing what you expect as by my calculations this would be setting the minimum delay to a bit over 6 hours (e^10)? I think it probably makes sense to use logic more similar to https://github.com/balena-io-modules/pinejs-client-js/blob/v6.14.6/src/index.ts#L1123 for the next delay time

src/tasks/worker.ts
```diff
@@ -0,0 +1,263 @@
...
+			['failed', 'succeeded'].includes(body.status) &&
+			task.is_scheduled_with__cron_expression != null
+		) {
+			await this.client.post({
+				resource: 'task',
+				passthrough: {
+					tx,
+					req: permissions.root,
+				},
+				body: {
```
Page-, line 171:
```suggestion
				},
				options: {
					returnResource: false,
				},
				body: {
```

src/tasks/worker.ts
```diff
@@ -0,0 +1,263 @@
...
+				if (!(err instanceof TransactionClosedError)) {
+					throw err;
+				}
+			} finally {
+				// Update task with results
+				await this.finalize(tx, task, startedOnTime, status, error);
+			}
+		} catch (err) {
+			// This shouldn't happen, but if it does we want to log and kill the process
+			console.error('Task execution failed:', err);
```
Page-, line 114: Can you also add some details about the task, eg the task handler name, since I expect it'll help significantly with debugging

src/tasks/worker.ts
```diff
@@ -0,0 +1,263 @@
...
+					`Invalid parameter set: ${ajv.errorsText(handler.validate.errors)}`,
+				);
+				return;
+			}
+
+			// Execute handler
+			let status: TaskStatus = 'queued';
+			let error: string | undefined;
+			try {
+				await sbvrUtils.db.transaction(async (handlerTx) => {
```
Page-, line 87: What's the reasoning between us handling the transaction creation? I could definitely imagine tasks that don't need a db transaction and so creating one for them would unnecessarily use up db connections

src/tasks/model.sbvr
```diff
@@ -0,0 +1,60 @@
+Vocabulary: tasks
```
Page-, line 1: I'd name this file `tasks.sbvr` to be a bit clearer

src/tasks/index.ts
```diff
@@ -0,0 +1,174 @@
...
+				} catch (_) {
+					throw new BadRequestError(
+						`Invalid cron expression: ${request.values.is_scheduled_with__cron_expression}`,
+					);
+				}
+			}
+
+			// Assert that the provided start time is far enough in the future
+			if (request.values.is_scheduled_to_execute_on__time != null) {
+				const now = new Date(new Date().getTime() + tasksEnv.queueIntervalMS);
```
Page-, line 120:
```suggestion
				const now = new Date(Date.now() + tasksEnv.queueIntervalMS);
```

src/tasks/index.ts
```diff
@@ -0,0 +1,174 @@
...
+				request.values.is_scheduled_with__cron_expression != null &&
+				request.values.is_scheduled_to_execute_on__time == null
+			) {
+				try {
+					request.values.is_scheduled_to_execute_on__time = cronParser
+						.parseExpression(request.values.is_scheduled_with__cron_expression)
+						.next()
+						.toDate()
+						.toISOString();
+				} catch (_) {
```
Page-, line 111:
```suggestion
				} catch {
```

src/tasks/index.ts
```diff
@@ -0,0 +1,174 @@
...
+
+	// Create trigger function if it doesn't exist
+	await createTrigger(tx);
+
+	// Create indexes if they don't exist
+	await createIndexes(tx);
+
+	const client = new PinejsClient({
+		apiPrefix: `/${apiRoot}/`,
+	});
```
Page-, line 78:
```suggestion
	const client = sbvrUtils.api[apiRoot];
```

src/tasks/index.ts
```diff
@@ -0,0 +1,174 @@
...
+	// Async task functionality is only supported on Postgres
+	if (db.engine !== 'postgres') {
+		return;
+	}
+
+	// Create trigger function if it doesn't exist
+	await createTrigger(tx);
+
+	// Create indexes if they don't exist
+	await createIndexes(tx);
```
Page-, line 74: This should be able to be done by defining and `initSql` in the model definition the same way as user models would

src/tasks/index.ts
```diff
@@ -0,0 +1,174 @@
...
+
+// Register a task handler
+export function addTaskHandler(
+	name: string,
+	fn: TaskHandler['fn'],
+	schema?: Schema,
+): void {
+	if (worker == null) {
+		return;
+	}
```
Page-, line 164: I assume this is because we only support tasks on postgres, which is fine but should throw an error here rather than silently/surprisingly doing nothing

src/tasks/index.ts
```diff
@@ -0,0 +1,174 @@
...
+
+export const config = {
+	models: [
+		{
+			modelName: apiRoot,
+			apiRoot,
+			modelText,
+			customServerCode: exports,
+		},
+	] as sbvrUtils.ExecutableModel[],
```
Page-, line 26: `satisfies` is better because it doesn't cast in the same way
```suggestion
	] satisfies sbvrUtils.ExecutableModel[],
```
although I think it'd be better to use
```ts
export const config ConfigLoader.Config = {
```
so that the whole thing is typed and matches the expectation

src/tasks/common.ts
```diff
@@ -0,0 +1,14 @@
+import Ajv from 'ajv';
+
+// Root path for the tasks API
+export const apiRoot = 'tasks';
+
+// Channel name for task insert notifications
+export const channel = 'task_insert';
```
Page-, line 7: Can this be something like `pinejs$task_insert` to try to namespace it more?

src/database-layer/db.ts
```diff
@@ -689,6 +696,22 @@ if (maybePg != null) {
...
+				if (name === 'notification' && options?.channel === undefined) {
+					throw new Error('Missing channel option for notification listener');
+				}
+
+				const client = await pool.connect();
+				client.on(name, (msg) => {
+					fn(msg).catch((error) => {
+						console.error('Error handling message:', error);
+					});
+				});
```
Page-, line 709: I'd prefer try/catch with await so we can handle both sync/async exceptions, just as additional safety
```suggestion
				client.on(name, async (msg) => {
					try {
					 await fn(msg);
					 } catch (error) {
						console.error('Error handling message:', error);
					}
				});
```

src/tasks/worker.ts
joshbwlng: @Page- My mistake here was `* 1000`, not sure how I missed it. I didn't come up with the formula myself, I pulled it from graphile/worker: https://worker.graphile.org/docs/exponential-backoff

src/tasks/worker.ts
joshbwlng: @Page- The reasoning was probably to handle cases in which the transaction would be necessary, but I see your point and agree. We could accept a handler-level boolean option to determine if the task runner should create a transaction or not. Or just let the callers execute their logic within a transaction if they want and the task runner stays out of it. One reason I could see for creating transactions at the task runner level would be to reduce common duplicate code (a lot of tasks would want to execute within transactions).

src/tasks/worker.ts
```diff
@@ -0,0 +1,281 @@
...
+// notifications and polls the database for tasks to execute. It will execute
+// tasks in parallel up to a certain concurrency limit.
+export class Worker {
+	public handlers: Record<string, TaskHandler> = {};
+	private readonly concurrency: number;
+	private readonly interval: number;
+	private client: PinejsClient;
+	private executing = 0;
+
+	constructor(client: PinejsClient) {
```
otaviojacobi, line 36: I think this can become: `constructor(private readonly client: PinejeClient)` and then you don't need the assigning nor the declaration above. If you think it is too implicit what it does, we might just want to add the `private readonly` on the client, as it is never reassigned

src/tasks/worker.ts
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	// Check if instance can execute more tasks
+	private canExecute(): boolean {
+		return (
+			this.executing < this.concurrency && Object.keys(this.handlers).length > 0
+		);
+	}
+
+	// Execute a task
```
otaviojacobi, line 49: I don't think this comment is necessary, the function name/scope is quite clear :)

src/tasks/types.ts
```diff
@@ -0,0 +1,58 @@
...
+import type { PinejsClient } from '../sbvr-api/sbvr-utils';
+
+export const taskStatuses = [
+	'queued',
+	'cancelled',
+	'succeeded',
+	'failed',
+] as const;
+export type TaskStatus = (typeof taskStatuses)[number];
+export interface Task {
```
otaviojacobi, line 13: Just a small reminder most of this can go aways/be auto generated once we merge: https://github.com/balena-io/pinejs/pull/763

src/tasks/worker.ts
Page-: I would lean towards letting the task runner handle it and then if it's commonplace enough we can add sugar for making it easier. Basically I want solid foundations in the pinejs side that we can then extend/improve based on feedback from actual usage without preemptively adding solutions for problems that end up not existing (or needing a different solution), eg for this particular case it assumes a very narrow use case for tasks (ones that do db calls, and only db calls) and will work poorly for any that need to do stuff outside of a transaction (eg fetch some data in a transaction, close transaction, remote sync that data)

src/tasks/worker.ts
```diff
@@ -0,0 +1,278 @@
...
+				const handlerNames = Object.keys(this.handlers);
+				const binds = handlerNames
+					.map((_, index) => `$${index + 1}`)
+					.join(', ');
+				if (!this.canExecute()) {
+					return;
+				}
+				await sbvrUtils.db.transaction(async (tx) => {
+					const result = await tx.executeSql(
+						`SELECT ${selectColumns}
```
Page-, line 208: If this could be built up via abstractSql instead that would be good but is not a blocker (I'm happy for that to wait)

src/tasks/worker.ts
```diff
@@ -0,0 +1,278 @@
...
+							t."attempt count" <= t."attempt limit" AND
+							(
+								t."is scheduled to execute on-time" IS NULL OR
+								t."is scheduled to execute on-time" <= CURRENT_TIMESTAMP + INTERVAL '${Math.ceil(this.interval / 1000)} second'
+							)
+						ORDER BY
+							t."is scheduled to execute on-time" ASC,
+							t."priority" DESC,
+							t."id" ASC
+						LIMIT ${Math.max(this.concurrency - this.executing, 0)}
```
Page-, line 222: Can you make this use a bind variable rather than embedding varables, embedding vars scares me

src/tasks/worker.ts
```diff
@@ -0,0 +1,278 @@
...
+					const result = await tx.executeSql(
+						`SELECT ${selectColumns}
+						FROM task AS t
+						WHERE
+							t."is executed by-handler" IN (${binds}) AND
+							t."status" = 'queued' AND
+							t."attempt count" <= t."attempt limit" AND
+							(
+								t."is scheduled to execute on-time" IS NULL OR
+								t."is scheduled to execute on-time" <= CURRENT_TIMESTAMP + INTERVAL '${Math.ceil(this.interval / 1000)} second'
```
Page-, line 216: Can you make this use a bind variable rather than embedding varables, embedding vars scares me

src/tasks/worker.ts
```diff
@@ -0,0 +1,278 @@
...
+		void (async () => {
+			try {
+				const handlerNames = Object.keys(this.handlers);
+				const binds = handlerNames
+					.map((_, index) => `$${index + 1}`)
+					.join(', ');
+				if (!this.canExecute()) {
+					return;
+				}
+				await sbvrUtils.db.transaction(async (tx) => {
```
Page-, line 206: This transaction is held open whilst the tasks are being executed which could definitely cause issues if the tasks take any real amount of time, since that would mean the transaction times out/fails - can test what happens for a task that essentially does something like `await setTimeout(65000)`

src/database-layer/db.ts
```diff
@@ -689,6 +696,24 @@ if (maybePg != null) {
...
+				client.on(name, async (msg) => {
+					try {
+						await fn(msg);
+					} catch (error) {
+						console.error('Error handling message:', error);
+					}
+				});
+
+				if (name === 'notification' && options?.channel !== undefined) {
+					await client.query(`LISTEN "${options.channel}";`);
```
Page-, line 714: Can this use a bind var?

src/tasks/common.ts
```diff
@@ -0,0 +1,14 @@
...
+
+// Root path for the tasks API
+export const apiRoot = 'tasks';
+
+// Channel name for task insert notifications
+export const channel = 'pinejs$task_insert';
+
+// Setting inlineRefs=false as without it we run into a
+// "Maximum call stack size exceeded" error apprarently caused
+// by String.prototype._uncountable_words being set in sbvr-parser?
```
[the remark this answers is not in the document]
Page-: https://github.com/balena-io-modules/sbvr-parser/pull/59

src/database-layer/db.ts
joshbwlng: I double-checked, but Postgres doesn't support parameterized queries for `LISTEN`. We could maybe use something like `pg-format` (although its quite old at this point) or Postgres' [format](https://www.postgresql.org/docs/current/functions-string.html#FUNCTIONS-STRING-FORMAT) with `%I` I guess.

src/tasks/worker.ts
joshbwlng: Updated

src/tasks/worker.ts
joshbwlng: Updated

src/tasks/common.ts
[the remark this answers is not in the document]
joshbwlng: Updated sbvr-parser to 1.4.5 and removed inlineRefs option, things look fine now. Not planning on updating sbvr-parser to 1.4.5 with this PR, but will rebase once renovate does the bump for me (although renovate is probably being rate limited so will check the dependency dashboard).

src/tasks/index.ts
```diff
@@ -0,0 +1,158 @@
...
+				throw new BadRequestError(
+					'Creating tasks with missing actor on req is not allowed',
+				);
+			}
+
+			// Set defaults
+			request.values.status = 'queued';
+			request.values.attempt_count = 0;
+			request.values.attempt_limit ??= 1;
+			// TODO: Implement a balancer to better enqueue tasks based on actor usage
```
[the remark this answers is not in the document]
joshbwlng: Removed priority support

src/tasks/index.ts
```diff
@@ -0,0 +1,158 @@
...
+			if (handler == null) {
+				throw new BadRequestError(
+					`No task handler with name '${handlerName}' registered`,
+				);
+			}
+
+			// Assert that the provided parameter set is valid
+			if (handler.validate != null) {
+				if (!handler.validate(request.values.is_executed_with__parameter_set)) {
+					throw new BadRequestError(
```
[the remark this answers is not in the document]
joshbwlng: No, I know its going back to the caller. I thought it more useful than `throw new Error()` as that results in a 404. At least in this case the statusCode and statusMessage could be more helpful and less misleading.

test/fixtures/08-tasks/task-handlers.ts
otaviojacobi: FWIW I have used zod for quite a few things and altough I like the library a lot I think json sechema better serves in here as zod is a bulkier library that can do much more, and we would probably want to leverage all the zod type validation on several more places

src/tasks/worker.ts
```diff
@@ -0,0 +1,276 @@
...
+			// Get specified handler
+			const handler = this.handlers[task.is_executed_by__handler];
+			const startedOnTime = new Date();
+			if (handler == null) {
+				await this.finalize(
+					tx,
+					task,
+					startedOnTime,
+					'failed',
+					'Matching task handler not found',
```
dfunckt, line 58: I would rather we made this impossible to happen by ensuring code cannot enqueue work without a corresponding handler. That wouldn’t help with outstanding tasks with a deploy that removes handlers in between, but that could also be solved by having a consistency check when initializing the task system at startup to exit immediately if tasks with no handler are found.

My overarching point is we should treat this case as a programmer error, rather than an error we can recover from (by ignoring it, like we do now).

src/tasks/worker.ts
```diff
@@ -0,0 +1,276 @@
...
+							...handlerNames,
+							Math.ceil(this.interval / 1000),
+							Math.max(this.concurrency - this.executing, 0),
+						],
+					);
+					if (result.rows.length === 0) {
+						return;
+					}
+
+					// Tasks found, execute them in parallel
```
dfunckt, line 228: What’s the reasoning behind having one global worker that executes several tasks concurrently instead of several concurrent workers that poll and execute one task each?

At the expense of N poll queries instead of just one (where N=number of workers) you get much simpler code that doesn’t need to juggle multiple things.

joshbwlng: @dfunckt I think you're raising the same point Page did here: https://github.com/balena-io/pinejs/pull/739#discussion_r1644658715

I completely agree, and will look at implementing this once I have some more time

src/tasks/worker.ts
joshbwlng: We do have a check in the task creation hook to ensure the requested handler exists. I like the idea of also exiting early with an error if there are pending tasks scheduled to be executed with handlers that no longer exist. Although it would be an extra step during init with a possibly costly query (depending on the number of pending tasks we have).

src/sbvr-api/sbvr-utils.ts
```diff
@@ -783,6 +785,9 @@ export const postExecuteModels = async (tx: Db.Tx): Promise<void> => {
 			await setExecutedMigrations(tx, modelKey, pendingToSetExecutedMigrations);
 		}
 	}
+
+	// Initialize task worker and create required hooks
+	await tasks.setup(db);
```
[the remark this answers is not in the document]
joshbwlng: After some debugging I realized that this line is entirely unnecessary as the config loader is executing my setup function anyway. Didn't realize that was the case when I initially wrote this.

src/tasks/worker.ts
joshbwlng: @Page- In this case the transaction is automatically closed according to `env.db.timeoutMS`, which makes sense. I want to support tasks taking a bit longer than normal transactions, but we should still have a limit. What do you think about adding an optional `options` object to `sbvrUtils.db.transaction()` to allow overriding the default `timeoutMS` that would eventually be passed to the `AutomaticClose` class? I'd like to do something generic to give task executions a bit more room without affecting other transactions.

src/tasks/worker.ts
joshbwlng: @Page- I wrote most of it in abstractSql at one point (still have that code locally), but can we even write `FOR UPDATE SKIP LOCKED` at the moment?

src/tasks/index.ts
```diff
@@ -0,0 +1,176 @@
...
+		},
+	});
+	addPureHook('all', apiRoot, 'task', {
+		// Convert bigints to strings in responses
+		PRERESPOND: async ({ response }) => {
+			if (typeof response.body === 'object') {
+				convertBigIntsToStrings(response.body);
+			}
+		},
+	});
```
joshbwlng, line 140: Arguably this should be done globally in `sbvr-utils.handleResponse()` to handle bigints for all resources. I'd rather do something like that that affects other resources in a separate PR though and keep this work specific to tasks.

joshbwlng: Actually, since we're converting postgres bigints from strings (via the postgres node client) to bigints (sbvr-types) and now back to strings here, we really should keep things as strings in sbvr-types. This better matches the recent changes we made for a similar problem we had with dates, and makes this extra hook I've added here unnecessary. Requires another major in sbvr-types, but its the right direction to go.

src/tasks/worker.ts
joshbwlng: Actually, this error should literally never happen. The polling SQL will only select tasks that point to handlers that the task runner is already aware of. I want to keep the check here because I'm paranoid, but I'll make the error message we store in the DB mention that this really should never happen and something is scarily broken. Will also add a comment above this check to mention the same.

src/database-layer/db.ts
joshbwlng: Will check for double quotes before the `query()` call and throw if any are found in the channel name. Something simple and fast and should be plenty for real use cases.

src/tasks/worker.ts
joshbwlng: After talking with Page in a call, we decided to go with this `options` idea, but in a separate PR that will then be a pre-requisite for this PR. We also need to check timeouts at the Postgres level in production/staging when implementing tasks at the API level.

src/tasks/worker.ts
dfunckt: Sorry, I wasn’t very clear in my comment. What I mean is it should be considered an error to have outstanding jobs in the DB without a matching handler. Otherwise, you risk having orphan work scheduled that code might depend upon it running, leading to consistency/integrity issues. Is that clearer?

joshbwlng: @dfunckt  I think you were clear before, but my response added some confusion. There are two problems here: (1) the check itself technically not being necessary as the polling SQL should never pick up a task it doesn't know the handler for and (2) a pending task existing in the queue with an unknown handler. I fully responded to (1) and only partially responded to (2). We can add a check for (2) during initialization like you mentioned, but I suspect in the future we may have instances for specific use cases (like we currently have for a couple of things in production) meaning that we may end up with an instance that only knows about the "foobar-handler" for example. But thats just speculation and I think you're idea does make sense in general, so I'll look at adding it along with probably a matching partial index.

src/tasks/worker.ts
```diff
@@ -0,0 +1,282 @@
...
+	// Update task and schedule next attempt if needed
+	private async update(
+		tx: Db.Tx,
+		task: PartialTask,
+		startedOnTime: Date,
+		status: Task['Read']['status'],
+		errorMessage?: string,
+	): Promise<void> {
+		const attemptCount = task.attempt_count + 1;
+		const body: AnyObject = {
```
Page-, line 141: This should be better as:
```suggestion
		const body: Task['Write'] = {
```
or maybe
```suggestion
		const body: Partial<Task['Write']> = {
```

src/tasks/worker.ts
Page-: Yeah, specifically I very much expect that having certain instances that only handle a single type of task is definitely a possibility, for ones that should be isolated either due to their importance (nothing should affect them) or their lack thereof (they shouldn't interact with anything else), and as such not every running pinejs instance would handle every type of task. That said I've realized that an alternative would be to still register them all but tell pinejs to only dequeue a subset for given instances

src/tasks/worker.ts
```diff
@@ -0,0 +1,282 @@
...
+// Map of column names with SBVR names used in SELECT queries
+const selectColumns = Object.entries({
+	id: 'id',
+	'is executed by-handler': 'is_executed_by__handler',
+	'is executed with-parameter set': 'is_executed_with__parameter_set',
+	'is scheduled with-cron expression': 'is_scheduled_with__cron_expression',
+	'attempt count': 'attempt_count',
+	'attempt limit': 'attempt_limit',
+	'is created by-actor': 'is_created_by__actor',
+})
```
Page-, line 48:
```suggestion
} satisfies Record<string, keyof Task['Read']>)
```

src/tasks/index.ts
```diff
@@ -0,0 +1,176 @@
...
+					);
+				}
+			}
+		},
+	});
+	addPureHook('all', apiRoot, 'task', {
+		// Convert bigints to strings in responses
+		PRERESPOND: async ({ response }) => {
+			if (typeof response.body === 'object') {
+				convertBigIntsToStrings(response.body);
```
Page-, line 137: The accidental major sbvr-types bump has been reverted so this should no longer be necessary and we also plan to change sbvr-types so it will not do the automatic conversion to bigint since it shouldn't be automatically converting to JSON incompatible types in fetchProcessing

src/sbvr-api/sbvr-utils.ts
```diff
@@ -83,6 +83,7 @@ export {
 	addPureHook,
 	addSideEffectHook,
 } from './hooks';
+export { addTaskHandler } from '../tasks';
```
Page-, line 86: I don't think sbvr-utils should be re-exporting this based on the idea that the tasks should be compatible with being a pluggable module where you could have it with full fledged functionality without being part of this repo

src/tasks/worker.ts
```diff
@@ -0,0 +1,282 @@
...
+
+	// Calculate next attempt time using exponential backoff
+	private getNextAttemptTime(attempt: number): Date | null {
+		const delay = Math.ceil(Math.exp(Math.min(10, attempt)));
+		return new Date(Date.now() + delay);
+	}
+
+	// Poll for tasks and execute them
+	// This is recursive and is spawned once per concurrency limit
+	private poll(): void {
```
Page-, line 205: This is acceptable/correct for the first release but a future optimization we might do if it turns out to be necessary is a form of auto-scaling, so if there are 2+ idle polls then we would automatically scale down to just 1 idle that is polling for new tasks and then scale back up if it found 1, so we only ever have 1 idle polling loop. Specifically I don't think that optimization should be included in the very first version as it I think it would be a bit premature and also unnecessarily complicate the code (and may also turn out to not even be an issue in the end anyway)

src/tasks/index.ts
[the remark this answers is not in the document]
joshbwlng: @Page- I found where the 404 is coming from when just throwing a normal error. Apparently we default to 404 here: https://github.com/balena-io/pinejs/blob/master/src/sbvr-api/sbvr-utils.ts#L1572

src/tasks/worker.ts
joshbwlng: @Page- Opened a PR for this here: https://github.com/balena-io/pinejs/pull/775

src/tasks/index.ts
joshbwlng: Opened a PR for this here: https://github.com/balena-io-modules/sbvr-types/pull/106

src/tasks/index.ts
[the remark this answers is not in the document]
Page-: Makes sense, check out https://github.com/balena-io/pinejs/pull/782 @joshbwlng

test/fixtures/04-translations/translations/hooks.ts
```diff
@@ -1,3 +1,3 @@
-import('./v1/hooks.js');
```
Page-, line 1: I handled these via https://github.com/balena-io/pinejs/pull/790 instead as that approach makes more sense imo and will also shrink this PR a bit

src/database-layer/db.ts
```diff
@@ -710,6 +717,29 @@ if (maybePg != null) {
...
+					}
+				});
+
+				if (name === 'notification' && options?.channel !== undefined) {
+					if (options.channel.includes('"')) {
+						throw new Error(
+							`Invalid channel name for task LISTEN: ${options.channel}`,
+						);
+					}
+					await client.query(`LISTEN "${options.channel}";`);
```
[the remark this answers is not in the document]
joshbwlng: So I tested having multiple `sbvrUtils.db.on('notification')` listeners for the same `channel` and each listener's callback was executed on new task record inserts. This is what I expected and I don't think this should cause any problems in production as the handler attempts to select the inserted task with a lock via `SELECT FOR UPDATE SKIP LOCKED`. Let me know if this isn't what you were specifically concerned about.

I also tried literally listening twice here with two `await client.query(`LISTEN "${options.channel}";`);` calls and didn't notice any difference. I was half expecting Postgres to complain, but I guess it gracefully handles multiple `LISTEN` calls for the same channel. It looked like a no-op as far as I could tell when running the tests with this `LISTEN` call duplicated.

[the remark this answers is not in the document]
Page-: OK, well that answered the question of if multiple `LISTEN`s result in duplicate notifications

src/database-layer/db.ts
```diff
@@ -710,6 +717,27 @@ if (maybePg != null) {
...
+					throw new Error(`Unsupported listener type: ${name}`);
+				} else if (options?.channel == null) {
+					throw new Error('Missing channel option for notification listener');
+				} else if (options.channel.includes('"')) {
+					throw new Error(
+						`Invalid channel name for task LISTEN: ${options.channel}`,
+					);
+				}
+
+				const client = await pool.connect();
```
Page-, line 731: This will take a client from the pool for every notification listener that is added which could potentially very quickly exhaust the pool, I think a better approach would be to use
```suggestion
				listenerClient ??= await pool.connect();
```
with a `let listenerClient;` in the closure scope above, so that it's possible to reuse a single listener client. Along with adding an `if (msg.channel !== options.channel) return;` to the callback

Page-, line 731: Have you tested what happens if the connection drops? eg due to a db restart/failover?

joshbwlng: It seems that the client is released and a reconnect is not attempted. I'll look at adding some bits to attempt automatic reconnects.

src/database-layer/db.ts
```diff
@@ -707,9 +714,55 @@ if (maybePg != null) {
...
+		let listenerClient: Pg.PoolClient | null = null;
+		async function listen(
+			channel: string,
+			fn: (...args: any[]) => Promise<void>,
+		) {
+			while (listenerClient == null) {
+				try {
+					listenerClient = await pool.connect();
+					listenerClient.on('end', async () => {
+						listenerClient = null;
```
Page-, line 727: I think you need to use `listenerClient.release()` to return it to the pool (probably supplying the error object to have the pool terminate/reconnect it via `listenerClient.release(err)`) otherwise I think there could still be pool exhaustion issues

src/database-layer/db.ts
```diff
@@ -707,9 +714,55 @@ if (maybePg != null) {
...
 		}
+
+		let listenerClient: Pg.PoolClient | null = null;
+		async function listen(
+			channel: string,
+			fn: (...args: any[]) => Promise<void>,
+		) {
+			while (listenerClient == null) {
+				try {
+					listenerClient = await pool.connect();
```
Page-, line 725: This is not safe because you could do
```
pg.on('notification', () => {}, 'foobar');
pg.on('notification', () => {}, 'foobar');
```
and have two concurrent loops attempting to connect and overwriting `listenerClient`

src/database-layer/db.ts
```diff
@@ -707,9 +714,55 @@ if (maybePg != null) {
 				`);
 			}
 		}
+
+		let listenerClient: Pg.PoolClient | null = null;
+		async function listen(
+			channel: string,
+			fn: (...args: any[]) => Promise<void>,
+		) {
+			while (listenerClient == null) {
```
Page-, line 723: This means that it's only possible to add a listener whilst `listenerClient == null`, ie just once as the first listener (unless you hit the race condition mentioned below)

src/database-layer/db.ts
joshbwlng: I did that at first, but it would always error saying that the client was already released. I can add it as a "just in case" measure though in a try catch. My understanding is that the client is already released when the end event is emitted, but I don't mind having it there just in case.

src/database-layer/db.ts
```diff
@@ -707,9 +714,65 @@ if (maybePg != null) {
...
+				setTimeout(() => {
+					void listen(channel, fn);
+				}, 1000);
+			};
+
+			try {
+				listenerClient = await pool.connect();
+				listenerClient.on('end', () => {
+					reconnect();
+				});
```
Page-, line 743:
```suggestion
				listenerClient.on('end', reconnect);
```

src/database-layer/db.ts
```diff
@@ -707,9 +714,65 @@ if (maybePg != null) {
...
+
+			// Clean up and reconnect listener
+			const reconnect = () => {
+				if (listenerClient != null) {
+					try {
+						listenerClient.release();
+					} catch (err) {
+						// Ignore listener client release errors
+					}
+				}
```
Page-, line 733:
```suggestion
				try {
					listenerClient?.release();
				} catch {
					// Ignore listener client release errors
				}
```

src/database-layer/db.ts
```diff
@@ -707,9 +714,65 @@ if (maybePg != null) {
 				`);
 			}
 		}
+
+		// Connect and listen for notifications
+		async function listen(
+			channel: string,
+			fn: (...args: any[]) => Promise<void>,
+		) {
+			let listenerClient: Pg.PoolClient | null = null;
```
Page-, line 723: This still needs to live outside the `listen` function so we share the client across notification listeners rather than using a separate client from the pool for each listener

Page- (approved):

PR #741, opened 2024-02-08
Add getRolePermissions

otaviojacobi:
Change-type: minor
See: https://github.com/balena-io/open-balena-api/pull/1543

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -1273,6 +1273,99 @@ export const getUserPermissions = async (
 	}
 };
 
+export const getUserPermissionsForRole = async (
+	userId: number,
+	roleName: string,
+	tx?: Tx,
+): Promise<string[]> => {
+	try {
+		return await $getUserPermissionsForRole(userId, roleName, tx);
+	} catch (err: unknown) {
+		sbvrUtils.api.Auth.logger.error(
+			`Error loading role permissions for ${userId} ${roleName}`,
+			err,
+		);
+		throw err;
+	}
+};
+
+const $getUserPermissionsForRole = (() => {
+	const getUserPermissionsForRoleQuery = _.once(() =>
+		sbvrUtils.api.Auth.prepare<{ userId: number; roleName: string }>({
+			resource: 'permission',
+			passthrough: {
+				req: rootRead,
+			},
+			options: {
+				$select: 'name',
+				$filter: {
+					is_of__role: {
+						$any: {
+							$alias: 'rhp',
+							$expr: {
+								rhp: {
+									role: {
+										$any: {
+											$alias: 'r',
+											$expr: {
+												r: {
+													name: { '@': 'roleName' },
+													is_of__user: {
+														$any: {
+															$alias: 'uhr',
+															$expr: {
+																uhr: { user: { '@': 'userId' } },
+																$or: [
+																	{
+																		uhr: { expiry_date: null },
+																	},
+																	{
+																		uhr: {
+																			expiry_date: { $gt: { $now: null } },
+																		},
+																	},
+																],
+															},
+														},
+													},
+												},
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+				// We orderby to increase the hit rate for the `_checkPermissions` memoisation
+				$orderby: {
+					name: 'asc',
+				},
+			},
+		}),
+	);
+
+	return env.createCache(
+		'rolePermissions',
+		async (userId: number, roleName: string, tx?: Tx) => {
+			const permissions = (await getUserPermissionsForRoleQuery()(
+				{
+					userId,
+					roleName,
+				},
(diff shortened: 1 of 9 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -1273,6 +1273,102 @@ export const getUserPermissions = async (
...
 
+export const getRolePermissions = async (
+	roleName: string,
+	userId: number,
+	tx?: Tx,
+): Promise<string[]> => {
+	try {
+		if (typeof userId === 'string') {
+			userId = parseInt(userId, 10);
+		}
```
joshbwlng, line 1284: `userId` is typed as a number but you're also expecting a string? If so, I'd update the parameter type to reflect that or remove the string check + parseInt if you know you're getting a number.

src/sbvr-api/permissions.ts
```diff
@@ -1273,6 +1273,102 @@ export const getUserPermissions = async (
...
+		if (!Number.isFinite(userId)) {
+			throw new Error(`User ID has to be numeric, got: ${typeof userId}`);
+		}
+		return await $getRolePermissions(roleName, userId, tx);
+	} catch (err: any) {
```
joshbwlng, line 1289:
```suggestion
	} catch (err: unknown) {
```
Technically I think its more type safe to use unknown instead of any in this case.

src/sbvr-api/permissions.ts
otaviojacobi: This was mostly just mimic from https://github.com/balena-io/pinejs/blob/master/src/sbvr-api/permissions.ts#L1134-L1148 but I agree, we should have the correct typing. My understanding is that this will always be numeric so I will remove the check actually

src/sbvr-api/permissions.ts
```diff
@@ -1273,6 +1273,96 @@ export const getUserPermissions = async (
...
+			)) as Array<{ name: string }>;
+			return permissions.map((permission) => permission.name);
+		},
+		{
+			primitive: true,
+			promise: true,
+			normalizer: ([roleName, userId]) => `${roleName}${userId}`,
+		},
+	);
+})();
```
fisehara, line 1364: I think that the role permissions are independent of the user itself?
If the role permissions can be pulled from the DB without a userId, we should only get them without a user.
Moreover, we should not cache rolePermissions on user+roleName level, but rather on roleName.
Then the user should resolve the cached values and see if the userId has the role and return it.

fisehara (commented):
Please add test cases to understand the usage.

src/sbvr-api/permissions.ts
otaviojacobi: @fisehara sorry - this is not the idea here, the idea is really to grant the role's permission IF the user has this role. This only allows you to generate a token with a smaller scope for all the roles you already have. E.g. if you have 3 roles, instead of creating a token with all the permissions granted by the 3 roles it allows you to create a token with just the permissions of one of these roles - It never gives you more permissions than you already have

otaviojacobi: @fisehara I see. Wouldn't that be worse as it requires two DB roundtrips instead of one?

src/sbvr-api/permissions.ts
```diff
@@ -1273,6 +1273,99 @@ export const getUserPermissions = async (
...
+				},
+				undefined,
+				{ tx },
+			)) as Array<{ name: string }>;
+			return permissions.map((permission) => permission.name);
+		},
+		{
+			primitive: true,
+			promise: true,
+			normalizer: ([userId, roleName]) => `${userId}${roleName}`,
```
dfunckt, line 1364: Nit for clarity:

```suggestion
			normalizer: ([userId, roleName]) => `${userId}-${roleName}`,
```

or

```suggestion
			normalizer: ([userId, roleName]) => `${userId}/${roleName}`,
```

dfunckt, line 1364: It seems we’re caching the same exact role data (which is hundreds of bytes) *per user* which feels extremely wasteful. It’s likely we’re doing that same thing with the other permissions methods, so you shouldn’t feel obliged to change this now, but we should deal with this at some point.

dfunckt (commented):
Good one 👍

src/sbvr-api/permissions.ts
Page-: balena-cloud already uses a distributed cache for permissions using the mechanisms pinejs exposes which handles this, it'll just need to have an entry added to handle the `rolePermissions` from `rolePermissions: false as CacheOpts,` above as well as the others it already does

dfunckt: Every user on balenaCloud has the default-user role which has a certain set of permissions. What I’m saying is not that every api instance has its own cache of these permissions but that we’re caching the exact set of permissions once for each user.

PR #745, opened 2024-03-05
Update @balena/sbvr-types major

fisehara:
Update `@balena/sbvr-types` from 6.1.1 to 7.0.1

pinejs provides an internal and external interface. The external interface returns ISODateString encoded data. The internal interface should be consistent to the external interface, so that the pinejs api internally used also returns an ISODateString encoded data. This is solvedi n `sbvr-types` and is a major in the submodule and made this also a major change for pinejs itself.

https://github.com/balena-io-modules/sbvr-types/pull/91

Change-type: major


joshbwlng:
@fisehara I looked through the rest of the code base and didn't see anything else that would need to be updated, but I'll let Page also take a look before approving since this is a major. I did notice a small typo in your commit though (`This is solvedi n` -> `This is solved in`).

Page- (approved):

PR #746, opened 2024-03-12
Avoid non-blob webresources fields on multipart requests

otaviojacobi:
Change-type: patch


src/webresource-handler/index.ts
```diff
@@ -124,6 +121,15 @@ export const getUploaderMiddlware = (
 		const bb = busboy({ headers: req.headers });
 		let isAborting = false;
 
+		const parsedOdataRequest = uriParser.parseOData({
```
otaviojacobi, line 124: Moved to avoid parsing the request twice for webresources

src/webresource-handler/index.ts
```diff
@@ -151,7 +157,9 @@ export const getUploaderMiddlware = (
 			completeUploads.push(
 				(async () => {
 					try {
-						if (!(await isFileInValidPath(fieldname, req))) {
+						if (
+							!(await isFileInValidPath(fieldname, req, parsedOdataRequest))
```
Page-, line 161: As a later improvement there's a lot of work done in `isFileInValidPath` that would be nice to avoid duplicating if there are multiple uploads in a single request. I don't know if the duplication is avoidable but I suspect so

src/webresource-handler/index.ts
```diff
@@ -182,7 +190,14 @@ export const getUploaderMiddlware = (
...
 		// This is a similar behavior as previous multer library did
-		bb.on('field', (name, val) => {
+		bb.on('field', async (name, val) => {
+			if (webResourcesFieldNames.includes(name)) {
+				isAborting = true;
+				bb.emit(
+					'error',
+					new errors.BadRequestError('WebResource field must be a blob.'),
+				);
+			}
```
Page-, line 200:
```suggestion
				bb.emit(
					'error',
					new errors.BadRequestError('WebResource field must be a blob.'),
				);
				return;
			}
```

src/webresource-handler/index.ts
```diff
@@ -67,6 +67,7 @@ export const getWebresourceHandler = (): WebResourceHandler | undefined => {
 const isFileInValidPath = async (
 	fieldname: string,
 	req: Express.Request,
+	odataRequest: uriParser.ParsedODataRequest,
 ): Promise<boolean> => {
 	if (req.method !== 'POST' && req.method !== 'PATCH') {
```
Page-, line 72: Do we allow multipart requests on requests that aren't POST/PATCH?

otaviojacobi: @Page- We don't do anything, basically. Whatever is the content of multpart/body in those requests is just dropped. IMHO it is the expected behaviour as it is the same thing of a GET request with a body (it works, but the body is just ignored as it should not be there).

src/webresource-handler/index.ts
otaviojacobi: done

src/webresource-handler/index.ts
otaviojacobi: Added in my todo list for webresources

src/webresource-handler/index.ts
```diff
@@ -182,7 +190,15 @@ export const getUploaderMiddlware = (
 		// multipart requests will have two main parts, the file contents and the form fields
 		// This receives the form fields and transforms them into a standard JSON body
 		// This is a similar behavior as previous multer library did
-		bb.on('field', (name, val) => {
+		bb.on('field', async (name, val) => {
```
joshbwlng, line 193:
```suggestion
		bb.on('field', (name, val) => {
```
Nit: I don't think the `async` notation is necessary here

src/webresource-handler/index.ts
otaviojacobi: done

src/webresource-handler/index.ts
```diff
@@ -63,30 +63,27 @@ export const setupWebresourceHandler = (handler: WebResourceHandler): void => {
...
 };
+const notValidUpload = () => Promise.resolve(false);
 
-const isFileInValidPath = async (
-	fieldname: string,
+const getRequestUploadValidator = async (
 	req: Express.Request,
-): Promise<boolean> => {
+	odataRequest: uriParser.ParsedODataRequest,
+): Promise<(fieldName: string) => Promise<boolean>> => {
```
Page-, line 71: Can you make this
```suggestion
): Promise<(fieldName: string) => Resolvable<boolean>> => {
```
and use
```ts
const notValidUpload = () => false;
```
instead please

otaviojacobi: done

Page- (approved):

PR #747, opened 2024-03-21
Fix `migration` and `migration status` typings.

fisehara:
Change-type: patch

### DIFF
--- a/src/migrator/utils.ts
+++ b/src/migrator/utils.ts
@@ -96,13 +96,13 @@ export class MigrationError extends TypedError {}
 
 export type MigrationStatus = {
 	migration_key: string;
-	start_time: Date;
-	last_run_time: Date | null;
+	start_time: string | undefined | null;
+	last_run_time: string | undefined | null;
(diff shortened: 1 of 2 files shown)


src/migrator/utils.ts
```diff
@@ -325,7 +325,9 @@ WHERE "migration"."model name" = ${1}`,
 	if (data == null) {
 		return [];
 	}
-	return sbvrUtils.sbvrTypes.JSON.fetchProcessing(data.executed_migrations);
+	return sbvrUtils.sbvrTypes.JSON.fetchProcessing(
+		data.executed_migrations,
+	) as string[];
```
Page-, line 330: Is this cast necessary? Can it be `satisfies` instead?

fisehara: @Page- 
No as it doesn't satisfy. 
Also a type guard isn't working with the current typings and typescript inferencing capabilities. 
Left a comment and leave it as a cast.

thgreasi (approved):

PR #748, opened 2024-03-26
WIP: Allow file uploads via presigned URLs

otaviojacobi:
Change-type: minor


otaviojacobi:
My biggest concern currently is the fact that because we are adding mandatory parameters to the exposed WebResourceHandler interface this should be a major (or somehow modified to be only non-mandatory fields), either way this is a POC of how we could achieve large file uploads (up to 5GB on AWS).

src/webresource-handler/handlers/NoopHandler.ts
```diff
@@ -18,4 +26,43 @@ export class NoopHandler implements WebResourceHandler {
 	public async onPreRespond(webResource: WebResource): Promise<WebResource> {
 		return webResource;
 	}
+
```
otaviojacobi, line 29: I really dislike this file as a whole, I am considering just dropping it as then treating all the cases where webResourceHandler is undefined properly

src/webresource-handler/handlers/S3Handler.ts
```diff
@@ -122,6 +131,83 @@ export class S3Handler implements WebResourceHandler {
...
+		// Implement JWT generation here
+		return jsonwebtoken.sign(
+			{
+				fileKey,
+				uploadParameters,
+				metadata,
+			},
+			this.jwtSigningKey,
+			{
+				expiresIn: this.signedUrlExpireTimeSeconds,
```
otaviojacobi, line 206: I am not sure if using the same expire timeout for both the token and the signed URL is a problem. Considering this would be slightly long (24h or so) timeouts I suppose it is fine

test/06-webresource.test.ts
```diff
@@ -1087,6 +1088,59 @@ describe('06 webresources tests', function () {
 			});
 		},
 	);
+
+	describe('presigned upload', () => {
```
otaviojacobi, line 1092: The tests can use a good rework here. This is just one simple test to see that the concept works, if the broader backend team agrees on the approach they will be far more sophisticated

test/fixtures/06-webresource/config.ts
```diff
@@ -14,6 +14,7 @@ const s3Handler: WebResourceHandler = new S3Handler({
 	accessKey: requiredVar('S3_ACCESS_KEY'),
 	secretKey: requiredVar('S3_SECRET_KEY'),
 	endpoint: requiredVar('S3_ENDPOINT'),
+	jwtSigningKey: requiredVar('S3_JWT_SIGNING_KEY'),
```
otaviojacobi, line 17: Breaking change I suppose? Should this upload via presign entire flow become optional?

src/webresource-handler/index.ts
```diff
@@ -216,6 +249,136 @@ export const getUploaderMiddlware = (
...
+	}
+
+	const decodedPayload = await handler.decodeUploadToken(token);
+	const hasPermissions = await canAccess(decodedPayload.uploadParameters);
+	if (!hasPermissions) {
+		return next(new errors.ForbiddenError());
+	}
+	// Note that this has to be done in this order, first commit on the storage and then on the DB
+	// to ensure that we will only store metadata if the file is actually stored
+	const webresource = await handler.commitUpload(decodedPayload);
```
dfunckt, line 361: You may have to provisionally update the db first in order to prevent overlapping operations (how would the system behave with multiple concurrent calls to “commit”?), then move the file, then finally update the db to finalize the upload.

src/webresource-handler/handlers/NoopHandler.ts
dfunckt: What is the purpose of the noop handler?

otaviojacobi: The sole purpose of this file is this line: https://github.com/balena-io/pinejs/blob/master/src/webresource-handler/handlers/NoopHandler.ts#L7

We must resume the file stream in order for the request to be able to move on. I am considering just dropping this thing and then checking if webResource is null, manually call the `.resume()`

src/webresource-handler/index.ts
otaviojacobi: I am not sure this is needed @dfunckt as it adds some (imo unnecessary) state to the DB.
In the case of multiple concurrent calls to commit, as long as one successfully copy the file and then delete it (and even if multiple copies one over the other, this is the same file being copied) after any succeeds to delete the staging file, the subsequent commits will fail. Or even if the user re-uploads to the presigned URL while commiting, at the end we would still not have failed data integrity as the metadata stored on DB would be pointing to the latest upl

src/webresource-handler/handlers/S3Handler.ts
```diff
@@ -122,6 +131,145 @@ export class S3Handler implements WebResourceHandler {
...
+		return { token, uploadUrls };
+	}
+
+	public async decodeUploadToken(
+		token: string,
+	): Promise<MultipartUploadTokenPayload> {
+		return jsonwebtoken.verify(
+			token,
+			this.jwtSigningKey,
+		) as MultipartUploadTokenPayload;
```
otaviojacobi, line 177: Result of `verify` is `string | JwtPayload` where JwtPayload is basically `any` so we need to cast as this does not satisfies MultipartUploadTokenPayload

dfunckt: How do you know it’s not a string?

otaviojacobi: @dfunckt The only place a valid token should be generate is above, where we generate from an object that has the correct `MultipartUploadTokenPayload` type. If verify works (e.g. the token was signed with a valid key) then it must have come from a MultipartUploadTokenPayload.

dfunckt: So, because of signing, right?

Can you please leave a short comment in the code to avoid causing a pause to think about it when someone reads that bit in the future?

PR #753, opened 2024-04-19
V17

otaviojacobi:
* Requires minimum node 18.18
* Remove S3 WebResourceHandler from core pinejs in favor of: https://github.com/balena-io-modules/pinejs-webresource-s3
* Adds mandatory beginMultipartUpload/commitMultipartUpload for WebResourceHandler

Change-type: major


src/webresource-handler/index.ts
```diff
@@ -36,15 +35,6 @@ export interface WebResourceHandler {
 	onPreRespond: (webResource: WebResource) => Promise<WebResource>;
 }
 
-export class WebResourceError extends TypedError {}
```
otaviojacobi, line 39: This is gone because it doesn't really belong here. Each handler needs to be able to tell its own error typing independent really & pinejs should just forward them

src/webresource-handler/index.ts
```diff
@@ -193,17 +183,12 @@ export const getUploaderMiddlware = (
...
-					);
-				}
-
-				getLogger(getApiRoot(req)).error('Error uploading file', err);
-				next(err);
+				getLogger(getApiRoot(req)).warn('Error uploading file', err);
+				return sbvrUtils.handleHttpErrors(
+					req,
+					res,
+					new errors.BadRequestError(err),
```
otaviojacobi, line 190: I am slightly conflcited with this one. At one side I belive that the handlers should only use pine as a dev dependency for implementing proper types. On the other hand, not everything is a BadRequestError. I would accept suggestion/ideas here.

Page-: I think it's perfectly fine for handlers to depend on specific versions of pinejs, expected even, so them throwing specific http errors is fine by me. The only thing I would say is that they should probably use a peer dependency for it

PR #757, opened 2024-05-07
Deprecate LoggingClient in favor of a separate logger dictionary

Page-:
Change-type: patch

### DIFF
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -676,18 +676,19 @@ export const executeModels = async (
 				api[apiRoot].logger = { ...console };
 				if (model.logging != null) {
 					const defaultSetting = model.logging?.default ?? true;
-					const { logger } = api[apiRoot];
+					const { logger: log } = api[apiRoot];
 					for (const k of Object.keys(model.logging)) {
 						const key = k as keyof Console;
 						if (
 							key !== 'Console' &&
-							typeof logger[key] === 'function' &&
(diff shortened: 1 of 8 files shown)


src/sbvr-api/sbvr-utils.ts
```diff
@@ -676,18 +676,19 @@ export const executeModels = async (
 				api[apiRoot].logger = { ...console };
```
Page-, line 676: Avoiding this awkward monkey patching of the pinejs client instance is the main benefit, since it won't persist on a `.clone` or similar and generally makes typing life there a lot harder because it means we get the following issues:
```ts
const clone = loggingClient.clone()
clone.logger.log() // Cannot call `log` of undefined
// typeof clone !== typeof loggingClient
```
and makes typing around cloning just an unnecessary pain

joshbwlng:
Looks like somethings going wrong at runtime:
```
 Exception during run: TypeError: Cannot read properties of undefined (reading 'migrations')
    at Object.<anonymous> (/home/runner/work/pinejs/pinejs/src/migrator/async.ts:31:19)
```

Page-:
@joshbwlng it must be because I changed `(sbvrUtils.api?.migrations?.logger.info ?? console.info)(` to remove that first `?` because the variable is immediately declared so should never be able to be undefined.. I'm guessing it's because of import ordering and cyclic dependencies so I've put that `?` back and assuming it'll be fixed when moving to ESM (because it'll either just work or break until it's fixed properly)

joshbwlng (approved):

PR #759, opened 2024-06-04
17.x

Page-:


package.json
```diff
@@ -26,7 +26,7 @@
   },
   "dependencies": {
     "@balena/abstract-sql-compiler": "^9.1.4",
-    "@balena/abstract-sql-to-typescript": "^2.4.0",
+    "@balena/abstract-sql-to-typescript": "^3.0.2",
```
joshbwlng, line 29:
```suggestion
    "@balena/abstract-sql-to-typescript": "^3.1.0",
```

package.json
```diff
@@ -123,8 +123,8 @@
     "serve-static": "^1.15.0"
   },
   "engines": {
-    "node": ">=16.13.0",
-    "npm": ">=8.0.0"
+    "node": ">=20.14.0",
```
joshbwlng, line 126: `balena-api` and `open-balena-api` are at `^22.2.0`, whats the reason for staying at v20 here?

Page-: Because pinejs is a library that could be used by other things as well so we usually stick to at least supporting the LTS release. If there was a feature from the current release that we really needed then we could pin to that but otherwise sticking to LTS makes sense

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1001,8 +1003,6 @@ export type Passthrough = AnyObject & {
 };
 
 export class PinejsClient extends PinejsClientCore {
-	// @ts-expect-error This is actually assigned by `super` so it is always declared but that isn't detected here
-	public passthrough: Passthrough;
```
otaviojacobi, line 1005: I suppose this is what was breaking target: es2022?

otaviojacobi:
@Page- it looks good for me, @fisehara @thgreasi @joshbwlng is anyone planning any other major breaking change on pine that could ship with this?

src/sbvr-api/sbvr-utils.ts
Page-: Yes

joshbwlng:
I don't have any breaking changes to add, this looks good to me

otaviojacobi (approved):

PR #763, opened 2024-06-13
Model based typings

Page-:


src/data-server/sbvr-server.js
```diff
@@ -133,14 +133,15 @@ export async function setup(app, sbvrUtils, db) {
...
-				.then(async (/** @type { Array<{ [key: string]: any }> } */ result) => {
+				.then(async (result) => {
 					if (result.length === 0) {
 						throw new Error('No SE data model found');
 					}
 					const instance = result[0];
 					await sbvrUtils.executeModel(tx, {
 						apiRoot: instance.is_of__vocabulary,
-						modelText: instance.model_value.value,
+						// prettier-ignore
```
Page-, line 143: This is because we need the parens to stay intact below, I wanted to add a comment to this ignore but doing so makes prettier ignore the ignore bit.. so I couldn't

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1020,11 +1032,37 @@ export class PinejsClient extends PinejsClientCore {
 	}) {
 		return (await runURI(method, url, body, tx, req, custom)) as object;
 	}
+
+	public post<TResource extends keyof M & string>(
+		params: {
+			resource: TResource;
+			options: Params<M[TResource]>['options'] & { returnResource: false };
```
Page-, line 1039: We need the overrides in order to properly type this `returnResource: false` case. I'm thinking maybe move it into pinejs-client but if doing so I'd want to properly type the primary key (rather than the `id` hardcoding)

Page-:
Depends on #764

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1020,6 +1031,31 @@ export class PinejsClient extends PinejsClientCore {
...
+	): Promise<M[TResource]['Read']>;
+	public post<TResource extends keyof M & string>(
+		params: {
+			resource: TResource;
+			options: Params<M[TResource]>['options'] & { returnResource: boolean };
+		} & Params<M[TResource]>,
+	): Promise<Pick<M[TResource]['Read'], 'id'>>; // TODO: This should use the primary key rather than hardcoding `id`
+	/**
+	 * @deprecated POSTing via `url` is deprecated
+	 */
```
otaviojacobi, line 1050: Do we have an alternative/proposal/idea for doing `canAccess` now that posting via url is deprecated?

src/sbvr-api/hooks.ts
```diff
@@ -31,23 +30,27 @@ export interface HookReq {
...
+	api: (typeof api)[Vocab];
 	tx?: Tx | undefined;
 }
 export type HookResponse = PromiseLike<any> | null | void;
 
-export interface Hooks {
-	PREPARSE?: (options: Omit<HookArgs, 'request' | 'api'>) => HookResponse;
-	POSTPARSE?: (options: HookArgs) => HookResponse;
-	PRERUN?: (options: HookArgs & { tx: Tx }) => HookResponse;
+export interface Hooks<Vocab extends string = string> {
```
otaviojacobi, line 41: Couldn't this and the above be `export interface HookArgs<Vocab extends (keyof typeof api) = string>` ? I think all in all it is the same, but this is a bit more explicit for the reader on where the string should be coming from

src/sbvr-api/hooks.ts
```diff
@@ -31,23 +30,27 @@ export interface HookReq {
...
+	PREPARSE?: (
+		options: Omit<HookArgs<Vocab>, 'request' | 'api'>,
+	) => HookResponse;
+	POSTPARSE?: (options: HookArgs<Vocab>) => HookResponse;
+	PRERUN?: (options: HookArgs<Vocab> & { tx: Tx }) => HookResponse;
 	/** These are run in reverse translation order from newest to oldest */
-	POSTRUN?: (options: HookArgs & { tx: Tx; result: any }) => HookResponse;
+	POSTRUN?: (
+		options: HookArgs<Vocab> & { tx: Tx; result: any },
```
otaviojacobi, line 49: Couldn't result be inferred now? (not necessarily in this PR, just one thing we could keep in mind)

src/sbvr-api/hooks.ts
otaviojacobi: Maybe something like `type VocabKeys = (keyof typeof api)` and `Vocab extends VocabKeys` all over the place

src/sbvr-api/user.ts
```diff
@@ -0,0 +1,214 @@
+import type { Types } from '@balena/abstract-sql-to-typescript';
```
otaviojacobi, line 1: We could do similar as we do on balena-api to generate a line on top saying this file is auto-generated, it can avoid people mistakenly modifying stuff here

Page-: That makes a lot of sense, so much so that I think `@balena/abstract-sql-to-typescript` should do it so it's universal (and include the version used to generate the types)

Page-: https://github.com/balena-io-modules/abstract-sql-to-typescript/pull/68

src/sbvr-api/hooks.ts
Page-: In theory yes it could be that, but `keyof typeof api` is `string` anyway because of the index signature, so in practice there is no difference

src/sbvr-api/hooks.ts
Page-: Yes, although it's not related to the OData result directly: https://github.com/balena-io/pinejs/pull/765

src/sbvr-api/sbvr-utils.ts
Page-: Not yet, but this is to highlight any cases that come up so we can work on either converting them to standard ways that we've added support for or if they don't already have a way then to make sure we create one. However it would still be possible to change from
```ts
api.post({
	url,
	...
})
```
to
```
api.request({
	method: 'POST',
	url,
	...
})

otaviojacobi (approved):

PR #765, opened 2024-06-17
Improve `result` typing for `POSTRUN`/`PRERESPOND` hooks

Page-:
Change-type: patch

### DIFF
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -1541,7 +1541,8 @@ const runRequest = async (
 	if (env.DEBUG) {
 		log.log('Running', req.method, req.url);
 	}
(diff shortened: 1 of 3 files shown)


src/sbvr-api/sbvr-utils.ts
```diff
@@ -1839,26 +1843,26 @@ const runPost = async (
...
 
 	let result: AnyObject = { d: [{ id }] };
 	if (
-		location != null &&
+		id != null &&
 		!['0', 'false'].includes(
 			request?.odataQuery?.options?.returnResource as string,
 		)
 	) {
+		const location = odataResponse.resourceURI(
```
otaviojacobi, line 1861: It is used down there so that's why test fails

otaviojacobi (approved):

PR #766, opened 2024-06-19
Add multipart upload interface for handlers

otaviojacobi:
Change-type: minor


joshbwlng:
@otaviojacobi Are these new interfaces called or tested somewhere?

PR #771, opened 2024-06-19
Improve test assertion for not undefined

otaviojacobi:
Change-type: patch


thgreasi (approved):

thgreasi:
@otaviojacobi there are a few missing `$select` that you need to add as well it seems.

otaviojacobi:
> @otaviojacobi there are a few missing `$select` that you need to add as well it seems.

@thgreasi in fact this is a typing error that has to be fixed on client-core/client-supertest

PR #775, opened 2024-06-25
Add transaction timeoutMS override option

joshbwlng:
Change-type: minor

---

This is a prerequisite for https://github.com/balena-io/pinejs/pull/739 as we want to allow task transactions to stay open longer than others

### DIFF
--- a/src/database-layer/db.ts
+++ b/src/database-layer/db.ts
@@ -17,7 +17,10 @@ export interface CodedError extends Error {
 	code: number | string;
 }
 
-type CreateTransactionFn = (stackTraceErr?: Error) => Promise<Tx>;
+type CreateTransactionFn = (
+	stackTraceErr?: Error,
+	timeoutMS?: number,
+) => Promise<Tx>;
 type CloseTransactionFn = () => void;
 export interface Row {
 	[fieldName: string]: any;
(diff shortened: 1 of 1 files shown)


src/database-layer/db.ts
```diff
@@ -17,7 +17,10 @@ export interface CodedError extends Error {
 	code: number | string;
 }
 
-type CreateTransactionFn = (stackTraceErr?: Error) => Promise<Tx>;
+type CreateTransactionFn = (
+	stackTraceErr?: Error,
+	timeoutMS?: number,
+) => Promise<Tx>;
```
joshbwlng, line 23: We could also do this to avoid the parameter list on calls being unnecessarily long:
```typescript
options: {
    stackTraceErr?: Error,
    timeoutMS?: number,
}
```

src/database-layer/db.ts
```diff
@@ -218,20 +226,18 @@ class AutomaticClose {
...
 			if (this.stackTraceErr) {
 				console.error(this.stackTraceErr.stack);
 			}
 			void tx.rollback();
 		};
-		this.automaticCloseTimeout = setTimeout(
-			this.automaticClose,
-			env.db.timeoutMS,
-		);
+		this.automaticCloseTimeout = setTimeout(this.automaticClose, timeoutMS);
```
Page-, line 240: I can't comment on the relevant line but there's other uses of `env.db.timeoutMS` that also need updating

Page- (approved):

PR #776, opened 2024-06-25
Fix sbvr server typings

otaviojacobi:
I think I prefer #777 instead of this

Change-type: patch


src/data-server/sbvr-server.js
```diff
@@ -138,10 +138,12 @@ export async function setup(app, sbvrUtils, db) {
 						throw new Error('No SE data model found');
 					}
 					const instance = result[0];
+					// prettier-ignore
```
otaviojacobi, line 141: This is necessary otherwise prettier will remove the parenthesis and this inlining types into JS does not like that

src/migrator/migrations.ts
```diff
@@ -1,4 +1,4 @@
-// These types were generated by @balena/abstract-sql-to-typescript v3.2.1
+// These types were generated by @balena/abstract-sql-to-typescript v3.2.3
```
otaviojacobi, line 1: I am slightly afraid of this version here ): it reminds me of the cli that everytime it needs you to know the magic command you need to run before commiting ( and adding on a commit hook on oB/cloudAPI is not viable bcs it requires docker)

joshbwlng: I'm not sure what you mean by the version being a problem? Anyway we had something similar in Jellyfish - types generated from contract JSON schema definitions via AutumnDB. One way to enforce types being up-to-date before merge would be to generate types as a step in CI along with something like https://www.npmjs.com/package/catch-uncommitted to `exit 1` if unexpected diffs were found.

Page-: We do already enforce that they are updated, it's just that people have to actually update them. And if most of the time the actual output doesn't change it could quickly get annoying. Historically though `@balena/abstract-sql-to-typescript` gets very few updates and if we were to publish these standalone then it could be useful.. so I'm still on the side of keeping it and enduring some pain whilst all the model based typings stabilize and are merged since I expect the pain to be minimal once that work is completed

PR #777, opened 2024-06-25
Migrate sbvr-server to typescript

otaviojacobi:
Change-type: patch


src/data-server/sbvr-server.ts
```diff
@@ -411,4 +375,38 @@ export async function setup(app, sbvrUtils, db) {
 	});
 
 	await db.transaction(setupModels);
-}
+};
+
+export const config: Config = {
```
otaviojacobi, line 380: I just had to change the order config & setup variabels were defined as config uses the "setup" function and TS would complain we were defining it after having used it

joshbwlng:
I don't see any issues, but will let others also take a look before merging

src/data-server/sbvr-server.ts
Page-: That was because you changed it from being a function declaration which would be hoisted, into a function assignment to variable. Which was so it was easier to do the `SetupFunction` typing which is fine (I think it's technically possible to do with a function declaration.. but it's such a pain I also always avoid it)

Page- (approved):

PR #778, opened 2024-06-26
odata/abstract-sql-compiler: Add support for specifying the model name for config files

Page-:
This allows choosing which model to use when the config file contains multiple models

Change-type: minor

### DIFF
--- a/src/bin/utils.ts
+++ b/src/bin/utils.ts
@@ -48,8 +48,31 @@ ${rule.sql}`,
 	writeAll(output, outputFile);
 };
 
+const getConfigModel = (
+	fileContents: Model | AbstractSqlModel | Config,
+	modelName?: string,
+): Model | AbstractSqlModel => {
+	if ('models' in fileContents) {
+		if (fileContents.models.length === 0) {
+			throw new Error('No models found in config file');
+		}
+		if (modelName != null) {
+			const model = fileContents.models.find((m) => m.modelName === modelName);
+			if (model == null) {
+				throw new Error(
+					`Could not find model with name '${modelName}', found: ${fileContents.models.map((m) => m.modelName).join(', ')}`,
+				);
+			}
+			return model;
+		}
+		return fileContents.models[0];
+	}
+	return fileContents;
+};
+
 export const getAbstractSqlModelFromFile = (
 	modelFile: string,
+	modelName: string | undefined,
 ): AbstractSqlModel => {
 	let fileContents: string | Model | AbstractSqlModel | Config;
 	try {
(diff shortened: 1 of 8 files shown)


src/bin/utils.ts
```diff
@@ -68,7 +69,11 @@ export const getAbstractSqlModelFromFile = (
 			return fileContents;
 		}
 		const configModel =
-			'models' in fileContents ? fileContents.models[0] : fileContents;
+			'models' in fileContents
+				? modelName
+					? fileContents.models.find((m) => m.modelName === modelName)!
```
otaviojacobi, line 74: Can we treat the case where we passed a modelName that does not exist and throw a more specific error? Right now it throws with a very cryptic error message like:

```
        if ('abstractSql' in configModel && configModel.abstractSql != null) {
                          ^

TypeError: Cannot use 'in' operator to search for 'abstractSql' in undefined
    at getAbstractSqlModelFromFile (/home/otavio/git/balena-api/node_modules/@balena/pinejs/out/bin/utils.js:87:27)
    at Command.generateTypes (/home/otavio/git/balena-api/node_modules/@balena/pinejs/out/bin/abstract-sql-compiler.js:13:65)
    at async Command.parseAsync (/home/otavio/git/balena-api/node_modules/commander/lib/command.js:936:5)
```

src/bin/utils.ts
```diff
@@ -48,8 +48,30 @@ ${rule.sql}`,
...
+			}
+		}
+		return fileContents.models[0];
+	}
+	return fileContents;
+};
+
 export const getAbstractSqlModelFromFile = (
 	modelFile: string,
+	modelName: string | undefined,
```
otaviojacobi, line 74: `modelName?: string | undefined`

Page-: Apparently tests use this function which is why I didn't catch it (tests use a different tsconfig). I did intentionally make it required because for the cli use cases it should always be provided, but for tests it's a bit different - I think I'll make the tests bend to the better method for the real code

src/bin/utils.ts
```diff
@@ -48,8 +48,30 @@ ${rule.sql}`,
...
+		if (fileContents.models.length === 0) {
+			throw new Error('No models found in config file');
+		}
+		if (modelName != null) {
+			const model = fileContents.models.find((m) => m.modelName === modelName);
+			if (model == null) {
+				throw new Error(
+					`Could not find model with name '${modelName}', found: ${fileContents.models.map((m) => m.modelName).join(', ')}`,
+				);
+			}
```
otaviojacobi, line 65: ... and `return model;` :)

otaviojacobi, line 65: aand `return mode;`

otaviojacobi (changes requested):

otaviojacobi (approved):

PR #780, opened 2024-06-28
Use assertExists in place of expectNotUndefined

otaviojacobi:
It is a slightly improvement on the function definition (also more similar to what we use on other places) and also moving it to a common test lib instead of inside a single test module.

Change-type: patch


joshbwlng:
@otaviojacobi Hmm, the type generation comment is causing CI to fail with this in a few places:
```
-// These types were generated by @balena/abstract-sql-to-typescript v3.2.3
+// These types were generated by @balena/abstract-sql-to-typescript v3.3.0
```

It's definitely a bit annoying and makes me wonder if we should leave at least the minor and patch out of comment. Or just leave the version out entirely and only fail if the actual types (the actual important bit!) have changed.

otaviojacobi:
> @otaviojacobi Hmm, the type generation comment is causing CI to fail with this in a few places:
> 
> ```
> -// These types were generated by @balena/abstract-sql-to-typescript v3.2.3
> +// These types were generated by @balena/abstract-sql-to-typescript v3.3.0
> ```
> 
> It's definitely a bit annoying and makes me wonder if we should leave at least the minor and patch out of comment. Or just leave the version out entirely and only fail if the actual types (the actual important bit!) have changed.

Yes, I have raised the same concern but the idea is that @balena/abstract-sql-to-typescript is very rarely updated and we just have to endure a bit of updating while it becomes stable with the new typings

thgreasi (approved):

PR #781, opened 2024-06-29
Fix getIdField using idField of intermediate translation

otaviojacobi:
getIdField tries to return the name of the id field in the database model (and not in any translated model)
This turns out to be a problem when more than one layer of native translations are present for a given table as it tries to use the resource name of a given version on the final abstract sql model

This approach will use the actual translation (rather than the final one) in case the translated table exists, avoiding trying to find a non existent translation on the final model

Resolves: https://github.com/balena-io/pinejs/issues/794
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/345890-balena-io/topic/pinejs.20bump.20dependencies/near/455374721

### DIFF
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -1166,9 +1166,13 @@ const getIdField = (
 		| 'resourceName'
 		| 'vocabulary'
 	>,
-) =>
-	// TODO: Should resolveSynonym also be using the finalAbstractSqlModel?
-	getFinalAbstractSqlModel(request).tables[resolveSynonym(request)].idField;
+) => {
+	const resolvedSynonym = resolveSynonym(request);
+	const translatedTable = getAbstractSqlModel(request).tables[resolvedSynonym];
+	const resourceName =
+		translatedTable != null ? translatedTable.resourceName : resolvedSynonym;
+	return getFinalAbstractSqlModel(request).tables[resourceName].idField;
+};
 
 export const getAffectedIds = async <Vocab extends string>(
 	args: HookArgs<Vocab> & {
--- a/test/fixtures/04-translations/config.ts
+++ b/test/fixtures/04-translations/config.ts
@@ -9,6 +9,8 @@ const modelFile = __dirname + '/university.sbvr';
 import { v1AbstractSqlModel, v1Translations } from './translations/v1';
 import { v2AbstractSqlModel, v2Translations } from './translations/v2';
 import { v3AbstractSqlModel, v3Translations } from './translations/v3';
+import { v4AbstractSqlModel } from './translations/v4';
+import { v5AbstractSqlModel } from './translations/v5';
 
 export const abstractSql = getAbstractSqlModelFromFile(modelFile, undefined);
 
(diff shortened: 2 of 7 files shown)


src/sbvr-api/sbvr-utils.ts
```diff
@@ -1166,9 +1166,18 @@ const getIdField = (
...
+	const translatedTable =
+		request.abstractSqlModel?.tables[resolveSynonym(request)];
+
+	if (translatedTable == null) {
+		return getFinalAbstractSqlModel(request).tables[resolveSynonym(request)]
+			.idField;
+	}
+
+	return getFinalAbstractSqlModel(request).tables[translatedTable.resourceName]
+		.idField;
```
thgreasi, line 1179:
```suggestion
	const resolvedSynonym = resolveSynonym(request)
	const translatedTable = request.abstractSqlModel?.tables[resolvedSynonym];

	const resourceName = translatedTable != null ? translatedTable.resourceName : resolvedSynonym;
	return getFinalAbstractSqlModel(request).tables[resourceName].idField;
```

thgreasi: Or maybe
```suggestion
	const resolvedSynonym = resolveSynonym(request)
	const translatedTable = getAbstractSqlModel(request).tables[resolvedSynonym];
	
	const resourceName = translatedTable != null ? translatedTable.resourceName : resolvedSynonym;
	return getFinalAbstractSqlModel(request).tables[resourceName].idField;
```

otaviojacobi: Nice, much better, thanks

thgreasi:
I can confirm that this works w/ both oB & bC v7 changes w/o needing the `$toResource` `toVersion`->`$resin` workarounds :heart: 
See: https://github.com/balena-io/open-balena-api/pull/1684/files#diff-8c30cbe3a5e90e20f89a2916ad53db6a65d8b8afd746265c79ad0d17f78cafa0R91
See: https://github.com/balena-io/balena-api/pull/5120

otaviojacobi:
I also confirmed it works with "current" mater: 
See: https://github.com/balena-io/open-balena-api/pull/1732
See: https://github.com/balena-io/balena-api/pull/5246

I think this is a solution, I am just not sure if it is the optimal one

thgreasi:
I'm fine w/ this as is, but I would like Page to take a look to check it as well since I'm not that familiar w/ translations.
The code makes sense as I read it, so pending Page to confirm that it is also indeed the correct way.

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1166,9 +1166,13 @@ const getIdField = (
...
 		| 'vocabulary'
 	>,
-) =>
-	// TODO: Should resolveSynonym also be using the finalAbstractSqlModel?
-	getFinalAbstractSqlModel(request).tables[resolveSynonym(request)].idField;
+) => {
+	const resolvedSynonym = resolveSynonym(request);
+	const translatedTable = getAbstractSqlModel(request).tables[resolvedSynonym];
+	const resourceName =
+		translatedTable != null ? translatedTable.resourceName : resolvedSynonym;
```
Page-, line 1173: Why do we expect `translatedTable` to be null at any point?

Page-, line 1173: What happens if we have `v3TranslatedTable.resourceName == 'v4ResourceName'`, why do we not need to progressively go via each translation to find the final table?

otaviojacobi: My understanding is that this works in the case because the only place where we actually call `getIdField` is if `addReturning && request.affectedIds == null`, which means `affectedIds` hasn't been populated yet, which as far as I tested always happens on the last translation to the DB, so this case won't happen, but none (the current & PRed version) of getIdField is really generic

otaviojacobi: My understanding is that it will be `nullish` when it is in the DB model (it doesn't translate to anything)

PR #793, opened 2024-08-20
Add on notification listener support

joshbwlng:
Change-type: minor


src/database-layer/db.ts
```diff
@@ -707,9 +714,61 @@ if (maybePg != null) {
 				`);
 			}
 		}
+
+		// Connect and listen for notifications
+		async function listen(
+			channel: string,
+			fn: (...args: any[]) => Promise<void>,
+		) {
+			let listenerClient: Pg.PoolClient | null = null;
```
joshbwlng, line 723: This needs to live outside the `listen` function so we share the client across notification listeners rather than using a separate client from the pool for each listener

PR #797, opened 2024-08-28
Tests: stop using deprecated `faker` functions

Page-:
Change-type: patch

### DIFF
--- a/test/04-translations.test.ts
+++ b/test/04-translations.test.ts
@@ -30,7 +30,7 @@ describe('04 native translation tests', function () {
 				apiPrefix: 'university/',
 				resource: 'faculty',
 				body: {
-					name: faker.random.words(2),
+					name: faker.word.words(2),
 				},
 			})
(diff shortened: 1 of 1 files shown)


test/04-translations.test.ts
```diff
@@ -47,9 +47,9 @@ describe('04 native translation tests', function () {
 					apiPrefix: 'v1/',
 					resource: 'student',
 					body: {
-						matrix_number: faker.datatype.number({ min: 100000 }),
-						name: faker.name.firstName(),
-						lastname: faker.name.lastName(),
+						matrix_number: faker.number.int({ min: 100000, max: 2 ** 31 }),
```
Page-, line 50: Previously `max` defaulted to `min + 99999` and now it defaults to `Number.MAX_SAFE_INTEGER` which is bigger than a postgres int, so I changed the max to fit within a postgres int

thgreasi (approved):

PR #796, opened 2024-08-27
Add support for and prefer providing the unaliased `$toResource`

Page-:

PR #798, opened 2024-08-29
Translations: prefer translating to the unaliased resource of the target model

Page-:
This matches closer with the idea of a stage by stage translation where you translate as if you were a a request against the next model, then that translates to the next model and so on

Change-type: patch

### DIFF
--- a/src/sbvr-api/translations.ts
+++ b/src/sbvr-api/translations.ts
@@ -208,24 +208,24 @@ export const translateAbstractSqlModel = (
 		if (translationDefinition) {
 			const { $toResource, ...definition } = translationDefinition;
 			const hasToResource = typeof $toResource === 'string';
+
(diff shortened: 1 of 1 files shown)


thgreasi (approved):
LGTM, but I'm wondering whether this would work w/ https://github.com/balena-io/pinejs/pull/796 and getting rid of the direct $resin $toResource workardound we have in oB
https://github.com/balena-io/open-balena-api/blob/0fc281cf11dfc267b84f856481b5e643a5737196/src/translations/v6/v6.ts#L89-L95
A naive brief try in my side didn't work on getting rid of the $resin.

Page-:
Yeah, this serves different, albeit similar, goals to #796 so I wouldn't expect it to fix anything which that one covers (and vice versa)

PR #805, opened 2024-09-10
Update TypeScript to 5.6.2

thgreasi:
Change-type: patch


build/browser.ts
```diff
@@ -1,26 +1,38 @@
 import * as webpack from 'webpack';
 import sharedConfig from './config';
-const config = { ...sharedConfig };
```
thgreasi, line 3: Without these changes in the files under `build/`, balena-lint was failing for all of them with:
```
build/module.ts:3:7 - error TS4023: Exported variable 'config' has or is using name 'ExperimentsCommon' from external module "/home/teo/Resin.io/balena-io-modules/pinejs/node_modules/webpack/types" but cannot be named.

 const config = { ...sharedConfig };
```

The alternative "fix" was to do
```
const config = { ...sharedConfig } as typeof sharedConfig;
```
but I preferred the now more natural way of destructuring.

PR #808, opened 2024-09-13
Use updated pinejs-client `prepare` syntax

Page-:
This also fixes the returned actor id for `checkPassword` since the improved typings caught that it did not match the expected return type

Change-type: patch

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -1192,18 +1192,21 @@ const memoizedGetConstrainedModel = (
 	getBoundConstrainedMemoizer(abstractSqlModel)(permissionsLookup, vocabulary);
 
 const getCheckPasswordQuery = _.once(() =>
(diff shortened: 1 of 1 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -1220,76 +1223,76 @@ export const checkPassword = async (
...
-	const res = await sbvrUtils.sbvrTypes.Hashed.compare(password, hash);
+	const res = await sbvrUtils.sbvrTypes.Hashed.compare(password, user.password);
 	if (!res) {
 		throw new Error('Passwords do not match');
 	}
+	const userId = user.id;
 	const permissions = await getUserPermissions(userId);
 	return {
 		id: userId,
-		actor: actorId,
```
otaviojacobi, line 1233: How was that working???

Page-: Presumably nothing depended on the resultant actor id being the number it said it was

otaviojacobi (approved):

PR #814, opened 2024-09-30
Add getAuthCompiler for building permissions

otaviojacobi:
Change-type: minor

### DIFF
--- a/src/sbvr-api/sbvr-utils.ts
+++ b/src/sbvr-api/sbvr-utils.ts
@@ -36,7 +36,7 @@ import {
 } from '@balena/odata-to-abstract-sql';
 import sbvrTypes from '@balena/sbvr-types';
 import deepFreeze = require('deep-freeze');
-import type { Params } from 'pinejs-client-core';
+import type { ODataOptions, Params } from 'pinejs-client-core';
 import { PinejsClientCore, type PromiseResultTypes } from 'pinejs-client-core';
 
 import { ExtendedSBVRParser } from '../extended-sbvr-parser/extended-sbvr-parser';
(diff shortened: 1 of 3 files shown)


src/sbvr-api/compile-auth.ts
```diff
@@ -0,0 +1,53 @@
+import type { Filter, PinejsClientCore, Resource } from 'pinejs-client-core';
+
+const asPermission = (compiledFilter: string): string => {
+	return compiledFilter
+		.split('$filter=')[1]
```
otaviojacobi, line 5: TODO: more strict checks on this... probably regex match the entire filter to avoid possible issues in case somehow a $select went by or something of the sort

src/sbvr-api/compile-auth.ts
```diff
@@ -0,0 +1,51 @@
...
+		.replace(/Auth\.canAccess\(\)/g, 'canAccess()');
+};
```
Page-, line 7: You should be able to do something like:
``````suggestion
const asPermission = (url: string): string => {
	const queryParamsIndex = url.indexOf('?');
	if (queryParamsIndex === -1) {
		throw ...
	}
	const queryParams = new URLSearchParams(url.slice(queryParamsIndex));
	const compiledFilter = queryParams.get('$filter');
	if (compiledFilter == null) {
		throw ...
	}
	return compiledFilter.replace(/Auth\.canAccess\(\)/g, 'canAccess()');
};
```
to ensure proper parsing of it as query params and whatnot

src/sbvr-api/compile-auth.ts
```diff
@@ -0,0 +1,51 @@
...
+		compileAuth: <TResource extends string & keyof M>({
+			resource,
+			access,
+			options,
+		}: {
+			resource: TResource;
+			access: string;
+			options?: {
+				$filter: Filter<M[TResource]['Read']>;
+			};
```
Page-, line 34: It's probably clearer as
```suggestion
			options?: Pick<ODataOptions<M[TResource]['Read']>, '$filter'>;
```
to show that it's an ODataOptions object, but just the `$filter`

src/sbvr-api/compile-auth.ts
```diff
@@ -0,0 +1,51 @@
+import type { Filter, PinejsClientCore, Resource } from 'pinejs-client-core';
+
+const asPermission = (compiledFilter: string): string => {
+	return compiledFilter
+		.split('$filter=')[1]
+		.replace(/Auth\.canAccess\(\)/g, 'canAccess()');
+};
+
+export const getAuthCompiler = <M extends { [key in keyof M]: Resource }>({
```
Page-, line 9: I would be fine adding this to the internal pinejs-client as well so it can be used as `sbvrUtils.api.foo.compileAuth(...)`

otaviojacobi: Yeah, I considered this possibility but my point "agaisnt" it was basically that in most places we use ClientCore/PineClient interchangeably e.g. we sometimes override stuff etc but completely adding new features is something that could be confusing, that's why I went with this approach. And if I think of how I woudl use it in oB/cloudAPI I would end up writing a wrapper e.g. `const resinAuth = sbvrUtils.api.foo.compileAuth(...);` so I went with directly exposing the wrapper

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1049,6 +1049,44 @@ export class PinejsClient<
...
 	}
+	public compileAuth<TResource extends keyof M & string>(params: {
+		modelName: string;
+		access: string;
+		resource: TResource;
+		options?: Pick<ODataOptions<M[TResource]['Read']>, '$filter'>;
+	}): string {
+		const { modelName, access, resource, options } = params;
+		if (typeof modelName !== 'string') {
+			throw new Error(`The modelName property must be specified.`);
```
Page-, line 1060:
```suggestion
			throw new Error('The modelName property must be specified.');
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1049,6 +1049,44 @@ export class PinejsClient<
...
+		}
+		if (typeof access !== 'string') {
+			throw new Error(`The access property must be specified.`);
```
Page-, line 1063:
```suggestion
			throw new Error('The access property must be specified.');
```

src/sbvr-api/sbvr-utils.ts
```diff
@@ -1049,6 +1049,44 @@ export class PinejsClient<
...
+		if (typeof resource !== 'string') {
+			throw new Error('The resource must be specified.');
+		}
+
+		const authBase = `${modelName}.${resource}.${access}`;
+		const url =
+			options?.$filter != null ? this.compile({ resource, options }) : null;
+		if (url == null) {
+			return authBase;
+		}
```
Page-, line 1074: This might be slightly nicer as
```suggestion
		if (options?.$filter == null) {
			return authBase;
		}
		const url = this.compile({ resource, options });
```
assuming of course it can play nice with typings

Page- (approved):

PR #823, opened 2024-10-14
Add tasks worker `stop` method

otaviojacobi:
Change-type: minor

### DIFF
--- a/src/tasks/worker.ts
+++ b/src/tasks/worker.ts
@@ -56,6 +56,7 @@ export class Worker {
 	public handlers: Record<string, TaskHandler> = {};
 	private readonly concurrency: number;
 	private readonly interval: number;
+	private running = false;
 	private executing = 0;
 
 	constructor(private readonly client: PinejsClient) {
@@ -244,9 +245,15 @@ export class Worker {
 				if (!executed) {
 					await setTimeout(this.interval);
 				}
-				this.poll();
+				if (this.running) {
+					this.poll();
+				}
 			}
 		})();
+	}
+
+	public async stop(): Promise<void> {
+		this.running = false;
 	}
 
 	// Start polling for tasks
@@ -281,6 +288,7 @@ export class Worker {
 			);
 		}
 
+		this.running = true;
 		// Spawn children to poll for and execute tasks
 		for (let i = 0; i < this.concurrency; i++) {
 			this.poll();
--- a/test/08-tasks.test.ts
+++ b/test/08-tasks.test.ts
@@ -7,6 +7,7 @@ import { testInit, testDeInit, testLocalServer } from './lib/test-init';
 import { tasks as tasksEnv } from '../src/config-loader/env';
 import type Model from '../src/tasks/tasks';
 import * as cronParser from 'cron-parser';
+import { PINE_TEST_SIGNALS } from './lib/common';
 
 const actorId = 1;
 const fixturesBasePath = __dirname + '/fixtures/08-tasks/';
@@ -407,6 +408,66 @@ describe('08 task tests', function () {
 					},
 				})
 				.expect(500);
+		});
+
+		it('should not pick up tasks while worker is not running', async () => {
+			pineServer.send(PINE_TEST_SIGNALS.STOP_TASK_WORKER);
+
+			const name = randomUUID();
+			const task = await createTask(pineTest, apikey, {
+				is_executed_by__handler: 'create_device',
+				is_executed_with__parameter_set: {
+					name,
+					type: randomUUID(),
+				},
+			});
+
+			// Await more than the queue interval to ensure the task is not picked up
+			await setTimeout(2 * tasksEnv.queueIntervalMS);
+
+			await expectTask(pineTest, task.id, {
+				status: 'queued',
+				error_message: null,
+				started_on__time: null,
+				ended_on__time: null,
+				attempt_count: 0,
+			});
+
+			const { body: devices } = await pineTest
+				.get({
+					apiPrefix: 'example/',
+					resource: 'device',
+					options: {
+						$filter: {
+							name,
+						},
+					},
+				})
+				.expect(200);
+			expect(devices.length).to.equal(0);
+
+			pineServer.send(PINE_TEST_SIGNALS.START_TASK_WORKER);
+
+			await waitFor(async () => {
+				const { body: $devices } = await pineTest
+					.get({
+						apiPrefix: 'example/',
+						resource: 'device',
+						options: {
+							$filter: {
+								name,
+							},
+						},
+					})
+					.expect(200);
+				return $devices.length === 1;
+			});
+
+			await expectTask(pineTest, task.id, {
+				status: 'succeeded',
+				error_message: null,
+				attempt_count: 1,
+			});
 		});
 	});
 });
--- a/test/lib/pine-in-process.ts
+++ b/test/lib/pine-in-process.ts
@@ -1,6 +1,8 @@
 import { exit } from 'process';
 import type { PineTestOptions } from './pine-init';
 import { init } from './pine-init';
+import { tasks } from '../../src/server-glue/module';
+import { PINE_TEST_SIGNALS } from './common';
 
 export async function forkInit() {
 	try {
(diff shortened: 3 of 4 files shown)


joshbwlng:
I got things working and the test passing with these changes:
```
diff --git a/test/08-tasks.test.ts b/test/08-tasks.test.ts
index ef17b417..3c740765 100644
--- a/test/08-tasks.test.ts
+++ b/test/08-tasks.test.ts
@@ -421,7 +421,7 @@ describe('08 task tests', function () {
 			});
 
 			// SIGUSR1 is used to stop the task worker
-			pineServer.send('SIGUSR1');
+			pineServer.kill('SIGUSR1');
 
 			// Wait a few seconds to ensure the task is not executed
 			await setTimeout(3000);
diff --git a/test/lib/pine-in-process.ts b/test/lib/pine-in-process.ts
index 1d54fda1..b8268945 100644
--- a/test/lib/pine-in-process.ts
+++ b/test/lib/pine-in-process.ts
@@ -1,6 +1,7 @@
 import { exit } from 'process';
 import type { PineTestOptions } from './pine-init';
 import { init } from './pine-init';
+import { tasks } from '../../src/server-glue/module';
 
 export async function forkInit() {
 	try {
@@ -27,6 +28,9 @@ export async function forkInit() {
 				default: { initTaskHandlers },
 			} = await import(processArgs.taskHandlersPath);
 			initTaskHandlers();
+			process.on('SIGUSR1', () => {
+				tasks.worker?.stop();
+			});
 		}
 
 		if (processArgs.routesPath) {
diff --git a/test/lib/test-init.ts b/test/lib/test-init.ts
index a4798a36..4f285004 100644
--- a/test/lib/test-init.ts
+++ b/test/lib/test-init.ts
@@ -3,7 +3,6 @@ import { fork } from 'child_process';
 import { boolVar } from '@balena/env-parsing';
 import type { PineTestOptions } from './pine-init';
 import type { OptionalField } from '../../src/sbvr-api/common-types';
-import { tasks } from '../../src/server-glue/module';
 export const listenPortDefault = 1337;
 export const testLocalServer = `http://localhost:${listenPortDefault}`;
 
@@ -34,11 +33,6 @@ export async function testInit(
 			},
 		);
 
-		testServer.on('SIGUSR1', () => {
-			console.info(`TestServer received SIGUSR1`);
-			tasks.worker?.stop();
-		});
-
 		await new Promise((resolve, reject) => {
 			testServer.on('message', (msg: { init: string }) => {
 				console.info(`init pine in separate process`);
```

src/tasks/worker.ts
```diff
@@ -56,11 +56,13 @@ export class Worker {
 	public handlers: Record<string, TaskHandler> = {};
 	private readonly concurrency: number;
 	private readonly interval: number;
+	private isStopped: boolean;
```
joshbwlng, line 59:
```suggestion
	private isStopped = true;
```

src/tasks/worker.ts
```diff
@@ -56,11 +56,13 @@ export class Worker {
...
 	private executing = 0;
 
 	constructor(private readonly client: PinejsClient) {
 		this.concurrency = tasksEnv.queueConcurrency;
 		this.interval = tasksEnv.queueIntervalMS;
+		this.isStopped = true;
```
joshbwlng, line 65:
```suggestion
```

src/tasks/worker.ts
joshbwlng: Personally I'd probably name this something like `running` and default to `false` instead of the opposite.

test/lib/pine-in-process.ts
```diff
@@ -39,6 +40,10 @@ export async function forkInit() {
 		if (process.send) {
 			process.send({ init: 'success' });
 		}
+
+		process.on('SIGUSR1', () => {
```
joshbwlng, line 44: I grepped around in balena-api/open-balena-api and didn't find any other references to the `SIGUSR1` signal. I'm sure you have a good reason for choosing this signal, so could let me know and also add a comment about the reasoning?

test/08-tasks.test.ts
```diff
@@ -408,5 +408,44 @@ describe('08 task tests', function () {
 				})
 				.expect(500);
 		});
+
+		it('should not pick up new task after worker is stopped', async () => {
```
joshbwlng, line 412: Can we also add a test that ensures that a stopped worker can then be started again with `worker.start()` and have it pick back up where it stopped? Just to make sure things make sense from the consumer side and work as expected.

src/tasks/worker.ts
otaviojacobi: Much better. thanks!

test/08-tasks.test.ts
otaviojacobi: So I wanted to add a separate test for it, however, the more interesting test IMHO is a test that picks up a task that was created before the worker is stopped and then picked up. To break this down into two tests I would have to forward the task id and all other parameters from one test to another, which I didn't like. So I ended up with one test that stops the worker, test that it is not picked up, starts it and ensure the tasks runs.

test/lib/pine-in-process.ts
otaviojacobi: So the long story is that I wanted to use SIGTERM to be as similar as possible, however, handling sigterm makes the tests break as if I don't do `process.exit()` on it, the tests just hang after running.

Then I had the idea of using a custom signal (such as sigusr1 or sigusr2). This works, however, it sucks as it is a bit too magical and messes up with nodejs default usage of sigusr signals for debugging see https://nodejs.org/api/process.html#signal-events

My current approach now detatches from using any default unix signal to a list of custom pine_test signals handled using the standard nodejs `process.on('message')`. WDYT?

src/tasks/worker.ts
```diff
@@ -244,11 +245,17 @@ export class Worker {
...
 				}
-				this.poll();
+				if (this.running) {
+					this.poll();
+				}
 			}
 		})();
 	}
 
+	public stop(): void {
```
otaviojacobi, line 255: I also would like your opinion @joshbwlng on this, do you think we should maybe make this async already just to future proof we needing to do some async work to stop tasks?

joshbwlng: I like the idea. I can see the argument against it, but I find it quite likely that we would want to change this to async in the future, so 👍 from me.

test/08-tasks.test.ts
```diff
@@ -408,5 +409,66 @@ describe('08 task tests', function () {
...
+					name,
+					type: randomUUID(),
+				},
+				is_scheduled_to_execute_on__time: new Date(Date.now() + 2000),
+			});
+
+			pineServer.send(PINE_TEST_SIGNALS.STOP_TASK_WORKER);
+
+			// Wait a few seconds to ensure the task is not executed
+			await setTimeout(3000);
```
joshbwlng, line 427: You could also stop the worker first and then create the task. This makes the `is_scheduled_to_execute_on__time` unnecessary and safe to reduce the wait from 3s to say 2s. Removing the future execution argument also makes the test clearer as it removes any ambiguity as to why the task may have not been executed (it's definitely because the worker is stopped). What do you think?

test/lib/pine-in-process.ts
joshbwlng: Sounds good to me. At the end of the day, we just want need to prove that the new `stop()` does what it should. Maybe add a comment at the message handler in `pine-in-process.ts` to explain that this is an example of how a consumer could be configured to call `worker.stop()`.

test/08-tasks.test.ts
otaviojacobi: I think it is a great idea, if we stop the worker first and then create the task, I believe we don't need any `await setTimeout` in fact. There is maybe a very very tiny chance of the task being picked up while the event loop process the stop promise, but I think it is so small that we wouldn't see any intermitency ever. I ran the test locally 1k times and it succeded in all of them without any await.

joshbwlng (approved):

src/tasks/worker.ts
```diff
@@ -281,6 +288,7 @@ export class Worker {
 			);
 		}
 
+		this.running = true;
```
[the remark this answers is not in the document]
Page-: https://github.com/balena-io/pinejs/pull/830

PR #831, opened 2024-10-23
Linting fixes

Page-:
Change-type: patch

### DIFF
--- a/test/08-tasks.test.ts
+++ b/test/08-tasks.test.ts
@@ -96,7 +96,7 @@ describe('08 task tests', function () {
 			.expect(201);
 	});
 
-	after(async () => {
+	after(() => {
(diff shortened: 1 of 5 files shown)


test/08-tasks.test.ts
```diff
@@ -349,7 +349,7 @@ describe('08 task tests', function () {
 			const cron = '0 0 10 1 1 *';
 
 			// Test both succeeded and failure cases
-			['succeeded', 'failed'].forEach(async (status) => {
+			for (const status of ['succeeded', 'failed']) {
```
Page-, line 352: This is the linting fix, the rest is just making the test succeed now that it is actually run properly

otaviojacobi (approved):

PR #834, opened 2024-10-23
Infer async task fn callback arguments typing from schema

otaviojacobi:
Change-type: patch


src/tasks/index.ts
```diff
@@ -118,10 +118,12 @@ export async function setup(): Promise<void> {
...
 // Register a task handler
-export function addTaskHandler(
+export function addTaskHandler<T extends Schema>(
 	name: string,
-	fn: TaskHandler['fn'],
-	schema?: Schema,
+	fn: T extends undefined
+		? TaskHandler<undefined>['fn']
+		: TaskHandler<FromSchema<NonNullable<T>>>['fn'],
+	schema?: T,
```
thgreasi, line 126:
```suggestion
export function addTaskHandler<T extends Schema>(
	name: string,
	fn: TaskHandler<NonNullable<Task['Read']['is_executed_with__parameter_set']>>['fn'], // unknown would have been simpler perhaps :)
	schema?: undefined,
): void;
export function addTaskHandler<T extends Schema>(
	name: string,
	fn: TaskHandler<FromSchema<NonNullable<T>>>['fn'],
	schema: T,
): void;
export function addTaskHandler<T extends Schema>(
	name: string,
	fn: TaskHandler['fn'],
	schema?: T,
): void {
```

test/fixtures/08-tasks/task-handlers.ts
```diff
@@ -51,7 +51,7 @@ export const initTaskHandlers = () => {
 			throw new Error('This task is supposed to fail');
 		} catch (err: any) {
 			return {
-				status: 'failed',
+				status: 'failed' as const,
```
thgreasi, line 54: With the proposed change this is no longer needed

src/tasks/index.ts
otaviojacobi: I love it. thanks

otaviojacobi: @thgreasi mayb better than using `Task['Read']['is_executed_with__parameter_set']` we could use just `{}` as this is whhat it will default to when no schema is present. The `NonNullable<Task['Read']['is_executed_with__parameter_set']> could also be an array. See: https://github.com/balena-io/pinejs/blob/c94bbc91395a1afbabf44f09053c96d813ab9d82/src/tasks/worker.ts#L117

joshbwlng:
Tested locally after a rebase and it looks like a good improvement to me. Try out the changes with a real use case in `open-balena-api` before merging just in case.

src/tasks/index.ts
thgreasi: Afaict, tasks atm could work with arrays as a parameter set though, since the DB allows it on that field, we don't have any runtime check for arrays, and the ajv validation would work as far as you define your schema as an array. So I would lean towards sticking to what the DB says atm.

package.json
```diff
@@ -60,6 +60,7 @@
     "deep-freeze": "^0.0.1",
     "eventemitter3": "^5.0.1",
     "express-session": "^1.18.1",
+    "json-schema-to-ts": "3.1.1",
```
otaviojacobi, line 63: This needs to become an actual dependency (rather than dev dependency) as `addTaskHandler` is exposed externally and depends on it

otaviojacobi:
@joshbwlng tested with the service install async task in oB and all looks good :)

joshbwlng (approved):

PR #840, opened 2024-11-11
Run tests on UTC TZ

otaviojacobi:
Change-type: patch

### DIFF
--- a/docker-compose.npm-test.yml
+++ b/docker-compose.npm-test.yml
@@ -6,6 +6,7 @@ services:
       POSTGRES_USER: docker
       POSTGRES_PASSWORD: docker
       POSTGRES_DB: postgres
+      TZ: UTC
     ports:
       - "5431:5432"
   minio-server:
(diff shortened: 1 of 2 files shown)


joshbwlng:
I'd recommend also setting `TZ: UTC` on the postgres service in `docker-compose.npm-test.yml` to make sure the test runner and db are in sync. Currently the postgres container defaults to UTC, but I'd like to make it explicit.

As was discussed internally, we should switch to using `timestamp with time zone` data for a proper fix, but I think we should still merge this PR.

otaviojacobi:
@joshbwlng From https://www.postgresql.org/docs/8.3/libpq-envars.html
I believe the correct env var we should set for postgesql container is in fact `PGTZ` 

> PGTZ sets the default time zone. (Equivalent to SET timezone TO ....)

joshbwlng:
@otaviojacobi From what I understand, `PGTZ` is for client sessions while `TZ` is affects the server-side system time zone. I tested setting `TZ=Asia/Tokyo` and `PGTZ=UTC` at the same time and the timestamps were wonky like before.

otaviojacobi:
@joshbwlng  Oh I see, thanks! I tested locally with both set to UTC which wasn't very smart...
Should be fixed now

joshbwlng (approved):

PR #851, opened 2025-01-03
Tests: run against compiled source rather than transpiling on the fly

Page-:
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,10 +19,9 @@
     "webpack-server": "grunt server",
     "webpack-build": "npm run webpack-browser && npm run webpack-module && npm run webpack-server",
     "lint": "balena-lint -t tsconfig.dev.json -e js -e ts src build typings Gruntfile.ts && npx tsc --project tsconfig.dev.json --noEmit",
-    "test": "npm run lint && npm run build && npm run webpack-build && npm run test:compose && npm run test:generated-types",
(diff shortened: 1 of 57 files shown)


package.json
```diff
@@ -69,6 +68,7 @@
   },
   "devDependencies": {
     "@balena/lint": "^8.2.8",
+    "@balena/pinejs": "file:./",
```
joshbwlng, line 71: Nice, I've used `file:<path>` locally for testing but not for anything like this.

joshbwlng:
@Page- I noticed this diff after running `npm run test:generated-types`:
```diff
diff --git a/src/tasks/tasks.ts b/src/tasks/tasks.ts
index d499ff4b..9930e261 100644
--- a/src/tasks/tasks.ts
+++ b/src/tasks/tasks.ts
@@ -41,4 +41,6 @@ export interface Task {
 
 export default interface $Model {
 	task: Task;
-}
+
+
+}
\ No newline at end of file
```

joshbwlng:
> @Page- I noticed this diff after running `npm run test:generated-types`:
> 
> ```diff
> diff --git a/src/tasks/tasks.ts b/src/tasks/tasks.ts
> index d499ff4b..9930e261 100644
> --- a/src/tasks/tasks.ts
> +++ b/src/tasks/tasks.ts
> @@ -41,4 +41,6 @@ export interface Task {
>  
>  export default interface $Model {
>  	task: Task;
> -}
> +
> +
> +}
> \ No newline at end of file
> ```

Ah, but I see this happens on master as well. Unrelated to this PR, but still something that should probably be fixed.

joshbwlng (approved):

Page-:
@joshbwlng I would guess that's because of the difference between directly generated vs after linting

PR #856, opened 2025-01-20
Compile in as `(= ANY($singleListBinding))`

otaviojacobi:
Update @balena/abstract-sql-compiler from 10.1.0 to 10.2.0
Update @balena/odata-parser from 3.1.2 to 4.1.0
Update @balena/odata-to-abstract sql from 7.0.1 to 7.1.0

Change-type: patch
See: https://balena.fibery.io/search/rLsEn#Work/Project/Pinejs-$in-as-a-single-bind-895

### DIFF
--- a/src/sbvr-api/abstract-sql.ts
+++ b/src/sbvr-api/abstract-sql.ts
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 
+import type { Engines } from '@balena/abstract-sql-compiler';
 import AbstractSQLCompiler from '@balena/abstract-sql-compiler';
 import type { BindKey } from '@balena/odata-parser';
 import {
@@ -78,7 +79,7 @@ export const getAndCheckBindValues = async (
 	return await Promise.all(
 		bindings.map(async (binding) => {
 			let fieldName = '';
-			let field: { dataType: string };
+			let dataType: string;
 			let value: any;
 			if (binding[0] === 'Bind') {
 				const bindValue = binding[1];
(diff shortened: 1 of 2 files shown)


src/sbvr-api/abstract-sql.ts
```diff
@@ -140,14 +141,32 @@ export const getAndCheckBindValues = async (
...
 			}
 		}),
 	);
 };
 
+const validateBindingType = async (
+	engine: Engines,
+	$value: any,
+	field: { dataType: string },
+): Promise<any> => {
```
Page-, line 156: Is `any` the correct type here? I assume it probably is but want to check

src/sbvr-api/abstract-sql.ts
```diff
@@ -140,14 +141,32 @@ export const getAndCheckBindValues = async (
...
 				throw new BadRequestError(`"${fieldName}" ${err.message}`);
 			}
 		}),
 	);
 };
 
+const validateBindingType = async (
+	engine: Engines,
+	$value: any,
+	field: { dataType: string },
```
Page-, line 155: I think it'd make sense to do
```suggestion
	dataType: string,
```
and then pass in `field.datatype` rather than having to create an object via `{ dataType }` just to recurse

otaviojacobi: @Page- Makes sense. I ended up refactoring the whole piece of code above (the content inside `bindings.map`) as I don't think we needed to have the `field` top level construct anymore.

src/sbvr-api/abstract-sql.ts
otaviojacobi: Yes, it is the incoming type from abstract-sql-compiler

```
dataTypeValidate: (value: any, field: Pick<AbstractSqlField, "dataType" | "required">) => Promise<any>;
```

I initially tried using `ReturnType<Awaited<typeof AbstractSQLCompiler[engine]dataTypeValidate>>`  but the dynamic `engine` break this typing

Page- (approved):

PR #866, opened 2025-03-04
Use 204 status for POSTs that return no result

thgreasi:
Change-type: patch
See: https://github.com/balena-io/balena-api/pull/5663#discussion_r1975578000


thgreasi:
Abandoned since we really don't expect pine to be normally returning an empty response.
For our current use case, we will use the PRERESPOND hook in the API.

PR #873, opened 2025-03-24
Update cron-parser to v5

joshbwlng:
Change-type: patch

### DIFF
--- a/src/tasks/index.ts
+++ b/src/tasks/index.ts
@@ -1,5 +1,5 @@
 import type { Schema } from 'ajv';
-import cronParser from 'cron-parser';
+import { CronExpressionParser } from 'cron-parser';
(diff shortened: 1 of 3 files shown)


src/tasks/index.ts
```diff
@@ -77,11 +77,12 @@ export function setup(): void {
...
-						.parseExpression(request.values.is_scheduled_with__cron_expression)
-						.next()
-						.toDate()
-						.toISOString();
+					request.values.is_scheduled_to_execute_on__time =
+						CronExpressionParser.parse(
+							request.values.is_scheduled_with__cron_expression,
+						)
+							.next()
+							.toISOString();
```
joshbwlng, line 85: The extra call to `toDate()` was unnecessary so dropping that as well

thgreasi (approved):

PR #888, opened 2025-04-16
Postgres: add the option to prepare db queries after N executions

Page-:
Change-type: minor

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "@balena/odata-to-abstract-sql": "^7.2.0",
     "@balena/sbvr-parser": "^1.4.9",
     "@balena/sbvr-types": "^9.2.2",
+    "@sindresorhus/fnv1a": "^3.1.0",
     "@types/body-parser": "^1.19.5",
     "@types/compression": "^1.7.5",
     "@types/cookie-parser": "^1.4.8",
--- a/src/config-loader/env.ts
+++ b/src/config-loader/env.ts
@@ -112,6 +112,11 @@ export const db = {
 	 * Defaults to true when in DEBUG mode, false otherwise
 	 */
 	checkReadOnlyQueries: DEBUG,
+	/**
+	 * Prepare the statement after N executions, false to never prepare, true to always prepare, otherwise a positive number
+	 * Currently only supported for postgres
+	 */
+	prepareAfterN: false as boolean | number,
 };
 
 export const PINEJS_ADVISORY_LOCK = {
--- a/src/database-layer/db.ts
+++ b/src/database-layer/db.ts
@@ -10,6 +10,7 @@ import _ from 'lodash';
 import { TypedError } from 'typed-error';
 import * as env from '../config-loader/env.js';
 import { fromCallback, timeout } from '../sbvr-api/control-flow.js';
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -64,6 +65,7 @@
     "memoizee": "^0.4.17",
     "pinejs-client-core": "^8.1.4",
     "randomstring": "^1.3.1",
+    "string-hash": "^1.1.3",
```
otaviojacobi, line 68: @Page- are we 100% sure we want to go with this module? Altough it is extensively used, the repo has been deprecated/made read-only:
https://github.com/darkskyapp/string-hash/blob/master/index.js

Do you consider using native crypto.createHash and create a simple md5 or sha hash too problematic/slow here?

src/config-loader/env.ts
```diff
@@ -112,6 +112,11 @@ export const db = {
 	 * Defaults to true when in DEBUG mode, false otherwise
 	 */
 	checkReadOnlyQueries: DEBUG,
+	/**
+	 * Prepare the statement after N executions, false to never prepare, true to always prepare, otherwise a positive number
+	 * Currently only supported for postgres
+	 */
+	prepareAfterN: false as boolean | number,
```
otaviojacobi, line 119: Why is this typed as `false as boolean | number`  when the comment leads me to believe that `true` is a valid type?

src/database-layer/db.ts
```diff
@@ -166,6 +167,26 @@ export function registerTransactionLockNamespace(
...
+	if (env.db.prepareAfterN === true) {
+		return sqlHash;
+	}
+
+	maybePrepareCache[sqlHash] ??= 0;
+	if (maybePrepareCache[sqlHash] > env.db.prepareAfterN) {
+		return sqlHash;
+	}
+	maybePrepareCache[sqlHash]++;
+	return;
```
otaviojacobi, line 187: Is the empty return necessary here when the function just ends?

src/config-loader/env.ts
Page-: `true` is valid, that's why we have the cast to make the typing include it

src/database-layer/db.ts
Page-: It's not necessary, it's just a remnant from a previous version - removed

package.json
otaviojacobi: Also library mantainer said library is no longer mantained here: https://github.com/darkskyapp/string-hash/issues/12#issuecomment-619037758 And deleted gh account. Altough it is a simple module in this same issue they are discussing possible optimisations etc which I would hope are baked in the native module

Page-: It's the module we use in odata-to-abstract-sql which is why I reused it, I've checked for alternatives and gone with fnv1a by sindresorhus as it seems to be a good mix of speed/support - we don't need a cryptographically secure hash and speed is more important

src/config-loader/env.ts
Page-: That is the type @otaviojacobi, this is a literal object with defaults, with an `as boolean | number` to avoid the type being inferred as `false` from the default and instead use the expected `boolean | number` for the type of the object whilst still initializing with the default value

PR #891, opened 2025-04-25
Share `namespaceRelationships` logic between permissions and translat…

Page-:
…ions

Change-type: patch

### DIFF
--- a/src/sbvr-api/abstract-sql.ts
+++ b/src/sbvr-api/abstract-sql.ts
@@ -1,6 +1,11 @@
 import _ from 'lodash';
 
-import type { Engines } from '@balena/abstract-sql-compiler';
+import type {
+	Engines,
+	Relationship,
+	RelationshipInternalNode,
+	RelationshipLeafNode,
+} from '@balena/abstract-sql-compiler';
 import AbstractSQLCompiler from '@balena/abstract-sql-compiler';
 import type { BindKey } from '@balena/odata-parser';
 import {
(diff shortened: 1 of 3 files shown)


src/sbvr-api/abstract-sql.ts
```diff
@@ -232,3 +237,37 @@ export const isRuleAffected = (
...
+	for (const [key, relationship] of Object.entries(
+		relationships as RelationshipInternalNode,
+	)) {
+		if (key === '$') {
+			continue;
+		}
+
+		let mapping = (relationship as RelationshipLeafNode).$;
+		if (mapping != null && mapping.length === 2) {
+			if (opts?.keyFilter?.(key) !== false) {
```
thgreasi, line 258: Took me a bit to realize that this means v, but it might be b/c it's Monday :)
```suggestion
			if (opts.keyFilter == null && opts.keyFilter(key)) {
```

src/sbvr-api/abstract-sql.ts
```diff
@@ -232,3 +237,37 @@ export const isRuleAffected = (
...
+				mapping = _.cloneDeep(mapping);
+				// we do check the length above, but typescript thinks the second
+				// element could be undefined
+				mapping[1]![0] = `${mapping[1]![0]}$${alias}`;
+				(relationships as RelationshipInternalNode)[`${key}$${alias}`] = {
+					$: mapping,
+				};
+				if (opts?.deleteOriginal) {
```
thgreasi, line 266:
```suggestion
				if (opts.deleteOriginal) {
```

thgreasi (approved):

src/sbvr-api/abstract-sql.ts
Page-: `opts` can still be undefined here

src/sbvr-api/abstract-sql.ts
Page-: The alternative would actually be `if (opts.keyFilter == null || opts.keyFilter(key) !== false) {` btw

PR #893, opened 2025-04-30
Add support for defining odata actions

otaviojacobi:
See: https://balena.fibery.io/search/9OI72#Work/Improvement/Add-odata-actions-foundation-for-pinejs-2818
Change-type: minor


src/sbvr-api/sbvr-utils.ts
```diff
@@ -1386,8 +1387,16 @@ const runODataRequest = (req: Express.Request, vocabulary: string) => {
 
 				let request: uriParser.ODataRequest | uriParser.ODataRequest[];
 				if (Array.isArray(parsedRequest)) {
+					if (parsedRequest.some(isActionRequest)) {
```
otaviojacobi, line 1390: I suppose it is not too hard to add support for it inside $batch by passing a tx to `runAction` which is then shared across all requests in the $batch but I would rather start without it considering we don't even properly support $batch

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,141 @@
...
+
+	const applicationApi = api[request.vocabulary];
+	if (!applicationApi) {
+		throw new BadRequestError(`Could not find model ${request.vocabulary}`);
+	}
+
+	const res = await api[request.vocabulary].request({
+		method: 'POST',
+		url: canAccessUrl,
+		body: { action: actionName },
```
otaviojacobi, line 123: This is one of the most interesting parts IMHO. We can actually ensure you only run the action if you have permission to it (see here for example of how it looks like). When we have filtered canAccess this can be further improve to run the filtered canAccess so that you can have full permissions over the action execution

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,141 @@
...
+			request: request,
+			tx,
+			api: applicationApi,
+			req: req,
+			id,
+		});
+	});
+};
+
+export const addAction = <Vocab extends string>(
```
otaviojacobi, line 94: Different than addHook we do not pass the method (as per odata definition actions are always POST) - however, wen need to pass a name

test/fixtures/09-actions/config.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export default {
+	models: [
+		{
+			modelName,
+			modelFile,
+			apiRoot,
+		},
+		{
+			apiRoot: 'v1actionsUniversity',
```
otaviojacobi, line 16: I am using the full name here instead of just `v1` just because I wanted the tests to  [be able to infer the PinejsClient typings](https://github.com/balena-io/pinejs/pull/893/files#diff-27e738a0525e0effbc7a062e4fa79c0874932afe5f3be566ee10d6308000eaa7R5-R9) without overiding the type for any model called v1 (which ended up breaking the typings of the other tests so this is just a namespacing)

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,141 @@
...
+	const actionName = request.odataQuery.property.resource;
+	const action =
+		actions?.[request.vocabulary]?.[request.resourceName]?.[actionName];
+	if (!action) {
+		throw new BadRequestError();
+	}
+
+	return sbvrUtils.db.transaction(async (tx) => {
+		// in practice, the parser does not currently allow actions without a key
+		// e.g. /vocab/resource(...)/action compiles but
```
otaviojacobi, line 72: In practice, this allows only for Bound Actions:
https://learn.microsoft.com/en-us/odata/webapi-8/fundamentals/actions-functions#bound-action-1

Unbound actions can be done but require extra support on the parser side

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,146 @@
...
+	if (!canAccessUrl.endsWith('/canAccess')) {
+		throw new UnauthorizedError();
+	}
+
+	const applicationApi = api[request.vocabulary];
+	if (!applicationApi) {
+		throw new BadRequestError(`Could not find model ${request.vocabulary}`);
+	}
+
+	const res = await api[request.vocabulary].request({
```
Page-, line 125:
```suggestion
	const res = await applicationApi.request({
```

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export const canRunAction = async (
+	request: ParsedODataRequest,
+	req: ActionReq,
+	actionName: string,
+	tx: Tx,
+) => {
+	const canAccessUrl = request.url
+		.slice(1)
+		.split('?')[0]
```
Page-, line 113:
```suggestion
		.split('?', 1)[0]
```

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,146 @@
...
+
+		const applicationApi = api[request.vocabulary].clone({
+			passthrough: { tx, req },
+		});
+
+		return await action({
+			request: request,
+			tx,
+			api: applicationApi,
+			req: req,
```
Page-, line 88:
```suggestion
			req,
```

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,146 @@
...
+		const id = request.odataQuery.key
+			? await canRunAction(request, req, actionName, tx)
+			: undefined;
+
+		const applicationApi = api[request.vocabulary].clone({
+			passthrough: { tx, req },
+		});
+
+		return await action({
+			request: request,
```
Page-, line 85:
```suggestion
			request,
```

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,146 @@
...
+	req: ActionReq,
+): Promise<Response> => {
+	if (api[request.vocabulary] == null) {
+		throw new BadRequestError();
+	}
+
+	const actionName = request.odataQuery.property.resource;
+	const action =
+		actions?.[request.vocabulary]?.[request.resourceName]?.[actionName];
+	if (!action) {
```
Page-, line 66: Again I assume this is probably a `!= null` in disguise?

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,146 @@
...
+	args: ODataActionArgs<Vocab>,
+) => Promise<Response>;
+
+const actions = {} as {
+	[vocab: string]: {
+		[resourceName: string]: {
+			[actionName: string]: ODataAction;
+		};
+	};
+};
```
Page-, line 39:
```suggestion
const actions: {
	[vocab: string]: {
		[resourceName: string]: {
			[actionName: string]: ODataAction;
		};
	};
} = {};
```

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,146 @@
...
+	return canAccessResourceId(res);
+};
+
+const canAccessResourceId = (canAccessResponse: AnyObject): unknown => {
+	const item = canAccessResponse?.d?.[0];
+	if (!item || typeof item !== 'object') {
+		throw new UnauthorizedError();
+	}
+	const keys = Object.keys(item);
+	if (keys.length !== 1 || !item[keys[0]]) {
```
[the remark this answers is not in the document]
otaviojacobi: I think you meant `== null` but yes, you are correct.

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,146 @@
...
+	[vocab: string]: {
+		[resourceName: string]: {
+			[actionName: string]: ODataAction;
+		};
+	};
+} = {};
+
+export const isActionRequest = (
+	request: ParsedODataRequest,
+): request is ODataActionRequest => {
```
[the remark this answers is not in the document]
otaviojacobi: I think both approaches are valid - I slightly prefer the current as it makes explicit the action & request are tied to gether (e.g. I am not running an action for this request but rather I am running a request that is an action) and also makes it less work for the caller code

test/fixtures/09-actions/translations/v1/v1actionsUniversity.sbvr
```diff
@@ -0,0 +1,25 @@
...
+
+Fact Type: student has name
+	Necessity: each student has exactly one name
+	Necessity: each name is of exactly one student
+
+Fact Type: student has current semester
+	Necessity: each student has exactly one current semester
+	Necessity: each student has a current semester that is greater than 0 and is less than 12
+
+Fact Type: student is repeating
```
thgreasi, line 22: I'm not sure what that means, but not a big deal to me

test/fixtures/09-actions/actions.ts
```diff
@@ -0,0 +1,67 @@
...
+				options: {
+					$select: ['semester'],
+				},
+			});
+
+			await api.patch({
+				resource: 'student',
+				id,
+				body: {
+					semester: student!.semester + 1,
```
thgreasi, line 52: Maybe we can bring assertExists here as well.

test/fixtures/09-actions/translations/v1/v1actionsUniversity.sbvr
otaviojacobi: Yeah just it basically means if this user is taking the same semester again

test/fixtures/09-actions/actions.ts
otaviojacobi: done

src/sbvr-api/actions.ts
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	const actionName = request.odataQuery.property.resource;
+	const action =
+		actions[request.vocabulary]?.[request.resourceName]?.[actionName];
+	if (action == null) {
+		throw new BadRequestError();
+	}
+
+	return await sbvrUtils.db.transaction(async (tx) => {
```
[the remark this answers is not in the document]
otaviojacobi: I went for the simpler version where it runs on its own transaction (which can be extended). 

The runTransaction in sbvrUtils does a bunch of other stuff to the request which we don't want to do for actions - such as rolling back the requestHooks (which will be rolled back if an action triggers an inner request) but we for sure don't want to rollback a resource hooks for an action on that resource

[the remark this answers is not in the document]
thgreasi: This came up while I was checking how the current `/canAccess()` action is atm re-using any existing tx, so if we were to re-implement it using this new standardized action mechanism, then we would need to re-use the existing one (otherwise any uncommitted changes wouldn't be visible to the new canAccess).
So I think that might be enough of a reason to try re-using any existing tx.
And as a result of that, we would probably also need the `transactions.push(tx);` & `tx.on('rollback')` callls.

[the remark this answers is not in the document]
otaviojacobi: FWIW the canAccess should be a function and follow its implementation pattern there - not an action. 

With that said, I agree we can try to use an existing tx in case it exists, however, I don't think we should have the tx events added to this transaction directly as if a tx is created by any handler that does any other pine request (forwarding this tx) then pine will add the proper handles during that request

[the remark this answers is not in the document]
thgreasi: I would be fine leaving this for later, but that would require us blocking using the new standardized actions when a tx already exists (in other words only allow it for external requests).
That way changing the actions later to use the same tx would not be a functional change.

But I think that we now agree that the correct target state would be to re-use any existing tx.

test/fixtures/09-actions/config.ts
thgreasi: I would say this justifies being a comment in the src b/c otherwise nobody is going to remember the reason by just looking the code.

src/sbvr-api/actions.ts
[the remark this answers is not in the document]
thgreasi:
```
					const tx = req.tx;
					const runInTx = tx != null ? async <T>(fn: (tx: Db.Tx) => Resolvable<T>): Promise<T> => {
						return await fn(tx);
					} : sbvrUtils.db.transaction
```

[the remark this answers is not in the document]
otaviojacobi: I wrote slightly different but I agree - let's reuse the tx in case it exists.

test/fixtures/09-actions/actions.ts
```diff
@@ -0,0 +1,70 @@
...
+			);
+		}
+
+		const { grades } = request.values;
+
+		if (!Array.isArray(grades)) {
+			throw new errors.BadRequestError('Invalid payload');
+		}
+
+		const average = grades.reduce((a, b) => a + b) / grades.length;
```
thgreasi, line 29:
```suggestion
		const average = grades.reduce((sum, grade) => sum + grade) / grades.length;
```

PR #896, opened 2025-05-04
Delete files with pine tasks when it is available

otaviojacobi:
See: https://balena.fibery.io/search/bGknx#Work/Improvement/Make-webresources-file-deletion-more-reliable-2819

### DIFF
--- a/src/tasks/index.ts
+++ b/src/tasks/index.ts
@@ -11,6 +11,7 @@ import type TasksModel from './tasks.js';
 import type { Task } from './tasks.js';
 import type { FromSchema } from 'json-schema-to-ts';
 import { importSBVR } from '../server-glue/sbvr-loader.js';
+import { addPineTaskHandlers } from './pine-tasks.js';
 
 export type * from './tasks.js';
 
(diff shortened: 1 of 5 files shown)


src/tasks/index.ts
```diff
@@ -44,6 +45,11 @@ export const config: ConfigLoader.Config = {
 };
 
 export let worker: Worker | null = null;
+
+export function isPineTasksAvailable() {
```
joshbwlng, line 49:
```suggestion
export function canExecuteTasks() {
```

Maybe personal preference, but I find this to be clearer naming

src/webresource-handler/index.ts
```diff
@@ -320,7 +321,24 @@ const deleteFiles = async (
...
+					passthrough: {
+						req: permissions.root,
+					},
+					body: {
+						key: crypto.randomUUID(),
+						is_executed_by__handler: 'delete_webresource_file',
+						is_executed_with__parameter_set: {
+							fileKey: fileKey,
+						},
+						attempt_limit: 2 ** 31 - 1,
```
joshbwlng, line 337: Can you leave a comment as to why this is the attempt_limit

joshbwlng:
Technically this might be a minor as you're adding a new task handler that consumers will also be able to execute

joshbwlng (approved):

PR #898, opened 2025-05-12
Block root and rootRead of changing during runtime

otaviojacobi:
Change-type: patch


src/sbvr-api/permissions.ts
```diff
@@ -70,20 +70,20 @@ export interface PermissionReq {
...
+export const root: Readonly<PermissionReq> = Object.freeze({
 	user: {
 		id: 0,
 		actor: 0,
 		permissions: ['resource.all'],
 	},
-};
-export const rootRead: PermissionReq = {
+});
+export const rootRead: Readonly<PermissionReq> = Object.freeze({
```
Page-, line 80: You probably want to use the deep-freeze module to make sure all parts are frozen rather than just the top-level

PR #936, opened 2025-08-06
Always add the constrained definition to constrained tables

Page-:
This just avoids any potential edge cases and simplifies things

Change-type: patch

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -2,6 +2,7 @@ import type AuthModel from './user.js';
 import type {
 	AbstractSqlModel,
 	AbstractSqlQuery,
+	AbstractSqlTable,
 	AliasNode,
 	AnyTypeNodes,
 	Definition,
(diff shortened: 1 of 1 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -1033,34 +1034,50 @@ const getBoundConstrainedMemoizer = memoizeWeak(
...
-						// Otherwise constrain the non-bypass table
-						// When the table doesn't have a definition, then it's the last translation layer,
-						// in which case we define it as the virtual `resource$vDB$permissions"read"` resource,
+						const bypassTable = (constrainedAbstractSqlModel.tables[
+							bypassResourceName
+						] = {
+							...constrainedTable,
+							resourceName: bypassResourceName,
+						});
+						if (constrainedTable.definition) {
```
otaviojacobi, line 1061:
```suggestion
						if (constrainedTable.definition != null) {
```

otaviojacobi (approved):

PR #939, opened 2025-08-12
Add the ability to have computed fields generate/call functions

Page-:
https://balena.fibery.io/Work/Project/1766

Update @balena/abstract-sql-compiler from 11.0.0 to 11.1.0

Change-type: minor


otaviojacobi:
@Page- you are also bumping other packages, including @balena/odata-to-abstract-sql, would you mind adding it on the commit message for the nested changelog?

Page-:
The `"@balena/odata-to-abstract-sql` update is purely a test change so if we had the ability it wouldn't even have been a patch but rather just an unreleased change, hence why I didn't think it was worth propagating its changelog

otaviojacobi (approved):

PR #995, opened 2026-04-06
V24

thgreasi:
Change-type: major
See: https://github.com/balena-io-modules/lf-to-abstract-sql/pull/98
See: https://balena.fibery.io/Work/Project/2031
See: https://balena.fibery.io/Work/Project/2429


src/sbvr-api/permissions.ts
```diff
@@ -1643,9 +1642,8 @@ export const authorizationMiddleware = customAuthorizationMiddleware();
 
 export const resolveApiKey = async (
 	req: HookReq | Express.Request,
+	tx: Tx | undefined,
```
Page-, line 1645: Is it possible the reasoning for it being the second argument was to make it required?

thgreasi: Afaik the point was to force users to only be able to not provide it if they explicitly state so.
I think we still need to be able to pass an null/undefined value for the cases where a TX doesn't yet exist, like how we use it in here in a middleware.

package.json
```diff
@@ -131,7 +131,7 @@
     "serve-static": "^1.16.2"
   },
   "engines": {
-    "node": "^20.14.0 || ^22.0.0 || ^24.5.0",
+    "node": "^22.2.0 || ^24.5.0",
```
Page-, line 134: Can you also update `@types/node` to match the 22.x min, and similarly tsconfig to es2024

thgreasi: Opps! Of course

thgreasi: and ES target I guess...

Page- (approved):

PR #1004, opened 2026-06-02
Add support for binding per request with custom req

otaviojacobi:
See: https://balena.fibery.io/Work/Project/Ingest-audit-logs-from-balena-proxy-2469
Change-type: minor

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -15,10 +15,11 @@ import type Express from 'express';
 import type {
 	ODataBinds,
 	ODataQuery,
+	ParameterAliasBind,
 	SupportedMethod,
 } from '@balena/odata-parser';
 import type { Tx } from '../database-layer/db.js';
-import type { ApiKey, User } from '../sbvr-api/sbvr-utils.js';
+import type { Actor, ApiKey, User } from '../sbvr-api/sbvr-utils.js';
 import type {
 	AnyObject,
 	Dictionary,
@@ -58,11 +59,9 @@ import { importSBVR } from '../server-glue/sbvr-loader.js';
 
 const userModel = await importSBVR('./user.sbvr', import.meta);
 
-const DEFAULT_ACTOR_BIND = '@__ACTOR_ID';
-const DEFAULT_ACTOR_BIND_REGEX = new RegExp(
-	_.escapeRegExp(DEFAULT_ACTOR_BIND),
-	'g',
-);
+const EXTRA_BIND_PREFIX = '@__';
+const DEFAULT_ACTOR_BIND = `${EXTRA_BIND_PREFIX}ACTOR_ID` as const;
+const EXTRA_BIND_REGEX = new RegExp(_.escapeRegExp(EXTRA_BIND_PREFIX), 'g');
 
 export { PermissionError, PermissionParsingError };
 export interface PermissionReq {
@@ -1732,14 +1731,20 @@ const getGuestPermissions = memoize(
 		}
 		const guestPermissions = _.uniq(await getUserPermissions(result.id));
 
-		if (guestPermissions.some((p) => DEFAULT_ACTOR_BIND_REGEX.test(p))) {
-			throw new Error('Guest permissions cannot reference actors');
+		if (guestPermissions.some((p) => EXTRA_BIND_REGEX.test(p))) {
+			throw new Error(
+				'Guest permissions cannot reference actor or extra bindings',
+			);
 		}
 		guestPermissionsInitialized = true;
 		return guestPermissions;
 	},
 	{ promise: true },
 );
+const startsWithBindPrefix = (
+	bindName: string,
+): bindName is `${typeof EXTRA_BIND_PREFIX}${string}` =>
+	bindName.startsWith(EXTRA_BIND_PREFIX);
 
 const getReqPermissions = async (
 	req: PermissionReq,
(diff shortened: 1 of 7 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -1759,15 +1759,44 @@ const getReqPermissions = async (
 	})();
 
 	let actorPermissions: string[] = [];
-	const addActorPermissions = (actorId: number, perms: string[]) => {
+	const addActorPermissions = (
+		actorId: number,
+		perms: string[],
+		extraBinds?: sbvrUtils.ExtraBinds,
+	) => {
 		odataBinds[DEFAULT_ACTOR_BIND] = ['Real', actorId];
```
Page-: I think it makes more sense to do
```suggestion
	const addActorPermissions = (
		creds: Creds,
	) => {
		odataBinds[DEFAULT_ACTOR_BIND] = ['Real', creds.actor];
```
etc and simplify the calls as `addActorPermissions(req.user)`, since in practice they should both follow the same shape and breaking out the args is unnecessary

src/sbvr-api/permissions.ts
```diff
@@ -1759,15 +1759,44 @@ const getReqPermissions = async (
...
+		if (extraBinds != null) {
+			// tihs `as` cast is needed as Object.entries simplifies the narrow typing of `@${string}`
+			// from sbvrUtils.ExtraBinds to string
+			for (const [bindName, bindValue] of Object.entries(extraBinds) as Array<
+				[`@${string}`, sbvrUtils.ExtraBind | undefined]
```
Page-: Since these aren't guaranteed to always be passed in may be necessary to enforce that they follow a particular schema, eg the `@__${varName}` that we use for the actor id and then block users using parameter aliases that match that prefix just to avoid potential clashes - alternatively maybe a slightly different version for internal pinejs added ones and external user code added ones, to avoid clashes if we decide to add more default ones than just the actor id

src/sbvr-api/sbvr-utils.ts
```diff
@@ -137,8 +143,17 @@ const models: {
 	};
 } = {};
 
+export type ExtraBind = NumberBind | BooleanBind | TextBind | DateBind;
```
Page-: It might be worth adding a type for this to odata-parser that can be used instead as I think it also makes sense there (albeit named as `ParameterAliasBind` or similar)

src/sbvr-api/permissions.ts
otaviojacobi: the `Creds` typing does not really exist in pinejs, however I think `User | ApiKey' is close enough

src/sbvr-api/permissions.ts
otaviojacobi: Agree. I also think that if/when we use this inside JWTs we will want short names to not increase the payload too much, but I think the `@__` prefix is small and reserved enough, if people using URL bindings are appending @__ to their aliases, there is a bigger issue for sure

src/sbvr-api/permissions.ts
```diff
@@ -1759,15 +1761,45 @@ const getReqPermissions = async (
...
+					);
+				}
+				if (bindName === DEFAULT_ACTOR_BIND) {
+					throw new BadRequestError(
+						`The '${DEFAULT_ACTOR_BIND}' bind is reserved and cannot be overridden via extraBinds`,
+					);
+				}
+				// The `as` cast is necessary as Object.entries infers the key as string
+				// and startsWith does not narrow it down.
+				const extraBindName = bindName as `@__${string}`;
```
Page-: This can also be done via
```ts
const startsWithBindPrefix = (s: string): s is `${EXTRA_BIND_PREFIX}${string}` => bindName.startsWith(EXTRA_BIND_PREFIX)
```
and then
```ts
				if (!startsWithBindPrefix(bindName)) {
					throw new BadRequestError(
						`extraBinds keys must use the '${EXTRA_BIND_PREFIX}' prefix (got '${bindName}')`,
					);
				}

File: src/sbvr-api/permissions.ts [hunk 5]
```diff
@@ -1759,15 +1761,45 @@ const getReqPermissions = async (
...
+				let tree: ODataParser.ODataOptions;
+				try {
+					({ tree } = ODataParser.parse(`${bindName}=${bindValue}`, {
+						startRule: 'ProcessRule',
+						rule: 'QueryOptions',
```
Page-: You can skip a whole bunch of unnecessary work by skipping direct to the `ParameterAliasOption` rule:
```suggestion
						rule: 'ParameterAliasOption',
```

src/sbvr-api/permissions.ts
```diff
@@ -58,7 +59,8 @@ import { importSBVR } from '../server-glue/sbvr-loader.js';
 
 const userModel = await importSBVR('./user.sbvr', import.meta);
 
-const DEFAULT_ACTOR_BIND = '@__ACTOR_ID';
+const EXTRA_BIND_PREFIX = '@__';
+const DEFAULT_ACTOR_BIND = `${EXTRA_BIND_PREFIX}ACTOR_ID` as const;
 const DEFAULT_ACTOR_BIND_REGEX = new RegExp(
```
Page-: The guest permissions check should possibly be updated to block any types of binds? I'm not sure on that but  I'd rather restrict it now and relax later if it becomes necessary/useful and we can decide at that time if that is safe to do

src/sbvr-api/permissions.ts
Page-: I know, I was thinking about making a shared type - but in reality it looks like the typing to use would be the `Actor` type and move the `actor: number;` from being individually on `ApiKey`/`User` onto that parent `Actor` type instead (which also.. why is the `Actor` type the only one that doesn't include the `actor` property?? 🤔)

src/sbvr-api/permissions.ts
otaviojacobi: Lovely

src/sbvr-api/permissions.ts
otaviojacobi: @Page- Well, altough I agree it is weird, this would be a breaking change which is not really related to this - I think we can do that in a follow up major (with a TODO) and keep the typing as is OR also change the input to be `Actor & { actor: number }` as I proposed on the latest version

src/sbvr-api/permissions.ts
```diff
@@ -1759,15 +1764,42 @@ const getReqPermissions = async (
...
 
 	let actorPermissions: string[] = [];
-	const addActorPermissions = (actorId: number, perms: string[]) => {
-		odataBinds[DEFAULT_ACTOR_BIND] = ['Real', actorId];
-		actorPermissions = perms;
+	const addActorPermissions = (creds: Actor & { actor: number }) => {
+		odataBinds[DEFAULT_ACTOR_BIND] = ['Real', creds.actor];
+		if (creds.extraBinds != null) {
+			for (const [bindName, bindValue] of Object.entries(creds.extraBinds)) {
+				if (bindValue == null) {
```
Page-: This shouldn't really be able to happen surely? I think I would expect this to throw to make it easier for people to know something has gone wrong/they're doing something weird

otaviojacobi: :+1: This if clause is not necessary

Page- (approved):

PR #1017, opened 2026-06-23
Use zod schemas from sbvr-types to perform early validation before passing to hooks

Page-:
This means hooks will receive validated data that they can trust to be of the expected type and improve the developer experience along with safety

https://balena.fibery.io/Work/Project/1002

Change-type: minor

### DIFF
--- a/src/pinejs-session-store/pinejs-session-store.ts
+++ b/src/pinejs-session-store/pinejs-session-store.ts
@@ -72,7 +72,8 @@ export class PinejsSessionStore extends Store {
 	public set = ((sid, data, callback) => {
 		const body = {
 			session_id: sid,
-			data,
+			// Convert from a `Session` instance into a plain object that is valid to pass
+			data: { ...data },
 			expiry_time: data?.cookie?.expires ?? null,
 		};
 		void asCallback(
--- a/src/sbvr-api/validation.ts
+++ b/src/sbvr-api/validation.ts
@@ -1,0 +1,45 @@
+import type {
+	AbstractSqlModel,
+	AbstractSqlTable,
+} from '@balena/abstract-sql-compiler';
+import { sqlNameToODataName } from '@balena/odata-to-abstract-sql';
+import z from 'zod';
+import $sbvrTypes from '@balena/sbvr-types';
+const { default: sbvrTypes } = $sbvrTypes;
+
+// Augment express.js with pinejs-specific attributes via declaration merging.
+declare module '@balena/abstract-sql-compiler' {
+	export interface AbstractSqlTable {
+		validator?: z.ZodObject;
+	}
+}
+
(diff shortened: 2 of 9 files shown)


src/pinejs-session-store/pinejs-session-store.ts
```diff
@@ -72,7 +72,7 @@ export class PinejsSessionStore extends Store {
 	public set = ((sid, data, callback) => {
 		const body = {
 			session_id: sid,
-			data,
+			data: { ...data },
```
otaviojacobi, line 76: Why is this now necessary?

Page-: Because it's an instance of the `Session` class which is not a valid "record", ie it's not just an object, but can be converted to one for as far as we care, and this is the easiest way to convert it now that we actually validate it before coercing it

src/sbvr-api/validation.ts
```diff
@@ -0,0 +1,59 @@
...
+export const deepFreezeExceptValidator = (obj: AnyObject) => {
+	Object.freeze(obj);
+
+	for (const prop of Object.getOwnPropertyNames(obj)) {
+		// We skip the definition because we know it's a property we've defined that will throw an error in some cases
+		if (
+			prop !== 'validator' &&
+			Object.hasOwn(obj, prop) &&
+			obj[prop] !== null &&
+			!['object', 'function'].includes(typeof obj[prop])
```
joshbwlng: So we're not freezing nested objects? Also, could `!Object.isFrozen()` help here?

src/sbvr-api/validation.ts
```diff
@@ -0,0 +1,59 @@
...
+		) {
+			deepFreezeExceptValidator(obj);
+		}
+	}
+};
+
+export const stringifyIgnoreValidator = (key: string, value: any) => {
```
joshbwlng, line 40: I wouldn't mind a quick comment about the point of this function, it took me a minute to realize what it was doing.

src/sbvr-api/validation.ts
```diff
@@ -0,0 +1,45 @@
...
+const { default: sbvrTypes } = $sbvrTypes;
+
+// Augment express.js with pinejs-specific attributes via declaration merging.
+declare module '@balena/abstract-sql-compiler' {
+	export interface AbstractSqlTable {
+		validator?: z.ZodObject;
+	}
+}
+
+export const getValidator = (table: AbstractSqlTable) => {
```
otaviojacobi, line 17: Being very picky here, but maybe `setupAndGetValidator`? It feels wrong to be reading `getValidator` on `prepareModel`  being used without any return just for the side effect of setting it up on the table

otaviojacobi (approved):

PR #1029, opened 2026-07-09
Permissions: fix `deepFreezeExceptDefinition` recursion

Page-:
https://balena.fibery.io/Work/Project/1002

Change-type: patch

### DIFF
--- a/src/sbvr-api/permissions.ts
+++ b/src/sbvr-api/permissions.ts
@@ -703,18 +703,35 @@ const onceGetter = <T, U extends keyof T>(
 	});
 };
 
-const deepFreezeExceptDefinition = (obj: AnyObject) => {
-	Object.freeze(obj);
+const WILDCARD = Symbol('*');
+const TERMINATE = Symbol('TERMINATE');
+const deepFreezeExceptPaths = (
+	obj: AnyObject,
+	excludePaths: Array<Array<string | typeof WILDCARD | typeof TERMINATE>> = [],
+) => {
+	if (!excludePaths.some((path) => path.length === 0)) {
+		Object.freeze(obj);
+	}
 
-	for (const prop of Object.getOwnPropertyNames(obj)) {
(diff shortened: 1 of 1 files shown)


src/sbvr-api/permissions.ts
```diff
@@ -703,18 +703,33 @@ const onceGetter = <T, U extends keyof T>(
...
+			// If we're terminating at this path then we don't want to even try to access it as that could trigger getters or other side effects.
+			continue;
+		}
+
 		if (
-			prop !== 'definition' &&
 			Object.hasOwn(obj, prop) &&
 			obj[prop] !== null &&
-			!['object', 'function'].includes(typeof obj[prop])
+			(typeof obj[prop] === 'object' || typeof obj[prop] === 'function')
```
joshbwlng, line 732: Is this type check reversal intentional?

Page-: Yes, the previous one was broken because it was inverted. I don't know how/why that happened but I suspect it was because with it being correct we'd have hit infinite recursion due to `deepFreezeExceptDefinition(obj)` vs `deepFreezeExceptDefinition(obj[prop])` so it got inverted to "fix" it

src/sbvr-api/permissions.ts
```diff
@@ -703,18 +703,33 @@ const onceGetter = <T, U extends keyof T>(
...
+		Object.freeze(obj);
+	}
 
 	for (const prop of Object.getOwnPropertyNames(obj)) {
-		// We skip the definition because we know it's a property we've defined that will throw an error in some cases
+		const newExcludePaths = excludePaths
+			.filter(
+				(path) => path.length > 0 && (path[0] === prop || path[0] === WILDCARD),
+			)
+			.map((path) => path.slice(1));
```
otaviojacobi: Maybe for + if for single iteration?

otaviojacobi (approved):

